PR #220, opened 2013-08-08
Search

loicgasser:
I think it makes sense to merge it now. If you think I need to squash those commits I'll do it.

What's done:
1. Load suggestions using the search service of chsdi3
2. Search for layers is topic dependent (react on topic change)
3. Language dependent
4. Uses a bounding box to look for locations
5. Zoom in on location selection
6. Add Layer on map info selection or hovering
7. Styling close to specs
8. Clear button to erase the input value
9. Help buttons in the footer
10. Adapt correctly according to screen width
11. Prepared for translations
12. Example

What needs to be done:
1. Query features in a queryable layer is in the map
2. Add feature to the map when selected
3. Adapt it for mobile
4. Add double click (dropdown list doesn't close and the layer or map info are added)
5. Tests
6. More styling (screen width dependent for dropdown list for instance)
7. Legend button should get the legend... (alert msg at the moment)
8. Probably move some redundant functions in a unique service
9 Translations
10. Search button (don't really know the use case though)

Note:
Not happy with the footer definition, I tried to use ng-include, but I had digest errors... Any idea?

### DIFF
--- a/src/components/search/SearchDirective.js
+++ b/src/components/search/SearchDirective.js
@@ -156,7 +156,8 @@
                     return template;
                   },
                   remote: {
-                    url: gaUrlUtils.append(options.serviceUrl, 'type=locations'),
+                    url: gaUrlUtils.append(options.serviceUrl,
+                        'type=locations'),
                     dataType: 'jsonp',
                     cache: false,
                     replace: function(url, searchText) {


elemoine:
> I tried to use ng-include, but I had digest errors... Any idea?

This is due to an incompatibility between the `gaPermalink` service and Angular's `$location` service, onto which `ng-include` depends. We've recently been fighting with this bug and ended up with an ugly workaround. See https://github.com/camptocamp/ngeo/pull/160 for details.

This is just in case you'll need to use `ng-include` again in the future.

gjn:
Now that's ugly.

Thanks for the heads-up, Eric.

elemoine:
We really should fix Angular. But at this point, I don't know what the proper fix would look like.

Issue #2019, opened 2014-12-23
Query tool: searchable and queryable layers 


cedricmoullet:
Attribute filter tool is available only for searchable layers.
Spatial filter tool is available for searchable and queryable layers.

PR #2023, opened 2015-01-01
[Query Tool] Only allow searchable layers in query tool

gjn:
This is to address https://github.com/geoadmin/mf-geoadmin3/issues/2019

Note ( @cedricmoullet  @ltclm ): we have a conflict. `ch.bazl.luftfahrthindernis` is not marked as a searchable layer yet in the db, so with this PR, we can't use the query tool with this layer (haha). So we would need to adapt that. But this would mean that we need to have a sphinx index also for it to be searchable in the search box (as all searchable layers have), but are we ready for this with sphinx? (data is updated on a regular basis).

Alternatively, we could use yet another new flag...but not sure about that, as we already have many quite similar (e.g. selectbyrectangle flag and queryable flag)

### DIFF
--- a/src/components/query/QueryDirective.js
+++ b/src/components/query/QueryDirective.js
@@ -18,6 +18,7 @@
     var stored;
     $scope.queryType = 1; // Filter attributes
     $scope.searchableLayers = [];
+    $scope.selectByRectangleLayers = [];
     $scope.queriesPredef = [];
     $scope.filters = [];
 
@@ -181,7 +182,7 @@
         var geom = $scope.geometry.getExtent();
         var features = [];
         gaQuery.getLayersFeaturesByBbox($scope,
-            $scope.searchableLayers,
+            $scope.selectByRectangleLayers,
             geom
         ).then(function(layerFeatures) {
           features = features.concat(layerFeatures);
@@ -209,7 +210,7 @@
         var lang = $translate.use();
         var features = [];
         angular.forEach(
-            $scope.searchableLayers,
+            $scope.selectByRectangleLayers,
             function(layer) {
               gaQuery.getLayerIdentifyFeatures(
                   $scope,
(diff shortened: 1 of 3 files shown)


ltclm:
we should create the needed sphinx indexes make this layer 'standard' searchable=true.
db trigger will update the sphinx index if needed.

cedricmoullet:
Who will take care of this sphinx index ? @daguer ?

ltclm:
i will prepare a sphinx pr now.

ltclm:
The pr has been created https://github.com/geoadmin/service-sphinxsearch/pull/187 and applied on dev.

src/components/query/QueryDirective.js
```diff
@@ -352,8 +353,16 @@
 
     // Watcher/listener
     $scope.layers = $scope.map.getLayers().getArray();
-    $scope.layerFilter = gaLayerFilters.selectByRectangle;
-    $scope.$watchCollection('layers | filter:layerFilter', function(layers) {
+    $scope.selectByRectangleFilter = gaLayerFilters.selectByRectangle;
+    $scope.$watchCollection('layers | filter:selectByRectangleFilter',
```
cedricmoullet: Is it still needed ?

cedricmoullet:
@oterral please review

src/components/query/QueryDirective.js
gjn: Yes. We now have 2 sets of layers. 1 for query tool (searchable layers) and another for the select by rectangle tool (queryable layers)

oterral: Since we use the same service for bbox and filter requests (features identify which use the database). I think the searchable property is the property to use. We can get rid of selectByRectangle stuff here.

But maybe I miss something.

oterral:
Code LGTM . You can merge.

src/components/query/QueryDirective.js
gjn: The tool itself might be confusing for the user with this setup...

Issue #2001, opened 2014-12-20
Query tool: results should be all highlighted


cedricmoullet:
(in order to allow a map print of the result)

PR #2024, opened 2015-01-01
[Query Tool] Draw all features in featuretree results

gjn:
This is for https://github.com/geoadmin/mf-geoadmin3/issues/2001


oterral:
LGTM. thx for this.

Issue #1928, opened 2014-12-04
[Zeitreihen] set default timestamp 1864 when adding layer


mariokeusen:
This is a requirement & decision out of the "Zeitreihen" project group, which was negotiated today with David. According to Patrick, it's not (yet) directly possible to set "default" timestamps for time-enabled layers:

UC: open application, add the layer ch.swisstopo.zeitreihen
Result: the timeslider is set to 2012 (=newest) by default
Expected: the timeslider is set to 1864 by default (first timestamp, where the whole CH is covered with maps)

Please reassign if necessary.

Thanks in advance

PR #2025, opened 2015-01-01
[Zeitreihen] Force default timestamp for zeitreihen layer

gjn:
This is for https://github.com/geoadmin/mf-geoadmin3/issues/1928

I consider this a dirty hack. We should come up with a solution for layers that don't have the last timestamp as default timestamp.

We already have the 'timeBehaviour' parameter (with `all` and `last` settings) which we could extend to include a specific timestamp to be transmitted with the layers configuration.


cedricmoullet:
Indeed, this is a dirty hack

PR #2043, opened 2015-01-06
Fix the definition of requests parameters

### LINKED ISSUE #2042: [query tool] serial print of extended tooltips of result list only in German
UC: open the aviation topic on integration, switch to FR, open the query tool and make a query, print the results

Result: the html-extended tooltips of the objects of the query results appears in a new window in DE
Expected: the html-extended tooltips of the objects of the query results appears in a new window in FR

Same for IT and EN (but this is no showstopper for me for tomorrow).

oterral:
Fix #2042


oterral:
Trivial. self merge

PR #2047, opened 2015-01-06
Fix display default input text value for a predefined query on IE9

### LINKED ISSUE #2045: [query tool] IE9: default attribute values in predefined query not visible
UC: open topic aviation with IE9, open the query tool, choose the first predefined query, look at the values in the attribute fields.

Result: there is a value list visible in each field
Expected: there is a default value visible in each field (like in all other browsers)
query_ie9_default_predef_values

oterral:
Fix  #2045

Test [here](http://mf-geoadmin3.dev.bgdi.ch/teo_fix_ie9/?lang=fr&topic=aviation&bgLayer=ch.swisstopo.pixelkarte-grau&X=190000.00&Y=660000.00&zoom=1&catalogNodes=1379,1381,1514&layers=ch.bazl.luftfahrthindernis)

It's a little bit hackish, I think it could be better.

### DIFF
--- a/src/components/PlaceholderDirective.js
+++ b/src/components/PlaceholderDirective.js
@@ -50,6 +50,12 @@
           if (elt.val() == '') {
             displayPlaceholder(elt);
           }
+        }).change(function(evt) {
+          var elt = $(evt.target);
+          if (elt.val() != elt.attr('placeholder')) {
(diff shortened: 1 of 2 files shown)


gjn:
Code LGTM. For rewdability, would put the IE9 Ccondition first, but that's really minor.

PR #2061, opened 2015-01-08
Teo default time

### LINKED ISSUE #2054: zeitreise: date is not shown on launch
ios iphone

start  bit.ly/140bxVS

result
unbenannt

expected result: 1864 is shown in the year box as it is on desktop

oterral:
Fix #2054 

This PR contains the fix and small optimisation for time selector directives (avoid applying a year multiple times when the page opens with a permalink)

[Test](http://mf-geoadmin3.dev.bgdi.ch/teo_default_time/mobile.html?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&X=190000.00&Y=660000.00&zoom=1&layers_timestamp=18641231&time=1864&mobile=true&layers=ch.swisstopo.zeitreihen)


gjn:
Tested and looks good.

Not a big fan of replacing broadcast with emit and therefore introducing rootScope (I consider it a global object) into several directives. I know it's probably motivated by performance, but I don't think it's that big of an impact. Broadcasts are explicit and are not like watcher. Watcher are checked on every digest cycles while broadcasts travel the scope hieararchy on every call. Also, the timeselector is used rarely so this won't affect > 80% of the use cases

Did you measure a big performance boost while working with the timeselector?

oterral:
```Did you measure a big performance boost while working with the timeselector?````
No but I'm sure it's optimized now :)  
It was also  to see all components using the event in the same way, that makes  no sense that one component use the scope.$on and the other $rootScope,$on  and if we use $rootScope.$on using broadcast is useless.

``` travel the scope hieararchy on every call ```

exactly, if we can avoid to go through tousand of useless scopes I think it's good (mainly on mobile)

I think using rootScope makes sense because this event  affects all the application like gaTopicChange, $translateChangeEnd ,..

With that  we can differentiate global events and events called internally (between components in a sub  controller for example).

but i understand your concern may be should discuss that when you come back.

gjn:
I just prefer design principles over premature (e.g unmeasured) optimisation. I think the event is triggered so rarely (even more rare on mobile) that it's simply not worth to sacrifice the non-global-variable principle [1].

On the other hand, one could argue that rootScope is _not_ a global variable but it's injected...so it's pretty subjective :=)

[1] http://c2.com/cgi/wiki?GlobalVariablesAreBad

oterral:
After documentating more  myself it seems $broadcast was fixed in 1.3 and go through only scope which are registered with event so $emit is not anymore faster than $broadcast.

See : 
http://jsperf.com/rootscope-emit-vs-rootscope-broadcast
http://jsperf.com/rootscope-emit-vs-rootscope-broadcast/70
http://stackoverflow.com/questions/11252780/whats-the-correct-way-to-communicate-between-controllers-in-angularjs/19498009#19498009

My mistake, I remove the first commit and use scope.$on everywhere, agree?


Thx for having pointing this @gjn

oterral:
I've also read that trigger event on $rootscope should be done in service : 

http://onehungrymind.com/angularjs-communicating-between-controllers/

gjn:
Interesting links!

I agree to remove the first commit. Thanks alot.

gjn:
Regarding your last comment: one service specifically to wrap $rootScope.$broadcast?

oterral:
Yes I thought abour a TimeService which stored the current year applied and whcich broadcast the event when the value is changed. 

We could also add in this service all the duplicate  functions we have to transformi timestamps to a readable year.

oterral:
COmmit removed

Issue #1985, opened 2014-12-18
jpeg format don't work for feedback tool


pfanguin:

pfanguin:
![jpegformat](https://cloud.githubusercontent.com/assets/6862714/5658856/309b376c-9709-11e4-8860-5ff3b6ab103d.JPG)

Maybe the problem is the uppercase "JPG"

gjn:
Indeed, it's the uppercase. PR is ready to address this and ignore the case.

PR #2063, opened 2015-01-09
Check for feedback attachement is case insensitive

gjn:
This is for https://github.com/geoadmin/mf-geoadmin3/issues/1985


oterral:
good one thx

PR #2068, opened 2015-01-09
Preload the last zoom available in offline mode

### LINKED ISSUE #1606: [DEV APPCACHE] offline clientzoom fails
reproduce

1) save this perimeter offline http://mf-geoadmin3.dev.bgdi.ch/dev_appcache/?X=197896.86&Y=648843.25&zoom=9&lang=de&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.hiks-dufour,ch.kantone.cadastralwebmap-farbe&layers_visibility=false,true

2) go offline

3) zoom in to zoom=10

4) pan to the left/up/down keeping the zoom level 10

Result: 
a green area will be encoutered 
unbenannt

client zoom will not  work

Expected result:

client zoom works when entering an area which is not yet loaded

Workaround:
-Zoom out on the green area and then zoom in again-> client zoom works

oterral:
Fix the long time  #1606 

[Test](http://mf-geoadmin3.dev.bgdi.ch/teo_fix_1606/mobile.html?lang=fr&topic=luftbilder&bgLayer=ch.swisstopo.pixelkarte-grau&X=198239.06&Y=583349.38&zoom=13&catalogNodes=1179,1180,1186&mobile=true&layers=ch.kantone.cadastralwebmap-farbe)

### DIFF
--- a/src/components/map/MapService.js
+++ b/src/components/map/MapService.js
@@ -743,6 +743,7 @@
               extent: olSource.getProjection().getExtent(),
               minResolution: gaNetworkStatus.offline ? null :
                   layer.minResolution,
+              preload: gaNetworkStatus.offline ? gaMapUtils.preload : 0,
               maxResolution: layer.maxResolution,
               opacity: layer.opacity || 1,
               attribution: layer.attribution,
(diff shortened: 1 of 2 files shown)


oterral:
@gjn an opinion about this?

gjn:
For code perspective, not sure about doing the same thing in 2 locations (do we still pass inside mapservice when offline? If so, why can't we take this mapservice definition to use in the offline service?). I think the `6` should be defined once.

Otherwise, LGTM.

oterral:
Agree I had a cycling depency (gaOffline -> gaLayers -> gaOffline)  I think, otherwise I can put it in gaMapUtils,

oterral:
Updated

gjn:
Thanks.

PR #2069, opened 2015-01-09
Hide header, pulldown, copyright and footer until the language is loaded

oterral:
Avoid to see elements from the beginning of loading to the translation end event 

[Test](http://mf-geoadmin3.dev.bgdi.ch/teo_cloak/)

### DIFF
--- a/src/components/query/partials/query.html
+++ b/src/components/query/partials/query.html
@@ -70,7 +70,7 @@
   <button class="btn btn-default" 
           ng-disabled="loading || !filters[0].value"
           ng-click="searchByAttributes()">
-    <i ng-class="{'ga-loader icon-refresh': loading}"></i>
+    <i ng-class="{'icon-refresh': loading}"></i>
     <span translate>query_search<span>
   </button>
 </div>
--- a/src/components/query/style/query.less
+++ b/src/components/query/style/query.less
@@ -31,7 +31,11 @@
     line-height: 1.5;
     border-radius: 1px;
   }
- 
+
+  .icon-refresh {
+    .ga-spin();
+  }
+
   .btn-default{
     width: 100%;
   }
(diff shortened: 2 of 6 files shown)


gjn:
Very nice! On fast connections, this looks great/better.

Should be tested on slow connections, especially mobile. Might get the impression that nothing is loading. Would it be possible to add a loading indicator while the page is cloaked?

oterral:
I think we could add something like a loader or just display the actual loader (the red bar on the top)

oterral:
New version with an application loader only for mobile [test](http://mf-geoadmin3.dev.bgdi.ch/teo_cloak/mobile.html?mobile=true&lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&X=190000.00&Y=660000.00&zoom=1)

ponceta:
Nice! Works great on android 4.1.1.

oterral:
@davidoesch @cedricmoullet @gjn @loicgasser @procrastinatio any opinions for the loader?

davidoesch:
sorry to be here the fun breaker:
- by law the confederation logo shall not be animated (Wappenschutzgesetzt
1933 , no joke)

in general: "a loader " icon si a big step backward... : "this application
is loading...." etc is admintting that sth is slow. please no loading icon.
I'm 100% sure that @cedricmoullet agrees on that (since he always teased me
when i wanted a loading icon  ;-))

the approaach we had no: the top thin red line when loading the app on
mobile was very good...why chnage this behaviour?

ponceta:
Or we could just roll the geoportal logo? ;)

Ok I'm out!

oterral:
On IE11 mobile that was so beautiful :)

I can put the red loading bar instead.

davidoesch:
+11: red loading bar.....

oterral:
Loader removed,  the red bar or spin icon  is displaying when application loading (only on small screen)

Good to review

gjn:
Fantastico.

Issue #2067, opened 2015-01-09
Print: use QRcode link matching the PDF view


procrastinatio:
Now the QRcode url is the one of the page, not of the selected print area. Update X,Y and zoom paramters accordingly

PR #2071, opened 2015-01-09
QRcodeUrl: need to match the printing area

procrastinatio:
The QRcode has to zoom a approx. zoom level of the printing area with may be very different from the application's permalink. The X and Y are however the same (printing area is centered).

Fix https://github.com/geoadmin/mf-geoadmin3/issues/2067

### DIFF
--- a/src/components/print/PrintDirective.js
+++ b/src/components/print/PrintDirective.js
@@ -8,8 +8,8 @@
      'pascalprecht.translate']);
 
   module.controller('GaPrintDirectiveController', function($rootScope, $scope,
-      $http, $q, $window, $translate, $timeout, gaLayers, gaPermalink,
-      gaBrowserSniffer, gaWaitCursor) {
+      $http, $q, $window, $translate, $timeout, gaLayers, gaMapUtils, 
+      gaPermalink, gaBrowserSniffer, gaWaitCursor) {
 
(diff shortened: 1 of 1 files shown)


src/components/print/PrintDirective.js
```diff
@@ -564,6 +564,21 @@
       }
     };
 
+    var getZoomFromScale = function(scale) {
+      var diff, i, len;
+      var minDiff = Number.POSITIVE_INFINITY;
```
gjn: `diff` and `minDiff` variables are not used.

gjn:
Added a minor comment. Tested and otherwise looks good.

gjn:
@procrastinatio Could you remove the unneeded variables. Then we can merge it.

procrastinatio:
OK, done

cedricmoullet:
@procrastinatio could you please rebase ?

src/components/print/PrintDirective.js
```diff
@@ -564,6 +564,20 @@
       }
     };
 
+    var getZoomFromScale = function(scale) {
```
oterral: May be this one could be move to gaMapUtils , it could be useful.

gjn:
@procrastinatio Please rebase and merge.

PR #2074, opened 2015-01-12
Enlarge popups to avoid slidebar

ponceta:
Greater popups for greater content and less slidebar

Test [HERE](http://mf-geoadmin3.dev.bgdi.ch/ltpoa/)

### DIFF
--- a/src/components/popup/style/popup.less
+++ b/src/components/popup/style/popup.less
@@ -1,4 +1,4 @@
-@ga-popup-max-height: 400px;
+@ga-popup-max-height: 440px;
 @ga-popup-reduced-height: 40px;
 @ga-popup-max-width: 620px;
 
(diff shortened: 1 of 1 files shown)


ponceta:
Before
image

After
image

gjn:
Impact on mobile?

ponceta:
There are no help popups on mobile.
Actually there is no link to the help from mobile version (which is bad, no?) (or I didn't find it!)

ponceta:
And on very small screens the popups are reduced with slidebar natively. Since we change only the max-height and max-width it will not affect these.

gjn:
This PR changes all popups, not just help popups.

Tooltips and Feedback, I think, are active on mobile (there might be others). Please test them.

ponceta:
Fine on Android, @kandre could you please test feedback and tooltip on IOS?
I'll check on Win Mobile with lumia.

kandre:
I works well on ios iphone 4

gjn:
Very good. Thanks alot for testing and having a 2nd look!

ponceta:
Ty @kandre

PR #2081, opened 2015-01-14
[WIP]Use a feature style function for highlighted feature

### LINKED ISSUE #2076: Query tool highlight improvement
Currently, the query tool highlight uses the real geometry and highlight it. On small scale, the highlighted features are not very visible. 
We need to find a way to make the highlight more visible. What about a scale dependent behaviour ?

For zoom level 0 to 5:
- Point: use current circle
- Lines: if line bbox is smaller than the circle, use the circle, otherwise, like current highlight
- Polygons: if polygon bbox is smaller than the circle, use the circle, otherwise like current highlight

For zoom level 6 to 13: like now

Not sure that it's easily implementable.
@davidoesch @mariokeusen @gjn @oterral Thoughts ?

oterral:
Fix   #2076 still work in progress 

Display a feature with huge stroke between zoom 0 and 4 when the mouse is over a featurein the feature tree.

[Test](http://mf-geoadmin3.dev.bgdi.ch/teo_highlight/)

### DIFF
--- a/src/components/GeometryService.js
+++ b/src/components/GeometryService.js
@@ -1,0 +1,44 @@
+(function() {
+  goog.provide('ga_geometry_service');
+
+  var module = angular.module('ga_geometry_service', []);
+
+  module.provider('gaGeom', function() {
+    var getFlatCoordinates = function(coords) {
+      if (coords[0] instanceof Array) {
+        var flatCoords = [];
+        angular.forEach(coords, function(coord) {
+          flatCoords = flatCoords.concat(getFlatCoordinates(coord));
+        });
+        return flatCoords;
+      }
+      return [coords];
+    };
+    // Return an array of coordinates representing the geometry
+    var getCoordinates = function(geom) {
+      if (geom instanceof ol.geom.GeometryCollection) {
+        var flatCoords = [];
+        angular.forEach(geom.getGeometries(), function(aGeom) {
+          flatCoords = flatCoords.concat(getCoordinates(aGeom));
+        });
+        return flatCoords;
+      }
+      return getFlatCoordinates(geom.getCoordinates());
+    };
+    this.$get = function() {
+      var Geom = function() {
+        this.centroid = function(geom) {
+          var flatCoords = getCoordinates(geom);
+          var xSum = 0, ySum = 0, len = flatCoords.length;
+          angular.forEach(flatCoords, function(coord) {
+            xSum += coord[0];
+            ySum += coord[1];
+          });
+          return new ol.geom.Point([xSum / len, ySum / len]);
+        };
+      };
+      return new Geom();
+    };
+  });
+})();
+
--- a/src/components/StylesService.js
+++ b/src/components/StylesService.js
@@ -1,7 +1,11 @@
 (function() {
   goog.provide('ga_styles_service');
+  goog.require('ga_geometry_service');
 
-  var module = angular.module('ga_styles_service', []);
+
+  var module = angular.module('ga_styles_service', [
+    'ga_geometry_service'
+  ]);
 
   module.provider('gaStyleFactory', function() {
     var DEFAULT_FONT = 'normal 16px Helvetica';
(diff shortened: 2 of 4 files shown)


cedricmoullet:
Fine for "hover higlight", but we should find a solution also for the normal highlight

oterral:
I can do the same for select style.

oterral:
I have a better idea I wiil display a point instead of an ugly large stroke.

oterral:
Updated . the solution of the point is very good on the case on luftfahrthinderniss but bad for other layers.

loicgasser:
Nice! I would still leave the connecting line in between (zoom 0 - 4)
Rather than using a fixed resolution we could use

`max([abs(Xmin-Xmax), abs(Ymin-Ymax)])` as metric and relate it to the current resolution to produce a ratio. Would be adaptable to all layers.

loicgasser:
`X Y` being the bbox of the line.

oterral:
very good idea thx @loicgasser

oterral:
PR updated with @loicgasser solution

loicgasser:
So with the actual deployed branch, we have only one point for low resolution? (which is fine by me)

oterral:
exactly,  depending on the size of the feature's geometry (excepts for GeometryCollection)

oterral:
Update with the use of centroid

oterral:
Ready to review

gjn:
Not sure this works for all layers very well. E.g. http://mf-geoadmin3.dev.bgdi.ch/teo_highlight/?X=237855.25&Y=622201.63&zoom=5&lang=de&topic=aviation&bgLayer=ch.swisstopo.pixelkarte-grau&catalogNodes=1379,1381,1514&layers=ch.astra.ivs-nat (select by rectangle), it's very strange to have points for clearly visible polylines. (There might be something fishy with this layers as well)

Another example to test:
http://mf-geoadmin3.dev.bgdi.ch/teo_highlight/?X=134661.03&Y=585474.16&zoom=4&lang=de&topic=aviation&bgLayer=voidLayer&catalogNodes=1379,1381,1514&layers=ch.bafu.wasser-leitungen
Or Gemeindegrenzen, where at certain zoom levels the mixing of points and polygons is rather strange:
http://mf-geoadmin3.dev.bgdi.ch/teo_highlight/?X=122161.03&Y=641474.16&zoom=2&lang=de&topic=aviation&bgLayer=voidLayer&catalogNodes=1379,1381,1514&layers=ch.swisstopo.swissboundaries3d-gemeinde-flaeche.fill

Maybe try to show points at lower zoom levels and smaller geometries than currently?

@davidoesch @cedricmoullet Thoughts?

oterral:
For the layer, `ch.astra.ivs-nat` . There is something weird because not all the features are displayed on the tiles.
I agree it's not perfect for all the layers.

davidoesch:
-ch.astra.ivs-nat is weird.... must be in data
-plus one for "Maybe try to show points at lower zoom levels and smaller geometries than currently"

oterral:
what do you mean by "smaller geometries than currently" ?

gjn:
We should show points at lower zoom levels than currently. Currently, we show polyline/polygone features as points at zoom levels and geometry sizes where the standard polyline/polygone highlight would make more sense.

Maybe demo tomorrow.

cedricmoullet:
@ltclm and @oterral could you analyze the data ?

ltclm:
apparently there is an issue/misunderstanding with the latest data delivery of layer ``ch.astra.ivs-nat``.
The data we integrated on 2013-11 contained new geometries, but it was declared as attribute-only update, see https://redmine.prod.bgdi.ch/issues/4271
you can click on the end of the line to visualize the problem:
http://map.geo.admin.ch/?X=167476.08&Y=606381.02&zoom=12&lang=de&topic=aviation&bgLayer=voidLayer&catalogNodes=1379,1381,1514&layers=ch.astra.ivs-nat
@mariokeusen can you ask the datenherr if we can generate tiles with his last data delivery? otherwise we have to find?? and delete the non-visible features from the table.

Issue #1647, opened 2014-09-15
[DEV APPCACHE] KML warning


davidoesch:
We need a warning when someone want to save his KML file for offline usage and it does not  work:
Since it appears on the screen after a complete download... but when offline it is not there anymore

give it a try
http://mf-geoadmin3.dev.bgdi.ch/dev_appcache/?X=204700.00&Y=643050.00&zoom=4&lang=en&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=KML||https:%2F%2Fwww.google.ch%2Fmaps%2Fms%3Fauthuser%3D0%26vps%3D2%26hl%3Den%26ie%3DUTF8%26msa%3D0%26output%3Dkml%26msid%3D207618250985502323563.0004f1d5195a689db6de9

PR #2084, opened 2015-01-15
Save KML for offline use

### LINKED ISSUE #1839: [DEV APPCACHE] KML save offline
Hi 

Having a KML in a permalink either as
- local file
- URL

the user expects on offline to store the linked KML as well

@cedricmoullet @oterral makes sense?

oterral:
Fix #1839 #1647

Wait for https://github.com/openlayers/ol3/pull/3128

[Test](http://mf-geoadmin3.dev.bgdi.ch/teo_kml_offline/?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&X=190000.00&Y=660000.00&zoom=1&layers=KML%7C%7Chttp:%2F%2Fwww.bikingspots.ch%2FcreateKMLFile.php%3Ffilename%3Dbikingspots4c3a0a25255db.xml%26fc%3D1)

Missing some tests and translations


kandre:
Nice !! and tough mtb test !

ponceta:
No worry

image

We'll get this fine!

oterral:
good for review 

- [ ] Add missing translations

src/components/offline/OfflineService.js
```diff
@@ -395,6 +420,18 @@
             var layer = layers[i];
             layersIds.push(layer.id);
             layersOpacity.push(layer.invertedOpacity);
+
+            // if the layer is a KML
+            if (gaKml.isKmlLayer(layer) && /http/.test(layer.url)) {
```
gjn: Would make sense to also check for `https` here.

oterral: `http`is include in`https` but I can improve a bit the regex to be sure.

src/components/offline/OfflineService.js
```diff
@@ -97,9 +63,54 @@
     initDownloadStatus();
 
 
-    this.$get = function($rootScope, $timeout, $translate, $window,
-        gaBrowserSniffer, gaGlobalOptions, gaLayers, gaMapUtils,
+    this.$get = function($http, $rootScope, $timeout, $translate, $window,
+        gaBrowserSniffer, gaGlobalOptions, gaKml, gaLayers, gaMapUtils,
```
gjn: Hmm. I'm torn. gaKml is only needed to check if layer is kml, right (gaKml.isKmlLayer)? Maybe just create internal function that does the check instead of injecting a dependency?

As said, not sure here.

src/components/offline/OfflineService.js
gjn: dummy, of course. Improvement might still be a good idea. Can we use ^ and `://` ?

src/components/offline/OfflineService.js
oterral: Or I can move this utility function in `gaMapUtils`, what do you think?

src/components/offline/OfflineService.js
oterral: yes i'll do that

src/components/offline/OfflineService.js
gjn: Probably best, yes.

ponceta:
Can't save drawn kml!
Go test branch / draw something / try to dl it.

oterral:
@ponceta you're insane !!!!! it's only for impored KML.;)  . You can draw then import and then save :)

ponceta:
<3 sorry for catching bugs ;)

oterral:
Good for final review/test

src/components/ExportKmlService.js
```diff
@@ -91,13 +91,7 @@
...
 
           if (exportFeatures.length > 0) {
-            var node = new ol.format.KML().writeFeatures(exportFeatures);
-            kmlString = node.outerHtml;
-            //Some browser don't support the node.outerHtml property. For
-            //thos browsers, we need to serialize differently.
-            if (!kmlString) {
-              kmlString = new XMLSerializer().serializeToString(node);
-            }
+            kmlString = new ol.format.KML().writeFeatures(exportFeatures);
```
gjn: Does ol3 now take care of this?

gjn:
Code looks very good.

Just a small thing: when loading the kml in off-line mode, I don't get the start and end symbols (I get the track). Am I doing something wrong or is this problematic?

gjn:
Ok, it's normal (icons are from extern and we don't safe those).

src/components/ExportKmlService.js
oterral: Yes OL3 returns a string now

Issue #2027, opened 2015-01-05
Remove time hack for zeitreihen


cedricmoullet:
https://github.com/geoadmin/mf-geoadmin3/blob/d88beb0ff8b8f63a87b93a406f9d2700c9f47670/src/components/map/MapService.js#L878

PR #2085, opened 2015-01-16
Remove zeitreihen hack

gjn:
DB has been adapted. timeBehaviour parameter now contains the timestamp that we want to have per default. This could/can be used for different layers also. As of now, only 8 digit timestamps are supported.

The other timeBevhaviour possible valures ('last' and 'all') don't change behaviour.

This is for https://github.com/geoadmin/mf-geoadmin3/issues/2027

### DIFF
--- a/src/components/map/MapService.js
+++ b/src/components/map/MapService.js
@@ -874,12 +874,12 @@
           }
 
           if (!angular.isDefined(yearStr)) {
-            var timeBehaviour = this.getLayerProperty(bodId,
-                'timeBehaviour');
-            yearStr = (timeBehaviour === 'all' || timestamps.length == 0) ?
-                undefined : timestamps[0];
-            if (bodId == 'ch.swisstopo.zeitreihen') {
-              yearStr = '18641231';
+            var timeBehaviour = this.getLayerProperty(bodId, 'timeBehaviour');
+            //check if specific 4/6/8 digit timestamp is specified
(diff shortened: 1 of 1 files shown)


src/components/map/MapService.js
```diff
@@ -874,12 +874,12 @@
...
           if (!angular.isDefined(yearStr)) {
-            var timeBehaviour = this.getLayerProperty(bodId,
-                'timeBehaviour');
-            yearStr = (timeBehaviour === 'all' || timestamps.length == 0) ?
-                undefined : timestamps[0];
-            if (bodId == 'ch.swisstopo.zeitreihen') {
-              yearStr = '18641231';
+            var timeBehaviour = this.getLayerProperty(bodId, 'timeBehaviour');
+            //check if specific 8 digit timestamp is specified
+            if (/^\d{8}$/.test(timeBehaviour)) {
```
oterral: In case of a WMS layer like `ch.swisstopo.lubis-bildstreifen`  tiemstamps are 4 digits

gjn: This comes from db and 8 digits is the only possibility right now.

But it probably makes sense to include 4-8 digits.

gjn:
ok, I've adde 4 and 6 digits as a possibility. Thanks for the review.

oterral:
thx for this

Issue #2075, opened 2015-01-13
Whitelist ch.bazl.luftfahrthindernis


loicgasser:
Step to reproduce:
1. Open map.geo.admin.ch
2. Switch to topic aviation
3. Switch back to topic ech
4. Click a feature on the map

Result:
HTML popup request fails (No layer with id ch.bazl.luftfahrthindernis)

Expected result:
If ch.bazl.luftfahrthindernis is part of ech -> htmlpopup does not fail
if ch.bazl.luftfahrthindernis not part of ech -> layer is removed when topic is changed

@ltclm Is ch.bazl.luftfahrthindernis supposed to be part of ech topic?

ltclm:
@mariokeusen should the layer ch.bazl.luftfahrhindernis be part of topic ech?

mariokeusen:
Is not (yet) foreseen. But I have a meeting with the data owner next week and will request, resp. recommend that.

mariokeusen:
Will check that with customer on Monday. Tbc...

mariokeusen:
Customer gave me green light for ECH. 
@alcapat : i have added the layer in ID 1561 in the topic config. Please consider this for the next deploy

PR #2086, opened 2015-01-16
Fix layers on topic change

gjn:
https://github.com/geoadmin/mf-geoadmin3/issues/2075 showed that we introduced a regression on topic change. On topic change, all layers which are part of topic ech should remain in the layer section. On the other hand, layers that are not part of the new topic or topic ech should be removed.

This was originally done with https://github.com/geoadmin/mf-geoadmin3/commit/b4e1ab22c0172e7a1a0b0c218baa6c5cf1fc9fc3#diff-cc254e043ae817085972191ba79441de , then knowingly removed with https://github.com/geoadmin/mf-geoadmin3/commit/20aef49ebd4bd3fc11330ce3c74331b6f869fdee#diff-cc254e043ae817085972191ba79441de , but I forgot about it because back-end needed to be adapted (see https://github.com/geoadmin/mf-chsdi3/pull/1085) as we switched to 'all' topic for layersConfig because of offline. Oh well...

### DIFF
--- a/src/components/layermanager/LayermanagerDirective.js
+++ b/src/components/layermanager/LayermanagerDirective.js
@@ -288,8 +288,27 @@
           });
         });
 
+        var removeNonTopicLayers = function(topicId) {
+          // Assemble first to not remove from the iterated over array
+          var layersToRemove = [];
+          scope.map.getLayers().forEach(function(olLayer) {
+            var l = gaLayers.getLayer(olLayer.bodId);
+            var regex = new RegExp('(^|,)(ech|' + topicId + ')(,|$)', 'g');
+            if (l &&
+                l.topics &&
+                !regex.test(l.topics) &&
+                !olLayer.background) {
+              layersToRemove.push(olLayer);
+            }
+          });
+          layersToRemove.forEach(function(olLayer) {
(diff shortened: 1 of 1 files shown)


src/components/layermanager/LayermanagerDirective.js
```diff
@@ -288,8 +288,27 @@
...
 
+        var removeNonTopicLayers = function(topicId) {
+          // Assemble first to not remove from the iterated over array
+          var layersToRemove = [];
+          scope.map.getLayers().forEach(function(olLayer) {
+            var l = gaLayers.getLayer(olLayer.bodId);
+            if (l &&
+                l.topics &&
+                l.topics.indexOf('ech') < 0 && //'ech' layers always stay
+                l.topics.indexOf(topicId) < 0 &&
```
oterral: may be a regex `/(^|,)(ech|kgs)(,|$) /g`  could be nice

gjn: +1

oterral:
Minor comment otherwise lgtm.

gjn:
I've adapted to use the RegExp. @oterral please give a last review and merge if good.

PR #2093, opened 2015-01-19
Use triggerHandler instead of focus function

### LINKED ISSUE #2090: Swissearch permalink parameter doesn't work anymore
https://map.geo.admin.ch/?swisssearch=Berges%2037%20pay

oterral:
Fix #2090 

http://mf-geoadmin3.dev.bgdi.ch/teo_fix_2090/?lang=en&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&X=185240.00&Y=561289.00&zoom=10&swisssearch=Berges%2037%20pay

### DIFF
--- a/src/components/search/SearchDirective.js
+++ b/src/components/search/SearchDirective.js
@@ -748,7 +748,9 @@
             if (!gaBrowserSniffer.mobile ||
                 (angular.isDefined(searchParam) &&
                 searchParam.length > 0)) {
-              taElt.focus();
+              // According to jquery doc (http://api.jquery.com/focus/)
+              // we should use triggerHandler when an element if hidden
+              taElt.triggerHandler('focus');
             }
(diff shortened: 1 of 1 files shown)


gjn:
This was not working even before we introduced the cloaking. @oterral Do you know what introduced this changed behaviour?

Otherwise, LGTM. Thanks for the quick fix.

oterral:
are you sure?  that's working when I remove the translate-cloak directive. The translate-cloak was introduced [here](https://github.com/geoadmin/mf-geoadmin3/commit/1a359509da9c850e32890b808f9e1706580ce5a1)

gjn:
Not sure. I tried  r_141218, which didn't work. You commit though should be part of r_141218. I assuemd the cloaking was introduced just last week.

PR #2104, opened 2015-01-22
Adding swisssearch parameter e2e tests

gjn:
To avoid regressions like https://github.com/geoadmin/mf-geoadmin3/issues/2090

### DIFF
--- a/test/selenium/swisssearch_test.js
+++ b/test/selenium/swisssearch_test.js
@@ -1,0 +1,48 @@
+
+var webdriver = require('browserstack-webdriver');
+var assert = require('assert');
+
+var QUERYSTRING_OF_RARON = "X=128180.88&Y=630227.13";
+var QUERYSTRING_MOOS = "X=128630.00&Y=627650.00";
+
+var runTest = function(cap, driver, target){
+  //swissearch parameter with multiple results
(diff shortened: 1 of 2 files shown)


ponceta:
Nice, I would add as comment // 03 Swisssearch test using browserstack (on line 01)
To recall the tests order

gjn:
I think that's not ideal for maintenance. Imagine adding a test that needs to be run first, then you a) need to know that you need to adapt those comments in all other files and b) need to adapt the comments in every other file.

Order of tests is visible here: https://github.com/geoadmin/mf-geoadmin3/blob/master/test/selenium/tests.js#L49

And if there are conditions about a given order (x has to run before y), than we should add this as a comment there. Or if you want to have a comment in the test itself, put the pre-requisite as a comment to the runTest command definition.

ponceta:
Seems legit, I miss some experience with maintenance.

In that case we just have to ensure that start_test.js is the first one and print_test.js is the last one (for stability purposes)

gjn:
Yes, exactly.
In any case, code is the best comment :)

ponceta:
All is well, I tested it 4 times in between friday and today. The only fail that can happen is a freeze of the print server on dev (like now) But not related to this PR.

image

gjn:
Thanks alot for the review and repeated tests.

PR #2106, opened 2015-01-22
Query tool improvment

oterral:
This improves the design and allow combination of bbox and attributes search.

[Test](http://mf-geoadmin3.dev.bgdi.ch/teo_query_imp/?topic=aviation)


src/components/query/QueryDirective.js
```diff
@@ -254,22 +255,37 @@
...
         $scope.loading = false;
         return;
       }
-      var lang = $translate.use();
+
+      if (!$scope.useBbox) {
+        $scope.geometry = null;
+      }
+
+      var common = {
```
gjn: We could move 270-274 to 265 and then merge both if-statements on 266 and 275. I think it would increase readability.

src/components/query/QueryDirective.js
```diff
@@ -254,22 +255,37 @@
...
+        lang: $translate.use(),
+        geometryFormat: 'geojson'
+      };
+
+      if (!$scope.useBbox) {
+        $scope.hideBox();
+      } else {
+        $scope.showBox();
```
gjn: theoretically, scope.geometry could be undefined/null here, no?

cedricmoullet:
Very nice. I would put the "query by bbox" before the search button

On Thu, Jan 22, 2015 at 2:00 PM, oterral <notifications@github.com> wrote:

> This improves the design adn allow combination of bbox and attributes
> search.
>
> Test <http://mf-geoadmin3.dev.bgdi.ch/teo_query_imp/?topic=aviation>
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/geoadmin/mf-geoadmin3/pull/2106
> Commit Summary
>
>    - Add filters in a collapsed element
>    - Update translations
>
> File Changes
>
>    - *M* src/components/query/partials/query.html
>    <https://github.com/geoadmin/mf-geoadmin3/pull/2106/files#diff-0>
>    (121)
>    - *M* src/components/query/style/query.less
>    <https://github.com/geoadmin/mf-geoadmin3/pull/2106/files#diff-1> (25)
>    - *M* src/locales/de.json
>    <https://github.com/geoadmin/mf-geoadmin3/pull/2106/files#diff-2> (10)
>    - *M* src/locales/empty.json
>    <https://github.com/geoadmin/mf-geoadmin3/pull/2106/files#diff-3> (2)
>    - *M* src/locales/en.json
>    <https://github.com/geoadmin/mf-geoadmin3/pull/2106/files#diff-4> (10)
>    - *M* src/locales/fr.json
>    <https://github.com/geoadmin/mf-geoadmin3/pull/2106/files#diff-5> (10)
>    - *M* src/locales/it.json
>    <https://github.com/geoadmin/mf-geoadmin3/pull/2106/files#diff-6> (10)
>    - *M* src/locales/rm.json
>    <https://github.com/geoadmin/mf-geoadmin3/pull/2106/files#diff-7> (12)
>
> Patch Links:
>
>    - https://github.com/geoadmin/mf-geoadmin3/pull/2106.patch
>    - https://github.com/geoadmin/mf-geoadmin3/pull/2106.diff
>
> —
> <https://github.com/geoadmin/mf-geoadmin3/pull/2106>.
>



-- 

gjn:
+1 for cedrics comment.

1) Also, there seems to be a regression when having a custom filter. Before, it was possible to reset a condition (cross icon), now it's not possible anymore and it's hard to reset a condition.

2) Maybe it would also make sense to have a check-box for each condition to toggle like the rectangle.

3) Drawing a rectangle auto-launches a query. All other actions don't auto-launch (you have to press search buttons), even disabling/enabling the bbox checkbox. This feels awkward to me.

Code looks good to me.

oterral:
```
Very nice. I would put the "query by bbox" before the search button`
```
Done

```
1) Also, there seems to be a regression when having a custom filter. Before, it was possible to reset a condition (cross icon), now it's not possible anymore and it's hard to reset a condition.
```
I don't see what you mean it's the same as before 

```
2) Maybe it would also make sense to have a check-box for each condition to toggle like the rectangle.
```
Good idea. 

```
3) Drawing a rectangle auto-launches a query. All other actions don't auto-launch (you have to press search buttons), even disabling/enabling the bbox checkbox. This feels awkward to me.
 ```

It depends. If you want to modify the bbox checkbox then  modify a condition, I think it's annoying to see the list of results to appear before you finish your modification.

File: src/components/query/QueryDirective.js (hunk 2 above)
Comment (oterral) [inline src/components/query/QueryDirective.js]: if $scop.useBbox is true scope.geometry can't bu null.

File: src/components/query/QueryDirective.js (hunk 1 above)
Comment (oterral) [inline src/components/query/QueryDirective.js]: yes I fix.

**Comment by gjn** [comment]
1) Sorry, it's the same as in prod, so no regression. But it's pretty confusing in the upper right corner of a condition. Sometimes you have [copy,refresh] icons, sometimes you have [refresh, reset] icons. Any reason why we can't have reset icon all the time?

3) Agree when modifying condition. But if you do 2), you could auto-search when checkbox is clicked.

File: src/components/query/QueryDirective.js (hunk 2 above)
Comment (gjn) [inline src/components/query/QueryDirective.js]: Yes, but it's not obvious...

File: src/components/query/QueryDirective.js [hunk 3]
```diff
@@ -480,25 +496,38 @@
...
           });
           dragBox.on('boxend', function(evt) {
-            scope.isActive = true;
             boxFeature.setGeometry(evt.target.getGeometry());
             scope.geometry = boxFeature.getGeometry();
+            scope.useBbox = true;
+
             if (scope.selectByRectangleLayers.length == 0) {
+              scope.isActive = true;
+              scope.useBbox = true;
```
gjn: It's already set to true above, so this line could probably be removed.

src/components/query/QueryDirective.js
oterral: I agree I will add a comment

src/components/query/QueryDirective.js
oterral: yes I remove it

gjn:
Found another thing.

- bazl layer
- open tool via tools menu
- draw rectangle
- launch search with pre-defined filter -> search as expected
- disable use_bbox checkbox
- launch search -> search as expected

But use_bbox checkbox disappears and helper text is shown again. I think the checkbox should stay once the rectangle is drawn.

oterral:
I agree with you but  if we uncheck the bbox  and you don't relaunch the search you have a list of results (requested with a bbox)  without a bbox drawn. I think it's disturbing.

Maybe when we open the filter part of the accordion we should remove the features and display the bbox only if the checkbox is selected then when we open the results part of the accordion we display features and bbox only if the bbox is used for the current query) .

What do you think?

gjn:
Maybe a little mis-understanding. I'll show you.

gjn:
@oterral Travis failure? Or needs rebase?

oterral:
one translation is missing

gjn:
`query_with_bbox` not translated at all?

gjn:
Good improvement. Missing translations, otherwise we can merge/deploy this today.

Still not happy with the UI and interaction; but I'm aware that it's not easy. I would like to have a dirty indicator when query/selectbbox changes to notify somehow that what is defined does not correspond to the current list of results.

oterral:
Read to merge @gjn

gjn:
Thanks alot!

PR #2108, opened 2015-01-22
Don't remove preview features when a preview layer is added

### LINKED ISSUE #1202: LUBIS: if time selector is active: Enter key is closing
ALL OS ALL BROWSERS

OPen https://mf-geoadmin3.int.bgdi.ch/lubis_demo/

activate time move slider by mouse, place mouse on map

Hit enter button

result
timeslider dissapears

expected result:
timeslider does not dissapear

oterral:
Fix #1202 

[test](http://mf-geoadmin3.dev.bgdi.ch/teo_fix_2102/)

### DIFF
--- a/src/components/map/MapService.js
+++ b/src/components/map/MapService.js
@@ -1660,7 +1660,7 @@
         this.removeAll = function(map) {
           var layers = map.getLayers().getArray();
           for (var i = 0; i < layers.length; i++) {
-            if (layers[i].preview) {
+            if (layers[i].preview && !(layers[i] instanceof ol.layer.Vector)) {
               map.removeLayer(layers[i]);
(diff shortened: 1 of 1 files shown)


gjn:
This fixes the issue, but it won't hold up when we will have GeoJSON layers (which will be of type layer) as our infra (Loic is working on that) that can be in catalog and previewed this way. We can merge this but we should thinkg about that then.

Alternatively, we could probably just not set preview to the previewFeatures layers here: https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/map/MapService.js#L1374, but not sure about that.

gjn:
ok, I'll merge this and try to keep it in mind :)

PR #2111, opened 2015-01-23
Add coherent coordinates conversion to mouseposition

### LINKED ISSUE #2107: WGS84 Coords wrong in footer
open map.geo.admin.ch

" WGS84 Koordinaten vom Standort,. Nicht grad/min/sek.
Diese sind im readout verdreht angegeben. Siehe attach rot markiert. lat/long 9.54, 46.84 für Chur. Bei Grad Min Sek stimmt's."
unbenannt

ponceta:
Fix #2107

Test [here](http://mf-geoadmin3.dev.bgdi.ch/ltpoa/)

### DIFF
--- a/src/js/MousePositionController.js
+++ b/src/js/MousePositionController.js
@@ -32,7 +32,7 @@
           label: 'WGS 84 (lat/lon)',
           format: function(coordinates) {
             return ol.coordinate.toStringHDMS(coordinates) +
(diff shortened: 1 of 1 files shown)


cedricmoullet:
We already got this discussion 1 year ago. Both are wrong.... or both are correct.
So, fine for me...

PR #2112, opened 2015-01-23
Remove form from index page (https related)

gjn:
When running our site with https through [1], it's not considered fully secure because we have an insecure form in the index file.

This PR removes the insecure from element and replaces it by div, getting rid of this error. We need this to not have an exclamation mark in the browser when accessing the site through https.

Does anyone know why we had a from in the first place?

[1] https://www.whynopadlock.com/ which allows to test any site for insecure elements.

### DIFF
--- a/src/index.mako.html
+++ b/src/index.mako.html
@@ -151,7 +151,7 @@ itemscope itemtype="http://schema.org/WebApplication"
 % endif      
       </div>
 % if device == 'desktop':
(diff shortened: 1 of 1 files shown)


cedricmoullet:
no idea...

gjn:
@oterral Care to quickly review/test?

oterral:
I think it's good, I've already tested

Issue #2115, opened 2015-01-26
[WIP] [Testing Mobile] Browserstack


ponceta:
@gjn I'll create the mobile tests this week, but since it needs a few parameters different from the standard browsers, I suggest to create it separately from the actual testing machinery. Last time I tried this, it was a bit unstable so I think it's another reason to have it separated from the actual testing. 
(and maybe merge both in the future if it comes more stable)

ponceta:
@gjn @oterral is it possible to have a DOM indicator after the build.js has been loaded for e2e on mobile? Or what is the last element which is loaded? Because I have troubles with uncompleted app when testing on mobile.

![image](https://s3.amazonaws.com/testautomation/08cfc8e9dd07f7df96332d9df93790b2ed35c2ed/screenshot-5.jpeg)

It could be great to have something like https://github.com/geoadmin/mf-geoadmin3/blob/0a187bb55024342427a31bc95e863a1510086411/src/components/print/partials/print.html#L44-L45

gjn:
The catalog is one of the last, as in desktop (see start_test.js). Might be
on mobile, the element is invisible, but it should be in the DOM.

I'm sure there are others that could be used. Inspect the DOM on mobile
after the page is loaded. (There are lots of ivisible items)
Am 26.01.2015 10:22 schrieb "Arnaud Poncet-Montanges" <
notifications@github.com>:

oterral:
With the  version in master if the loader element (the red bar) is hidden , that means the page is loaded.

ponceta:
Yes, that's why I wanted to check if loader is hidden before beginning tests. But for that I need to find an empty DOM element that says that like we did for print.

oterral:
https://github.com/geoadmin/mf-geoadmin3/blob/master/src/index.mako.html#L199  is it not enough?

gjn:
IAs said, inspect the DOM. Check for attributes that change when page is
loaded. I'm sure you can use the catalog or loader element for this. The
loader likely adds an attribute when done, or you could check for
visibility or whatever.

We should avoid adding DOM elements for testing unless absolutely
necessary, as was the case for print.
Am 26.01.2015 11:36 schrieb "Arnaud Poncet-Montanges" <
notifications@github.com>:

ponceta:
![iphone5](https://cloud.githubusercontent.com/assets/8614223/5915135/3e21b994-a602-11e4-8b71-e150683b1eb7.jpeg)
 
[...]

PR #2116, opened 2015-01-26
[WIP] Add mobile tests to browserstack testing

ponceta:
See #2115 for details

WIP do not merge

### DIFF
--- a/test/selenium/browsers.js
+++ b/test/selenium/browsers.js
@@ -62,7 +62,7 @@ var mobileCapabilities = [
     createMobile('android', 'ANDROID', 'Google Nexus'),
     createMobile('android', 'ANDROID', 'Google Nexus 4'),
     createMobile('android', 'ANDROID', 'Google Nexus 5'),
-    createMobile('android', 'ANDROID', 'Google Nexus 6')
+//    createMobile('android', 'ANDROID', 'Google Nexus 6')
 ];
 
 module.exports.capabilities = capabilities;
--- a/test/selenium/browsers.js
+++ b/test/selenium/browsers.js
@@ -24,7 +24,19 @@ var create = function(os, osver, res, browser, brover) {
     'browserstack.bfcache' : '0',
     'browserstack.debug' : false // switch this to true to debug (visual logs) 
   };
-}
+};
+
+var createMobile = function(browserName, platform, device) {
+  return {
+    'browserName' : browserName,
+    'platform' : platform,
+    'device' : device,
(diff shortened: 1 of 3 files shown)


ponceta:
Still not stable, especially for iPad... Search fails often.

ponceta:
@gjn what do you think of this?

This actually works, takes a bit of time, but it could be great to assess a bit it's stability running with jenkins for example during a week and see then if we go into more complexity or if we avoid e2e emulated mobile testing.

gjn:
I like to have them separated, but maybe we can still keep the general framework? But the tests and caps in desktop/mobile collections and run each once?

ponceta:
Yes, that's exactly what I was meaning with "running with jenkins"

gjn:
The code is separated here and contains thus duplicated definitions (mobile_browsers.js/browser.js and mobilel_tests.js/tests.js share alot of same code) I'm not a fan of that.

We could separate the mobile tests from desktop tests by having
a) mobile and desktop capabilities
b) mobile and desktop tests.
in a simple collection.

That's what I meant by collections. Something like
'''
var capabilities = {
   mobile: [...],
   desktop: [...]
}

var testsToRun = {
   mobile: [....],
   desktop: [....]
}
'''

The rest of the framework won't need much adaption and we would avoid duplication of code. It would also allow to run same tests on mobile and desktop in future (can be a possibility).

ponceta:
Like a "-m" for mobile tests, I'm pretty fan of this. I'm convinced this is a good idea. 

But if mobile tests are not stable enough this will only be code corruption.

Shouldn't we test it for a week and see if it is stable enough and adapt it as "definitive code" afterwards? As it is for know, it has no influence at all on the existing code and can be removed easily.

gjn:
For testing a week, you can setup a new jenkins projects, make it dependant on existing geoadmin3 project, checkout this branch and run the mobile tests

ponceta:
This is tested on jenkins for a week or two to assess its stability

ponceta:
jenkins test

ponceta:
I merged both mobile and desktop together.

Mobile tests are quite stable at night which is great for our night runs but are less stable on day (when there's more trafic)

I'll adapt the jenkins temp test for this new version.

IMO it could be great to select with a -m if we want mobile testing and ommiting it if we don't!

@gjn what do you think?

ponceta:
jenkins test

ponceta:
jenkins test

ponceta:
jenkins test

PR #2120, opened 2015-01-28
Teo fix kml

oterral:
fix kml

I'm testing on IE11 mobile

Works on chrome and safari but not on IE mobile but it's an appcache problem


gjn:
Testing of FF mobile

oterral:
@gjn good to merge

gjn:
Thanks alot!

PR #2129, opened 2015-01-29
Cancel the print process

### LINKED ISSUE #2117: Add cancel button to stop print process
Like the offline component , the print button should be transformed in a cancel button when the print progress begins.

oterral:
Fix #2117

[Test](http://mf-geoadmin3.dev.bgdi.ch/teo_cancel/);

In the case a multiprint that send a cancel request to the print server.

-  [x] chsdi needs to be modify to handle the cancel request

### DIFF
--- a/src/components/offline/partials/offline-selector.html
+++ b/src/components/offline/partials/offline-selector.html
@@ -12,6 +12,6 @@
           translate>offline_save</button>
   <button ng-show="isDownloading"class="btn btn-danger" 
           ng-click="abort()" 
-          translate>offline_abort</button>
+          translate>abort</button>
 </div>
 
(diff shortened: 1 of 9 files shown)


gjn:
FYI, there won't be a possibility to cancel a multiprint serverside because of multiple back-ends and cookie less state.

We could work around that using zadara back-end (create and check cancel file on shared volume), but I'm not sure it's worth the pain (given how rarelry the multiprint is used).

procrastinatio:
@gjn I agree it is not very used, but with this PR, a single user may start, stop, change his mind and restart  print jobs and destroy the server.

gjn:
I'll create a PR tomorrow for back-end.

gjn:
PR is back-end is ready. https://github.com/geoadmin/mf-chsdi3/pull/1258

I've tested it with this branch and it works well. On deploy, also needs testing on int (because of multi-machine setup)

@procrastinatio @oterral Please test/review as well.

oterral:
@procrastinatio can you review  pls?

procrastinatio:
Looks good for this part, I am testing the backend.

oterral:
thx @procrastinatio

PR #2134, opened 2015-02-02
Remove opacity effect on hover for touch screen

### LINKED ISSUE #2133: ga-btn on mobiles devices stay on disabled mode after clicking on it
ga-btn on mobiles devices are stay on disabled mode after clicking on it. They only come back to initial state after an event such as touch the map.

geolocation_disabled

oterral:
FIx #2133


loicgasser:
LGTM

PR #2136, opened 2015-02-02
Fix close button css for popovers and modal

oterral:
The close button of homescreen popovers, context-popup and modal offline menu were not correctly aligned, this PR fixes this applying the same css as popups.

[Test](http://mf-geoadmin3.dev.bgdi.ch/teo_close/)

Before:

before

After:
after

TEst on desktop/mobile IE11, FF, defaut browser, chrome


procrastinatio:
Very nice.

PR #2142, opened 2015-02-06
Remove chat tool on dev mode

loicgasser:
Because we don't need it when developping...


gjn:
I would keep it, otherwise we might miss some interaction problems when it's only there in prod mode.

If you don't like it at all, you could a) add user-specific variable or b) use a user.js scriopt (greasemonkey, tamperkmonkey) in your browser to get rid of it.

loicgasser:
ok makes sense

Issue #2094, opened 2015-01-19
[query tool] integrate a warning text, if number of result is more than max (200 currently)


mariokeusen:
UC: open app, switch to topic aviation, search for all objects from the layer ch.bazl.luftfahrthindernis with "maximum heighth AGL" >= 100.
Result: 200 results in list, in reality there would be more results. No warning for the user
Expected: if the number of results is more than the maximum of now 200, a warning text appears like in this screenshot:
![warning_query_results](https://cloud.githubusercontent.com/assets/4577749/5802636/19c5f3f0-9ff7-11e4-9000-c14cf7e879a6.PNG)

My proposal for DE:
"ACHTUNG: Zu grosse Anzahl Ergebnisse. Es werden nur die ersten 200 Ergebnisse angezeigt!"

My proposal for EN: 
"ATTENTION: Too large number of results. Only the first 200 results are being displayed!"

Open for some other proposals and also versions in FR and IT.

Thanks in advance
Mario

cedricmoullet:
@oterral @gjn How can we solve this ?

oterral:
Can we close this one @mariokeusen ? problem has been fixed differently in the referenced PR

mariokeusen:
IMHO yes. Do you have the result on dev?

oterral:
yes [here](http://mf-geoadmin3.dev.bgdi.ch/teo_results/?X=190000.00&Y=660000.00&zoom=1&lang=fr&topic=aviation&bgLayer=ch.swisstopo.pixelkarte-grau&catalogNodes=1379,1381,1514&layers=ch.bazl.luftfahrthindernis)

mariokeusen:
Thanks, I tested a bit:

In general I think it's a good solution and we can put it on prod. As soon as we are I'll take care about updating the help, which seems important to me.

Just found a strange thing on IE10:
![query_ie10_highlight_issue](https://cloud.githubusercontent.com/assets/4577749/6409641/da72a048-be66-11e4-9b9f-bc7cddabcb36.PNG)
It seems that the highlight does only display line objects but no points (performance issue?). Anyway it works very well on Chrome and FF. (I always used the query [Maximum height AGL m] >= 120, which produces 455 results in total)

Another thought: shouldn't the "Zustand"/"Conditions" be opened, when the user chooses a "custom filter" in the dropdown above? In this case he will anyway want to adjust things there and we could save a click with that.
[...]

PR #2143, opened 2015-02-09
Add link to display all the possible features

### LINKED ISSUE #2087: [query tool] get rid of maximum number of results by introducing a "load more results" button
UC: open app, load layer ch.bazl.luftfahrthindernis, search for all objects with AGL altitude > 100m
Result: the first 200 results are listed and highlighted. In reality there are more results on that query and the user has no chance to access the further results
Expected: first 200 results are listed and highlighted and the user can load the previous/next 200 results with a paging (like in old mf-bazl)

To be discussed:
But according to the feedbacks this will be the main improvement bazl will want in the meeting next week
IMHO this is valuable for all other layers which will be queried
Do you see any technical difficulties?

Thanks for your 2 cents

oterral:
Fix #2087 , #2094 

[Test](http://mf-geoadmin3.dev.bgdi.ch/teo_results/?X=190000.00&Y=660000.00&zoom=1&lang=fr&topic=aviation&bgLayer=ch.swisstopo.pixelkarte-grau&catalogNodes=1379,1381,1514&layers=ch.bazl.luftfahrthindernis)

Work with https://github.com/geoadmin/mf-chsdi3/pull/1266

### DIFF
--- a/src/components/query/QueryDirective.js
+++ b/src/components/query/QueryDirective.js
@@ -21,7 +21,7 @@
     $scope.selectByRectangleLayers = [];
     $scope.queriesPredef = [];
     $scope.filters = [];
-
+    $scope.featuresByLayer = {};
     // return the year of the timestamp
     var getYear = function(time) {
       return (time && time.substr(0, 4) != '9999') ?
@@ -87,7 +87,7 @@
     // When the list or value of filters change
     $scope.onChange = function() {
       $scope.queryPredef = null;
-      $scope.options.features = [];
+      $scope.featuresByLayer = {};
     };
 
     // Load attributes of the selected layer in the select box
@@ -176,30 +176,6 @@
       }
     };
 
-    // Search by geometry using the search server
-    /*$scope.searchByGeometry = function() {
-      $scope.loading = true;
-      $scope.queryType = 0;
-
-      if ($scope.geometry) {
-        var geom = $scope.geometry.getExtent();
-        var features = [];
-        gaQuery.getLayersFeaturesByBbox($scope,
-            $scope.selectByRectangleLayers,
-            geom
-        ).then(function(layerFeatures) {
-          features = features.concat(layerFeatures);
-          $scope.options.features = features;
-          $scope.loading = false;
-        },function(reason) {
-          $scope.options.features = [];
-          $scope.loading = false;
-        });
-      } else {
-        $scope.loading = false;
-      }
-    };*/
-
     var getGeometryParams = function() {
       var imgDisplay = $scope.map.getSize().concat([96]).join(',');
       var mapExtent = $scope.map.getView().calculateExtent(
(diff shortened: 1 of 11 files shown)


oterral:
Branch updated with the more results button.

If the branch link doesn't work test on my home directory http://mf-geoadmin3.dev.bgdi.ch/ltteo/

oterral:
I think it's good now. @mariokeusen @davidoesch @cedricmoullet ? what do you think ?

davidoesch:
it is now a paging -- user can *view* more than 50 results. The issue comes
when I hit printing ( print icon in the query result) here more than 50 are
shown...and printable.as soon as we have the csv export: we should only
allow 50 for export and the use a real paging.
If someone wants to download all the data: use data.geo.admin.ch

mariokeusen:
Agree with David's comment. 
In addition what is important for me: the user should get the total number of results displayed on top of the window in a way it gets obvious that his query produces more than 200 results. 
E.g. in the top text: "objets (200/279)

gjn:
To know the total results is a performance issue. I suggest that we do some performance tests and decide upon them if we

1) show "objects 200/total"
2) show "objects 200, show more"

2) is certainly faster, potentially much faster.

mariokeusen:
Good for me, let's test and decide then.

davidoesch:
+1

gjn:
I think the indication that we have more than 200 results should be in a) objects tab header and b) layer entry (maybe bold text, or asterisk next to 200?)

It's cumbersome to scroll to the button to see if there are more results.

src/components/query/QueryDirective.js
```diff
@@ -210,71 +186,100 @@
...
+    };
+
     // Search by geometry using feature identify servioce
-    $scope.searchByGeometry = function() {
+    $scope.searchByGeometry = function(layerBodId, offset) {
       $scope.queryType = 0;
 
       if ($scope.selectByRectangleLayers.length == 0) {
-        $scope.options.features = [];
+        onSearchError();
```
gjn: Not really an error here, is it? maybe call the function resetResults()?

src/components/query/QueryDirective.js
```diff
@@ -210,71 +186,100 @@
...
       $scope.queryType = 1;
       $scope.loading = true;
 
-      var features = [];
       var params = getParamsByLayer($scope.filters);
       if (params.length == 0) {
         if ($scope.useBbox) {
           $scope.searchByGeometry();
+        } else {
+          onSearchError();
```
gjn: Same as above. Not really an error.

gjn:
Usability comments and a few minor code comments. Otherwise, code looks very well.

I would really update the UI to better indicate that there are more results.

davidoesch:
+1 for Gilbert UI comments..(a) objects tab header and b) layer entry (maybe bold text, or asterisk next to 200)

Another proposal : if more than xx results available : the first line in the results list contains the text " first  xxx results of zzz shown, scroll down for more @

gjn:
As mentioend in previous comment, the 'of zzz' has performance implications...

davidoesch:
"only first  xxx results shown, scroll down for more "

src/components/query/QueryDirective.js
oterral: Yes better name

oterral:
Can we just add a`+` before the 200 (accordion header and layer header) ?

gjn:
yes, I think that would be enough.

oterral:
PR upadted with the `+` stuff

gjn:
That's good. Please squash and merge.

oterral:
thx for the review

davidoesch:
nice!

Issue #2148, opened 2015-02-12
Tests fail on a regular basis


gjn:
Our mf-geoadmin3 tests nightly tests fail once out of 2 times. This renders the tests useless, because everybody starts to ignore them. We have to make them more robust.

If found out that most of the time, the print test fails on a safari/osx combination. The cases I look, it was always [1] that failed, but please analyze a little further. Please discuss on how to get those tests more stable by either:
- remove print tests on safari/osx
- make those tests more stable on safari/osx

Also ping @daguer and @ponceta 

[1] https://github.com/geoadmin/mf-geoadmin3/blob/master/test/selenium/print_test.js#L9

ponceta:
I propose to avoid print test on safari, because even if safari driver is not the best, it should be stable enough to test the rest.

We should have a look a bit on how is map.geo.admin behaving on safari.

gjn:
You're giving up too fast ;). Not a fan of disabling the test alltogether for safari (only as last resort), because it's not systematic problem. It works sometimes, so there must be reasons why it fails sometimes.

a) tomcat is down on the dev instance. Print tests will always fail for any browser. I've added a project to jenkins that restarts tomcat before all other projects are tested (2:00 in the night). Let's see if this helps for the geoadmin3 jenkins job (we don't have this on map.geo.admin.ch jenkins job)

b) on safari, it seems the animation to open the print pane is taking longer than in other browsers. The error we get is alwas 'can't click element that is not visible'. We have this both in geoadmin3 jenkins job and map.geo.admin.ch jenkins job.

To solve b) we should try to to improve [1] and instead of checking for the aria-expanded property of the printHeading element (which is set at _start_ of animation), to check for the 'in' class of the print element (div with id=print), which is set _after_ the animation is complete. You can observe this in browser.

What do you think, @ponceta, @alcapat and @daguer (Maybe we can sit together tomorrow and I can better demonstrate what I mean)

[1] https://github.com/geoadmin/mf-geoadmin3/blob/master/test/selenium/print_test.js#L9

ponceta:
I agree. My solution was just a temporary one to buffer the test failing serie.

I'm not sure restarting tomcat just before testing @procrastinatio will probably agree. But maybe restart it 5 minutes before will indeed increase the stability.

The boring thing is that the click method doesn't wait, and as soon as the DOM element is found, the click is fired!

gjn:
Tomcat is restarted 22 minutes before :)

[...]

PR #2159, opened 2015-02-17
Take animation into account for print tests

gjn:
This addresses a part of https://github.com/geoadmin/mf-geoadmin3/issues/2148

It takes into account that the opening element is animated and waits until the animation is finished before clicking the print button. This should get rid of a couple of false positives from browserstack.


ponceta:
Tested, works fine! Can be merged!

gjn:
Thanks!

PR #2162, opened 2015-02-18
[Testing] Add kml to print

### LINKED ISSUE #2158: [Testing] KML print
We should add at least two KML layers to the print test to 

- [ ] One which is generated from map.geo.admin.ch

- [x] One which is a standard one coming from internet.

(One which is drawn in real time (need support of seleniumjs mousemove methods before this (see future enhancements)))

ponceta:
Fix #2158

@gjn I have issues with timeouts sometimes when the alert popup does not shows up early enough! Any ideas are welcome!

### DIFF
--- a/test/selenium/print_test.js
+++ b/test/selenium/print_test.js
@@ -3,6 +3,24 @@
 var webdriver = require('browserstack-webdriver');
 
 var runTest = function(cap, driver, target) {
+  // Goto the travis deployed site.
+  driver.get(target + '/?lang=de');
+  //wait until topics related stuff is loaded. We know this when catalog is there
+  driver.findElement(webdriver.By.xpath("//a[contains(text(), 'Grundlagen und Planung')]"));
+  // Click on "Werkzeuge"
+  driver.findElement(webdriver.By.xpath("//a[@id='toolsHeading']")).click();
+  // Click on "KML Import"
+  driver.findElement(webdriver.By.xpath("//*[contains(text(), 'KML Import')]")).click();
+  // Click on "URL"
+  driver.findElement(webdriver.By.xpath("//a[contains(text(), 'URL')]")).click();
+  // Write URL of the chosen KML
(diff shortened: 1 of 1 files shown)


gjn:
Could you tell me which approaches you already tried?

test/selenium/print_test.js
```diff
@@ -3,6 +3,12 @@
 var webdriver = require('browserstack-webdriver');
 
 var runTest = function(cap, driver, target) {
+  // Goto the travis deployed site.
+  driver.get(target + '/?lang=de&layers=KML%7C%7Chttp:%2F%2Fopendata.utou.ch%2Furbanproto%2Fgeneva%2Fgeo%2Fkml%2FRoutes.kml');
+  // Accept the third party data
```
gjn: Would it be possible to use the wait function here to check if an alert is there? Pseudo code:
```
driver.wait(function() {
    //check for exisance of alert (not sure how, verify), return true when found
    if (driver.switchTo().alert()) {
        return true;
    } else {
        return false;
    }
}, 3000).then(function() {
    driver.switchTo().alert().accept();
    ...

});
```

daguer:
jenkins test

daguer:
jenkins test

daguer:
jenkins test

daguer:
jenkins test

gjn:
Just wondering: did you try my proposal?

ponceta:
Not yet, I used the standard way a user would go through this which is more likely what we want to add as e2e test right?

This now works smoothly :) 
See [This](https://www.browserstack.com/automate/builds/dddb1242fb9f3ffe297b057e6da2ea964b4caf1a/sessions/490e97ea9dfa826b9977da546654973dbdbfdd86#automate_button) (visual logs were activated to test)

gjn:
That's fine with me. I was just wondering if my approach would have worked. Might be usefull when we add more tests that will have an alert. At least, we would know how to handle those.

ponceta:
It is always a valid use case to be tested (import with 3rd party data message) (on mobile it's the only solution)
I'll try that too on another PR

gjn:
Great!

gjn:
jenkins test

Issue #1912, opened 2014-12-01
Profile: enhancements


kandre:
Further profile enhancements:

- [ ]  the profile box width could be reduced when the main geoadmin panel is extended and going over the profile (Do we really want that?)
- [ ]  center transect to the middle of the map
- [ ] show the surface type based on geocover

- [x]  if only first line:
    - [x] show azimuth
    - [x] show distance
    - [x]  highest/lowest point

- [x] if only a line (more than one point):
    - [x]  show total m up
    - [x]  show total m down 
    - [x] show distance
    - [x]  highest/lowest point

- [x] if a polyline
    - [x] show total m up
    - [x] show total m down
    - [x] show distance
    - [x]  highest/lowest point

- [x] cross-browser compatibility

- [x] Allow Import of measured path as KML

- [x] Add slope parameter in profile tootip
- [ ] bring print back for profile
- [x] add hiking time for each direction

ponceta:
Add " Allow Import of measured path as KML " #1944 as possible enhancement

ponceta:
[Better Than Google]

![aletsch](https://cloud.githubusercontent.com/assets/8614223/7020269/fc7cd526-dd17-11e4-850f-54d82269cca0.PNG)
![aletsch_google](https://cloud.githubusercontent.com/assets/8614223/7020270/fc7d0410-dd17-11e4-8486-70a85a2afc47.PNG)

Possible Improvements:
 - [ ] Height / Length display everywhere on the graph (not only on the colored side)
 - [ ] Responsive design for the tooltip height (for high rez monitors)
 - [ ] Improve the Profile Print (Improving png rez before sending it to the print server)
Harder improvements
 - [ ] Improve Profile quality for bigger resolutions (either with minimal spacing or min max detection -> avoid as much as possible "smoothing")

kandre:
@oterral, I know you are refining drawing and measure tool are in progress, then I suggest I could develop functionnalites anyway (such as highest/lowest point, slope, denivelé...) from the computation point of view, and not spending time too much on graphical interface and user interactions, what do you think ?

cedricmoullet:
+1 with @kandre proposal.

ponceta:
Since we extract coordinates, we should add the 3rd one (altitude) to the generated kml points.

davidoesch:
surface type is still not solved... move this to a new ticket ?

gjn:
[...]

PR #2176, opened 2015-02-26
More compact measure tool, more space for profile

loicgasser:
No functional change. Not tested on IE.

[Test Link] (http://mf-geoadmin3.dev.bgdi.ch/gal_profile_style/)

*Before*:
profile_1_before
profile_2_before

*After*:
profile_1_after
profile_2_after

### DIFF
--- a/src/components/measure/MeasureDirective.js
+++ b/src/components/measure/MeasureDirective.js
@@ -87,6 +87,7 @@
             style: scope.options.drawStyleFunction
           });
 
+
           // Activate the component: add listeners, last features drawn and draw
           // interaction.
           var activate = function() {
@@ -94,6 +95,14 @@
             scope.map.addLayer(layer);
             scope.map.addInteraction(drawArea);
             featuresOverlay.setMap(scope.map);
+
+            // Controls the display of the show profile button
+            var nbDrawnFeatures = layer.getSource().getFeatures().length;
+            if (nbDrawnFeatures > 0) {
+              scope.isEmptyDrawing = false;
+            } else {
+              scope.isEmptyDrawing = true;
+            }
 
             // Add events
             deregister = [
@@ -105,7 +114,9 @@
               drawArea.on('drawstart', function(evt) {
                 var nbPoint = 1;
                 var isSnapOnLastPoint = false;
-
+                scope.$apply(function() {
+                  scope.isEmptyDrawing = true;
+                });
                 // Clear the layer
                 layer.getSource().clear();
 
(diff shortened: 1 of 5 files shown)


oterral:
Can't we let  buttons on the right but reduce the popup height to 200px ?

loicgasser:
I tested that as well looks really small... But yes I can provide an alternative PR and we can choose what seems to be the best approach. We could also reduce the buttons size. Seems really big.

loicgasser:
Or align btns horizontally.

loicgasser:
@oterral I changed it to 280px, I think we can go down to 230-240px if we align btns horizontally and make them a little smaller. I think we should leave as much space as possible on the right to display the profile.

oterral:
Yes button horizontally could be good.

loicgasser:
Updated, it is much less invasive.
@davidoesch what about download icon with the save btn. http://fortawesome.github.io/Font-Awesome/3.2.1/icon/download/

I saw that people thought they were saving the state of the measure during user tests. Or simply change the text.

oterral:
What about hiding the display profile button when there is no profile (or no line drawn)?

oterral:
It's nice with horizontal button. There is just a font problem for the save button.

davidoesch:
icon : nope - we have to be consistent- either we use icons everywhere or nowhere.
therefore : changing the text from sauvgarde to export might make sense

loicgasser:
I like the idea of not displaying the show profile btn when there is no line drawn.
I even prefer the idea of using the drawing line/polygon (from drawing tool) if already here.
@davidoesch ?

I'll change the text.

gjn:
Or maybe even removing the profile button altogether and just draw the profile when there's a line?

(Yeah, I like removing stuff)

davidoesch:
- not displaying the show profile btn when there is no line drawn. see
https://github.com/geoadmin/mf-geoadmin3/issues/1912
-"Or maybe even removing the profile butto just draw the profile when
there's a line " +1

kandre:
Really nice ! Tested on IE11 and IE9, works great

gjn:
jenkins test

loicgasser:
rebased, I implemented the display/hide profile button as discussed above.
Translations changes should be done in a separate PR as these have an impact beyond the profile tool and might require other changes.

oterral:
@loicgasser the button should appear when you create the 2nd point, not when you move the mouse after the first point.

davidoesch:
otherwise- nice one!

gjn:
LGTM.

@loicgasser Can you adapt to latest request?

loicgasser:
ok do it now

loicgasser:
ok adapted on second point.

gjn:
@oterral Please merge before/after the deploy. You decide.

oterral:
I will do after the deploy

oterral:
thx for this @loicgasser

Issue #2141, opened 2015-02-05
widgets=print


davidoesch:
Analog to http://map.geo.admin.ch/?widgets=feedback 

can you activate http://map.geo.admin.ch/?widgets=print ?

gjn:
Will be there on next deploy.

PR #2178, opened 2015-02-26
Support widgets=print parameter to init app with print panel open

gjn:
Fix for https://github.com/geoadmin/mf-geoadmin3/issues/2141

### DIFF
--- a/src/js/MainController.js
+++ b/src/js/MainController.js
@@ -102,6 +102,8 @@
       // Activate the "features" permalink manager for the map.
       gaFeaturesPermalinkManager($scope.map);
 
+      var initWithPrint = /print/g.test(gaPermalink.getParams().widgets);
+
       $rootScope.$on('gaTopicChange', function(event, topic) {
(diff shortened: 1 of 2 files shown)


src/js/MainController.js
```diff
@@ -114,7 +116,10 @@
             isWindowTooSmall()) {
           showCatalog = false;
         }
-        $scope.globals.catalogShown = showCatalog;
+        if (!initWithPrint) {
+          $scope.globals.catalogShown = showCatalog;
+          initWithPrint = false;
```
oterral: initWithPrint is already false here

oterral: On the next topi change the print widget stays open and the catalog is always closed. I don't know if it's wanted or not

gjn: The if is wrong...I'll adapt.

gjn:
@oterral It's corrected. Catalog change now opens catalog (and closes print).

oterral:
Perfect thx

PR #2179, opened 2015-02-26
Add BrowserSniffer tests for mac and ios

oterral:
This PR clarifies a little bit what we detect in BrowserSniffer and add test for iOS and Mac


gjn:
LGTM. Thanks alot.

PR #2186, opened 2015-03-03
Remove safari tests on Windows

gjn:
I bring out the axe and also remove the safari on windows tests...as @ponceta noted and the last jenkins tests https://jenkins.dev.bgdi.ch/job/geoadmin3/295/ show, this fails sometimes as well.

Self-Merge.


ponceta:
Magicians have also axes? Your skill tree should be amazing! :)

It is clear that the purpose of testing should not be to test the testing framework but the product itself so that's a good idea, since we/they find a solution to make it more stable!

PR #2196, opened 2015-03-05
Check if page is loaded before testing content

ponceta:
On CI, tests fails because page is not fully loaded. This should improve this behaviour.

### DIFF
--- a/test/selenium/swisssearch_test.js
+++ b/test/selenium/swisssearch_test.js
@@ -8,6 +8,8 @@ var QUERYSTRING_MOOS = "X=128630.00&Y=627650.00";
 var runTest = function(cap, driver, target){
   //swissearch parameter with multiple results
   driver.get(target + '/?swisssearch=raron&lang=de');
+  //wait until topics related stuff is loaded. We know this when catalog is there
+  driver.findElement(webdriver.By.xpath("//a[contains(text(), 'Grundlagen und Planung')]"));
   driver.findElement(webdriver.By.xpath("//*[contains(text(), 'Raron')]"));
   driver.findElement(webdriver.By.xpath("//*[contains(text(), 'Flugfeld')]")).click();
   driver.findElement(webdriver.By.xpath("//a[contains(@href, '" + QUERYSTRING_OF_RARON + "')]"));
@@ -18,6 +20,8 @@ var runTest = function(cap, driver, target){
 
   //swisssearch parameter with multiple results (locations and layers), reset selection
   driver.get(target + '/?swisssearch=wasser&lang=de');
+  //wait until topics related stuff is loaded. We know this when catalog is there
+  driver.findElement(webdriver.By.xpath("//a[contains(text(), 'Grundlagen und Planung')]"));
   driver.findElement(webdriver.By.xpath("//*[contains(text(), 'Standorte')]"));
   driver.findElement(webdriver.By.xpath("//*[contains(text(), 'Karten')]"));
   driver.findElement(webdriver.By.xpath("//i[@ng-click='clearInput()']")).click();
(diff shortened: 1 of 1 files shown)


gjn:
LGTM.

ponceta:
This is a quick merge, but I found an issue with it (the pr is fine). The CI test does not fetch the actual branch!

We need to correct this on jenkins or without it testing e2e is useless. @alcapat can you investigate this?

Another thing is crucial, print server is not active on CI, so testing print will also fail. @gjn I'll look at it with @procrastinatio

gjn:
@ponceta Where did you check if the CI fetches the actual branch?

ponceta:
https://www.browserstack.com/automate/builds/dddb1242fb9f3ffe297b057e6da2ea964b4caf1a/sessions/deaa42073afd35776b1a238e4228a96366a35ab7#automate_button

ponceta:
I checked that it does not fetch it :)

ponceta:
We should have

Find element xpath=//a[contains(text(), 'Grundlagen und Planung')]

After each get url command if it was the case (like specified in the PR)

gjn:
hmm. how did you check that it does not fetch it? In which directory?

ponceta:
/var/lib/jenkins/jobs/geoadmin3_pullrequests/workspace ?

gjn:
that's really strange.

gjn:
project is defined in the wrong way (see https://jenkins.ci.bgdi.ch/job/geoadmin3_pullrequests/configure)

At lease the `branches to build` setting shouldn't be `origin/master` but `${sha1}` as specified https://github.com/jenkinsci/ghprb-plugin, maybe other settings are wrong too. I'm going to adapt.

gjn:
jenkins test

gjn:
btw: I don't think the failure we have on jenkins.ci is fixed with this, because it was working before no dev for a couple of weeks.

In any case, I'm able now to reproduce this by hand using FF35: (On dev)[http://mf-geoadmin3.dev.bgdi.ch/?swisssearch=brückenmoosstrasse 11 raron&lang=de] it's working, (On ci)[http://mf-geoadmin3.ci.bgdi.ch/?swisssearch=brückenmoosstrasse 11 raron&lang=de] it doesnt.

@ponceta Could you try to find the reason?

gjn:
I know the reason, the geometry is not send with the result on ci environement, because https://gitweb.prod.bgdi.ch/gitweb/?p=puppetmaster-stable/.git;a=blob;f=site-modules/chtopo/files/etc/varnish/3.0.5/routing.development.vcl;hb=c433f105a82f137bb5c5b31f2f84b44f276998d7#l121 blocks https://gitweb.prod.bgdi.ch/gitweb/?p=puppetmaster-stable/.git;a=blob;f=site-modules/chtopo/files/etc/varnish/3.0.5/routing.development.vcl;hb=c433f105a82f137bb5c5b31f2f84b44f276998d7#l130

(the else if at line 121 should be removed)

@alcapat @ponceta Can you contact c2c for this?

gjn:
Please ignore the above. I had eye problems.

gjn:
We need to adapt https://gitweb.prod.bgdi.ch/gitweb/?p=puppetmaster-stable/.git;a=blob;f=site-modules/chtopo/files/etc/varnish/services-restricted-access.vcl.include;h=bd64ecf786cc77a836fc2a7561995e49721e80f8;hb=refs/heads/stable3 and add ci as valid referer. I have informed @procrastinatio .

ponceta:
ok

gjn:
@ponceta The swissearch tests are now working (Marc added the varnish config); but the print still is not workin on ci, so the print test fails. What's the status with print on ci?

ponceta:
Always the same, need /var/cache/print dir to be writable from the print server to cache pdfs (need c2c permission?)

ponceta:
Else the print is working fine, config is readable from CI

gjn:
Yes, the ci instance needs to attach the `/var/cache/print` directory to the zadara storage. We can use the same as for dev.
172.29.224.131:/export/geodata_dev/mapfish_print                 5732575232 2351263744 3381311488  42% /var/cache/print

@ponceta Can you open an issue with c2c for that?

PR #2201, opened 2015-03-06
Update to latest 3 browsers in browserstack

gjn:
All in title. See https://www.browserstack.com/list-of-browsers-and-platforms?product=automate for reference.


ponceta:
Nice, CI build is ok, can be merged.

Issue #1904, opened 2014-11-27
Slower profile computation recently


kandre:
It is a regression. The waiting cursor ![cursor](https://cloud.githubusercontent.com/assets/8038575/5214024/b1b969f4-7616-11e4-9ed7-60f063ca7769.gif) runs a longer time than before when a profile is queried. This is not related to 'improve profile: add to bottom of map' #1729 since this new issue is also on prod. Any idea?

PR #2203, opened 2015-03-06
Reduce the number of requested points in the profile

cedricmoullet:
Make the profile 4 times faster. Quality reduction is small, from my point of view.

Can be tested here: http://mf-geoadmin3.int.bgdi.ch/moc_profile_optimization/

@davidoesch @gjn your thoughts ?

### DIFF
--- a/src/js/MeasureController.js
+++ b/src/js/MeasureController.js
@@ -23,7 +23,8 @@
                 left: 60
               },
               height: 250,
-              elevationModel: 'COMB'
+              elevationModel: 'COMB',
+              nb_points: 50
           },
           styleFunction: (function() {
             var styles = {};
@@ -105,7 +106,8 @@
               angular.toJson(coordinates) + '}'; 
           var template = $scope.options.profileUrl + '?geom=' +
                gaUrlUtils.encodeUriQuery(wkt) + '&elevation_models=' +
-              $scope.options.profileOptions.elevationModel;
+              $scope.options.profileOptions.elevationModel + '&nb_points=' +
+              $scope.options.profileOptions.nb_points;
           var http = $http.get(template);
           if (!callback) {
             callback = function(data, status) {


gjn:
For longer lines, there is a visible difference.

New
image
Old
image

cedricmoullet:
Then we need probably to make this number length dependent.

davidoesch:
made several tests , and the result are really different

5km (Arosa, weisshorn arosa) smoothness : hardly any difference ,
computation time: a bit faster

20km (Arosa-kloster): smoothness is weaker, computation time: a bit faster
-

+200km (Lausanne Chur) : this results in a completely different (max vlaues
and min values are not the same... ) result! computation time .. it is
faster ... but not that much (3x?)


If we switch we will ahve an issue...


Question: why was it faster while having the same sample resolution in re2?

gjn:
I would guess reason is Zadara storage.

We could simulate 'progressive loading' by getting profile for 20, then 50, then 100, then 200 points...or adapting service to streaming progressively.

ponceta:
IMHO Profile consumers prefer to pay a bit of time for a better quality tool than an averaged approximated google like result (as google does)

The quality of our profile is also that it adapts to the consumer's needs : We do not need the same precision on a 100m track than on a 10km track (different usages). Swisstopo is owner of the best DHM of the whole Switzerland, it should be a shame if we can't display a high quality profile.

We could maybe make the user aware of the space between two measure points (to give a scale precision).

On a 200m track it's a 1m resolution

Actual (with 200 points right?)
image

With 50 points
image

Google
image

davidoesch:
Fully aggree... @gjn what about refactoring the profile service to a
"faster" backend architecture? (s3, nosql, all in mem) etc..

ltclm:
i also think the number of points should be length dependant.
dtm2/comb a point each 2m
dhm25 a point each 25 m
reducing the number of points / quality for performance reasons is not a good idea - this is an 
expert tool.

gjn:
S3 is not usable (we need direct filesystem access because formate requires seeks in multiple files, same holds for nosql). All in mem would be luxury and also _very_ expensive (we are speaking 60GB per instance for COMB model only).

To test different fs storage, I would need an instance with 60GB free of EBS and/or SSD storage (dev only has 30GB) to do some real performance tests. I would assume an SSD backed storage might be quite a bit faster but it's quite expensive too (and would need to be tested first); but maybe EBS backed volume is enough for our needs.

I can try to prepare a PR with progressive profile loading....

cedricmoullet:
I think it's worth trying to improve the backend.
Progressive loading seems a bit complicated to me.

On Tue, Mar 10, 2015 at 8:23 AM, Gilbert Jeiziner <notifications@github.com>
wrote:

> S3 is not usable (we need direct filesystem access because formate
> requires seeks in multiple files, same holds for nosql). All in mem would
> be luxury and also *very* expensive (we are speaking 60GB per instance
> for COMB model only).
>
> To test different fs storage, I would need an instance with 60GB free of
> EBS and/or SSD storage (dev only has 30GB) to do some real performance
> tests. I would assume an SSD backed storage might be quite a bit faster but
> it's quite expensive too (and would need to be tested first); but maybe EBS
> backed volume is enough for our needs.
>
> I can try to prepare a PR with progressive profile loading....
>
> —
> <https://github.com/geoadmin/mf-geoadmin3/pull/2203#issuecomment-78006554>
> .
>



-- 

<http://t.signaletre.com/e1t/c/5/f18dQhb0S7lC8dDMPbW2n0x6l2B9nMJN7t5XYgf6M1WN1qfVVMRYFRqW8qlQyH56dByMf1q8vLz02?t=http%3A%2F%2Ftwitter.com%2Fcedricmoullet&si=6057645868318720&pi=17c1539f-c604-4b3b-fbac-b818a5d60438>

ltclm:
we ( @kandre @loicgasser ) already did some perf tests with an extra 100GB SSD EBS  on the mf-** instances see [C2C #190267]. Loic prepared a pr http://mf-chsdi3.int.bgdi.ch/gal_profile_ssdebs/dev. 
We did not see any difference in the performance between zadara and ssd/ebs.
https://github.com/geoadmin/mf-geoadmin3/issues/1904#issuecomment-67459443

davidoesch:
again-- why was it faster in RE2? what is the different setting? (IMHO we
had there ec2 and EBS..)

gjn:
I can confirm that preformance seem similar using zadara or EBS (at least for DTM2 model, which is similar to COMB model).
I'll look at the implementation to see where we can probably improve performance.

gjn:
I couldn't believe the above and I dug (sp?) deeper. I think I finally found the flaws in my previous tests (and most probably the same effect was in place when @loicgasser and @kandre did their tests). I'll explain below; first I describe the tests done.

The tests were done using DTM2 on zadara [1] and DTM2 on EBS mf0dev [2] using [this code](https://github.com/geoadmin/mf-chsdi3/blob/gjn_profile_perf/chsdi/tests/integration/test_profile.py#L22) from [this branch](https://github.com/geoadmin/mf-chsdi3/tree/gjn_profile_perf) (which based on current master), then use command: `buildout/bin/nosetests chsdi/tests/integration/test_profile.py:TestProfileView.timing_test`. you can adapt [model used here](https://github.com/geoadmin/mf-chsdi3/blob/gjn_profile_perf/chsdi/lib/raster/georaster.py#L25); the only code change between test runs happened in those 2 lines.

Basically, I test the profile service 100 times using a random line (at least 200km long) for each run. I did this 3 times for zadara and 3 times for EBS volume. Note that there is no apache/modwsgi involved. We access pyramids app directly. The numbers fit the user experience we currently have.

Now here's the thing: the index file of the DTM (swissalti3d/kombo_2m_dhm25/index.shp) apparently contains **absolute pathes** to the tiles of the DTM (@ltclm do you know why this is?). This means, even when copied locally to some location, the tiles uses where still pointing to /var/local/cartoweb/...which is on zadara. So I [applied a patch](https://github.com/geoadmin/mf-chsdi3/blob/gjn_profile_perf/chsdi/lib/raster/georaster.py#L85) to use the local tiles, and puff, here are the results:

Zadara average times per profile request: 8.1s; 8.8s; 9.8s
EBS volume average times per profile request: 1.5s; 1.5s; 1.3s

EBS volume is at least 5 times faster than Zadara backed storage. Ouch. I don't know if we have magnetic EBS on mf0dev or SSD. We would need 100GB of EBS on each of our 3 prod instances (SSD or magnetic). @ltclm what would the price of those be? [3]

I seems clear that whenever we need real-time fast access to files, we can't use zadara.

[1] /var/local/cartoweb/bund/swisstopo/swissalti3d/2m/
[2] home/ltjeg/tmp/2m/index.shp
[3] http://aws.amazon.com/de/ebs/pricing/

cedricmoullet:
Excellent findings. Thanks

On Wed, Mar 11, 2015 at 1:35 PM, Gilbert Jeiziner <notifications@github.com>
wrote:

> I couldn't believe the above and I dug (sp?) deeper. I think I finally
> found the flaws in my previous tests (and most probably the same effect was
> in place when @loicgasser <https://github.com/loicgasser> and @kandre
> <https://github.com/kandre> did their tests). I'll explain below; first I
> describe the tests done.
>
> The tests were done using DTM2 on zadara [1] and DTM2 on EBS mf0dev [2]
> using this code
> <https://github.com/geoadmin/mf-chsdi3/blob/gjn_profile_perf/chsdi/tests/integration/test_profile.py#L22>
> from this branch
> <https://github.com/geoadmin/mf-chsdi3/tree/gjn_profile_perf> (which
> based on current master), then use command: buildout/bin/nosetests
> chsdi/tests/integration/test_profile.py:TestProfileView.timing_test. you
> can adapt model used here
> <https://github.com/geoadmin/mf-chsdi3/blob/gjn_profile_perf/chsdi/lib/raster/georaster.py#L25>;
> the only code change between test runs happened in those 2 lines.
>
> Basically, I test the profile service 100 times using a random line (at
> least 200km long) for each run. I did this 3 times for zadara and 3 times
> for EBS volume. Note that there is no apache/modwsgi involved. We access
> pyramids app directly. The numbers fit the user experience we currently
> have.
>
> Now here's the thing: the index file of the DTM
> (swissalti3d/kombo_2m_dhm25/index.shp) apparently contains *absolute
> pathes* to the tiles of the DTM (@ltclm <https://github.com/ltclm> do you
> know why this is?). This means, even when copied locally to some location,
> the tiles uses where still pointing to /var/local/cartoweb/...which is on
> zadara. So I applied a patch
> <https://github.com/geoadmin/mf-chsdi3/blob/gjn_profile_perf/chsdi/lib/raster/georaster.py#L85>
> to use the local tiles, and puff, here are the results:
>
> Zadara average times per profile request: 8.1s; 8.8s; 9.8s
> EBS volume average times per profile request: 1.5s; 1.5s; 1.3s
>
> EBS volume is at least 5 times faster than Zadara backed storage. Ouch. I
> don't know if we have magnetic EBS on mf0dev or SSD. We would need 100GB of
> EBS on each of our 3 prod instances (SSD or magnetic). @ltclm
> <https://github.com/ltclm> what would the price of those be? [3]
>
> I seems clear that whenever we need real-time fast access to files, we
> can't use zadara.
>
> [1] /var/local/cartoweb/bund/swisstopo/swissalti3d/2m/
> [2] home/ltjeg/tmp/2m/index.shp
> [3] http://aws.amazon.com/de/ebs/pricing/
>
> —
> <https://github.com/geoadmin/mf-geoadmin3/pull/2203#issuecomment-78254448>
> .
>



-- 

ltclm:
Thanks a lot @gjn for the forensic inquiries. Everything seems to be clear now!
We are always using absolute paths in the index files, because in the "mapserver world" the geodata path is stable. I will change the index file of the dtm data accordingly. And yes i totally overlooked this when preparing the test data for the last performance tests.
 
I think we should put the DTM2 (47GB) and COMB (56GB) on ssd, so we would need 150GB SSD.
Moreover i thing we should add this SSD EBS to the entire mf line (1 dev, 2 int and 2 prod).

cedricmoullet:
What is price of 150 gb ssd ?

Envoyé de mon iPhone

ltclm:
According to [1]
the prices in ireland for 1 GB / month are: 
* General Purpose (SSD): 0.11$

5 instances with 150GB each would be:  83 $ / month (16.5 $/instance)

[1] http://aws.amazon.com/de/ebs/pricing/

gjn:
Does that include IO costs (if there is)?

cedricmoullet:
Ok for me. Try to make the ssd as small as possible.

Envoyé de mon iPhone

hpchrist:
Hi all,

there would be another possibility as well: SSD drives on Zadara, but this will cost more than double than the EBS solution. So I propose as well to go for the EBS approach with SSH. Since the data is updated only once a year, the additional overhead to update the data on 5 EBS volumes (2*prod, 2*int, 1*dev) should be acceptable.

@ltclm: you can order the EBS volumes.

gjn:
Forget Zadara. Even with SSD they will be noticably slower. Zadara is not for big, real-time throughput.

hpchrist:
That is was I propose: use EBS with SSD

PR #2207, opened 2015-03-10
Use ol.Observable.unByKey static function to remove listeners

loicgasser:
As it is said in the doc, using `ol.Observalble.unByKey` is to be preffered.


gjn:
Thank you.

PR #2209, opened 2015-03-11
Update ol and add list of exports to optimize the build

oterral:
https://github.com/openlayers/ol3/commit/b6adeea4f4cb31fbd0976ad1d91e70c1c66f39c7

This PR needs intensive tests [here](http://mf-geoadmin3.dev.bgdi.ch/teo_optim_ol/?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&X=190000.00&Y=660000.00&zoom=1)

We gain 180ko on the ol.js file. That could be improve specifying all the method we use but I don't think  it's necessary. what do you think?


gjn:
180KB would be awesome.

PR #2211, opened 2015-03-11
Fix #2163: don't print contextpopup overlay

### LINKED ISSUE #2163: A marker is printed when you made a right  click  on the map
Steps to reproduce : 
 1) Right click somewhere on the map
 2) Print the zone where  you've made the right click

Results:
A marker is printed with the map

Expected:
The marker is not printed

oterral:
Fix #2163 

[test](http://mf-geoadmin3.dev.bgdi.ch/teo_fix_marker/)


procrastinatio:
Very good.

PR #2216, opened 2015-03-12
Use agnotic url for embed map

oterral:
@gjn a review please, we would like to have this for today

[test](https://mf-geoadmin3.dev.bgdi.ch/teo_agnostic/)

### DIFF
--- a/src/components/permalink/PermalinkService.js
+++ b/src/components/permalink/PermalinkService.js
@@ -57,7 +57,8 @@
             if (angular.isDefined(newParams.mobile)) {
               delete newParams.mobile;
             }
-            var baseEmbed = base.replace(/(index|mobile)\.html$/, 'embed.html');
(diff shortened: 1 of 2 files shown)


gjn:
It's working and code looks good.

But not sure if iframe tag really supports scheme agnostic urls. Seems strange to me, but...

gjn:
According to http://stackoverflow.com/questions/5946607/is-an-empty-iframe-src-valid, the iframe src attribute must contain a valid URL. And a valid url must contain the scheme.

gjn:
http://www.w3.org/TR/2011/WD-html5-20110525/the-iframe-element.html#attr-iframe-src

gjn:
But maybe the same is true for the <a> tag. So I would just assure that it will work in all browsers that we support.

Then again, we don't yet fully/officially support https (and it as not been tested...). So if it's included in https page, we might have surprises....

PR #2217, opened 2015-03-12
Cancel old profile requests

gjn:
Currently, profile requests are never canceled. So when drawing a polyline and profile drawing is active, each point drawn issues a profile requests, possibly adding a couple of such requests before the first profile is returned. Then, each profile returned will be drawn, resulting in slow updates.

We can't even be sure that the last profile drawn is the profile corresponding to the polyline drawn on the map. The profile drawn is the one that last returned from the service.

This PR assures that old pending profile requests are canceled before issuing a new profile request and thus eliminating the things mentioned above.

Also did minor refactoring.

### DIFF
--- a/src/js/MeasureController.js
+++ b/src/js/MeasureController.js
@@ -8,8 +8,8 @@
     'pascalprecht.translate'
   ]);
   module.controller('GaMeasureController',
-      function($scope, $translate, $http, $rootScope, $window, gaBrowserSniffer,
-          gaGlobalOptions, gaUrlUtils, gaPrintService, $timeout) {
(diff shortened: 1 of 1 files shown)


loicgasser:
Nice. Additionally, what about caching the last request to avoid performing useless requests when displaying/hiding the profile if the drawing hasn't changed.

gjn:
That's what the `cache: true` parameter is there for.

loicgasser:
Ok didn t look at the code, good that will be solved as well.

kandre:
Really nice and fast, tested on IE11, FF and chrome

oterral:
Lgtm, very nice improvment

PR #2219, opened 2015-03-12
Print time enabled layer with no data

procrastinatio:
Do not print time enabled layer which do not have data for defined year

See https://github.com/geoadmin/mf-geoadmin3/issues/2214

[Test link](http://mf-geoadmin3.dev.bgdi.ch/mom_print_year_none/src/?X=190000.00&Y=660000.00&zoom=1&lang=de&topic=inspire&bgLayer=ch.swisstopo.pixelkarte-farbe&layers_timestamp=19651231,&layers=ch.swisstopo.lubis-luftbilder_schwarzweiss,ch.swisstopo.lubis-luftbilder_farbe&time=1965)

### DIFF
--- a/src/components/print/PrintDirective.js
+++ b/src/components/print/PrintDirective.js
@@ -659,7 +659,8 @@
 
       // Transform layers to literal
       angular.forEach(layers, function(layer) {
-        if (layer.visible) {
+        if (layer.visible && (!layer.timeEnabled ||
+            angular.isDefined(layer.time))) {
           var attribution = layer.attribution;
           if (attribution !== undefined &&
               attributions.indexOf(attribution) == -1) {


src/components/print/PrintDirective.js
```diff
@@ -660,22 +660,25 @@
...
-            encLayers = encLayers.concat(encs);
-          } else {
-            var enc = encodeLayer(layer, proj);
-            if (enc && enc.layer) {
-              encLayers.push(enc.layer);
-              if (enc.legend) {
-                encLegends = encLegends || [];
-                encLegends.push(enc.legend);
+          if (!layer.timeEnabled ||
+               (layer.timeEnabled && layer.time != undefined)) {
```
oterral: Add these conditions to the `ìf` before

oterral: I think we could potentially have a problem here if we have a Layer.Group which is `timeEnabled = false ` and its childs are `timeEnabled=true` what do you think?

procrastinatio: Don't know. What is the *default* behaviour ? Layer are treated individually (as for scale), or the parent/group ovverride its property (as for catalog ) ?

src/components/print/PrintDirective.js
```diff
@@ -659,7 +659,8 @@
 
       // Transform layers to literal
       angular.forEach(layers, function(layer) {
-        if (layer.visible) {
+        if (layer.visible && (!layer.timeEnabled ||
+               (layer.timeEnabled && layer.time != undefined))) {
```
oterral: testing `layer.timeEnabled` is not needed because you already test `!layer.timeEnabled` above, am I wrong ?

PR #2226, opened 2015-03-13
Draw feature only when the FeatureTree is active

### LINKED ISSUE #2215: [Query tool] result highlights reappear when query was closed and user changes the topic
UC: open the topic aviation, make a query with the query tool. Close the query window. Switch from topic aviation to topic sachplan (or any other topic)
Result: After topic change the highlights of the resulting objects of the initial query reappear
Expected: After topic change the new topic appears with eventually some default layers, but no previously highlighted objects

oterral:
Fix #2215

[Test](http://mf-geoadmin3.dev.bgdi.ch/teo_fix_ft/)

### DIFF
--- a/src/components/featuretree/FeaturetreeDirective.js
+++ b/src/components/featuretree/FeaturetreeDirective.js
@@ -86,7 +86,7 @@
           },
           link: function(scope, element, attrs) {
             var map = scope.map;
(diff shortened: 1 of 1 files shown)


gjn:
n1.

Issue #2182, opened 2015-02-26
Embed map


davidoesch:
- [x] create embed.html

Create interface
- [x] add Text "Embed map" with one line with default code and disclaimer below
- [x] add show More options above disclaimer
- [x] "Show more" modal window contains
 Pulldown with
    - [x] preset sizes
    - [x] Custom sizes (minimum 200x200px), no max size limit
    - [x] Preview map (adapting to size)
- [x] disclaimer "By embedding this map, you agree to the terms of service" with link to http://www.geo.admin.ch/internet/geoportal/en/home/geoadmin/contact.html#iFrame

Embed map
- [x] white Box (analogue gmap embed) with top left, transparent outside: it is a link it goes to the map.geo.admin.ch the full size application (deskop mode)
 - [x]   icon (Favicon ) 
 - [x]   Text "view on map.xxx.admin.ch" xxx replaced by source domain
 Icons 
- [x] no geolocation
- [x] no overview
- [x] Footer left keep scale ( max size to middle of width) 
- [x] Footer right "C Data:" as in mobile ( max size to middle of width) if bigger then add 3 points (...) on mouseover show all dataowners links etc
- [ ] detect if embed.html is used in an iframe or embed context, if not display only a text "The map.geo.admin.ch Maps Embed must be used in an iframe." (Analogue to Google maps embed)

davidoesch:
oh, and I forgot_ would be nice to have it online by mid march

gjn:
+1000 for this from my side.

Do we have a max-size limitation?

davidoesch:
Max size --- a good lets see what gmap has as max size use this one

davidoesch:
Max size --- not limits: (aka fullscreen) gmap does not have one neither

oterral:
@davidoesch  does we allow rotation?  time selector ?

davidoesch:
Nope

oterral:
Are you sur for the time selector if a lubis layer is displayed that can useful no?

davidoesch:
The User embeds  a permalink - if there is a rotation or time parameter in there: it will be taken into account in the embed map. But I'm not sure that we should show the time selector /slider as button. But we should since otherwise the year is not shown ... So take it into account for rotation (user knows where north is) and the mobile time selector (pull down so it fits even on the 200 by 200 px)

oterral:
Ok

oterral:
Here a first example of the most complicated use case we have :  
[...]

PR #2230, opened 2015-03-18
Detect if the embed page is not used in a iframe

oterral:
Finalize #2182

[test](http://mf-geoadmin3.dev.bgdi.ch/teo_detetc_iframe/)


procrastinatio:
Well, better [link](http://mf-geoadmin3.dev.bgdi.ch/teo_detetc_iframe/embed.html)

OK, good to me.

PR #2232, opened 2015-03-18
Get rid of mobile parameter for  the redirect to main map in the embed page

oterral:
As mention @gjn  the big map link in the embed page always redirect to desktop version. This PR fixes that

[Test](http://mf-geoadmin3.dev.bgdi.ch/teo_bigmap_link/)


gjn:
Works as expected. Thanks alot.

PR #2235, opened 2015-03-20
Use marker as default crosshair

cedricmoullet:


cedricmoullet:
@davidoesch do you agree ?
@pauloamado could you adapt the help ? http://help.geo.admin.ch/?id=54&lang=de marker is also one option of the crosshair parameter

davidoesch:
Do u Mean by marker: the one used for the search result preview on map? If yes ok.

In this case we have to adapt the help screenshot

cedricmoullet:
@davidoesch yes !
@pauloamado sorry, but the screenshot has also to be adapted.

pauloamado:
http://help.geo.admin.ch/?id=54&lang=de is now adapted. Currently no screenshots are concerned in help.geo.admin.ch. Should we add something in http://help.geo.admin.ch/?id=12&lang=de (link with "marker" crosshair)?

PR #2237, opened 2015-03-23
[WIP] Add description input in draw

oterral:
This PR fixes some minor issue and add the possibility to add a descripiton to a feature drawn

[test](http://mf-geoadmin3.dev.bgdi.ch/teo_descr/?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&X=196967.94&Y=650563.48&zoom=0)


AFoletti:
The description looks like to be "per drawing session"  and not "per object". Am I right.
If it's true, I find the usefulness quite limited ;)

oterral:
No it's per object

AFoletti:
yep, saw that.
But I had to test 3 workflows to make it happen. Low usability score ;)

My 2 cents:
- make the text you enter be associated with the last object drawn, even if the drawing has already finished (double-click for lines/poly, or point placement). At the moment you FIRST have to enter the text, then the object
- make the previously entered text disappear when you draw a new object

davidoesch:
The description is not visisble in the drawing --- this will irritate the user.
Either the Tag is visibile as text center of polygone line point or get rid of it.... and what is the difference from a text label? 
I have the impression we leave here KISS

AFoletti:
Completely opposite as KISS, but what about adding a title field for the object, with a checkbox to show/hide it on the map?

davidoesch:
Sorry Folks , BUT where is the actual request for that? ... This is a pure GIS function...feature creep. The interface is now overcrowded. We have to rethink the draw interface as a whole. IMHO : this feature is not needed. The next consequences would be: edit the attribute table. 
So: don't add it.

cedricmoullet:
Of course, feature creep. But drawing is already feature creep. I continue with private mail

ponceta:
This is then a great creep. For improved users, it's a good thing. I wish I had a simple kml editor like that.

gjn:
It's GIS-like attribute, not description (we have Labels for that). The magic of it is when drawing is exported as KML, then the user will have the info in the tooltip. (writing that a thought pops up: why not edit in a tooltip?)

We agreed that UI/interaction is not optimal now.

oterral:
I think adding description  will  be good.

Here it's a new concept I've done yesterday (not fully working) but the idea is here, what do you think? better? worse? the problem when we put all in a tooltip is that, if you want to add new properties in the tooltip, it  will become too big or too unsable.

We can also make a mix of what we had. Leave as it is an only add the possibility to modify the description in a tooltip at the end of drawing or when the feature is selected (best way if we want the functionnality quickly)

or we work on major redesign and we wait for it.

oterral:
I close this one since a new design is in preparation

PR #2243, opened 2015-03-24
Remove useless export

oterral:
According the api doc getResolutions is stable now and exported:

https://github.com/openlayers/ol3/blob/b6adeea4f4cb31fbd0976ad1d91e70c1c66f39c7/src/ol/tilegrid/tilegrid.js#L188


oterral:
trivial

Issue #2238, opened 2015-03-23
HTTPS per default


cedricmoullet:
Knowns issues:

 - [x] Help: images in html content
 - [ ] Import WMS : Url of tiles WMS layer added
 - [ ] Import KML: Images google in kml and link to KML validator
 - [ ] Map: Sometimes tiles SSL error with WMTS. and urls of WMS layers like luftfahrthindernisse are http

cedricmoullet:
@davidoesch @hpchrist Do you agree to go on HTTPS per default ?

hpchrist:
yes, on the infra side, the startcom extended validation SSL certificates are ready and seem to well support the wast majority of browsers (in comparison to the BIT certificates;-). But what about the application? Did you already check all directly embedded links to non https resources (in the worst case even outside our FSDI)?

davidoesch:
Help: images in html content
-> We move them away from dav to a https s3 bucket
Import WMS : Url of tiles WMS layer added
-> if we have an ssl warning: IMHO the user can live with that if he gets a ssl error due to a permalink with a WMS over http only
Import KML: Images google in kml and link to KML validator
- Images google: ok this one is tricky since it used often:can the user live with an SSL error here? what happens it will not work? other workarounds? galdos KML validator: we can move it to the help ( o gosh, https as well.....) 
Map: Sometimes tiles SSL error with WMTS. and urls of WMS layers like luftfahrthindernisse are http

-> this one is  a show stopper:"Sometimes tiles SSL error with WMTS" why? WMS layers we have to fix

davidoesch:
Help: images in html content
-> We move them away from dav to a https s3 bucket
Import WMS : Url of tiles WMS layer added
-> if we have an ssl warning: IMHO the user can live with that if he gets a ssl error due to a permalink with a WMS over http only
Import KML: Images google in kml and link to KML validator
- Images google: ok this one is tricky since it used often:can the user live with an SSL error here? what happens it will not work? other workarounds? galdos KML validator: we can move it to the help ( o gosh, https as well.....) 
Map: Sometimes tiles SSL error with WMTS. and urls of WMS layers like luftfahrthindernisse are http

-> this one is  a show stopper:"Sometimes tiles SSL error with WMTS" why? WMS layers we have to fix

davidoesch:
Help: images in html content
-> We move them away from dav to a https s3 bucket
Import WMS : Url of tiles WMS layer added
[...]

PR #2259, opened 2015-04-01
Add https to help links

ponceta:
In the purpose of going full HTTPS, this PR redirect all help queries to https://help.geo.admin.ch/

This goes in the sense of #2238


gjn:
Please merge when the source is correct as well.

ponceta:
I'll need to have access to the source to modify it.

oterral:
maybe a stupid question but where we use these ids `are_help_link_href` ?

ponceta:
tooltip?

gjn:
https://github.com/geoadmin/mf-geoadmin3/blob/master/src/index.mako.html#L176

gjn:
The source is a google spreadsheet. It's no use if it isn't changed there.

oterral:
ah yes , but a 2nd question question comes to my mind, It is always the same links, we could  just remove this and put an agnostic url in index.mako but probably there was a reason for that.

ponceta:
Non there's a script.js for the side menu, the fusion table is hosting only the content.

gjn:
yes, but this content (what I meant by source) needs to be updated as well, otherwise we'll overwrite again on next translate.

ponceta:
The doc is up to date, I've generated the json files with it.

gjn:
ok. Thanks alot.

oterral:
@gjn do you see my last comment? could just get rid of these translations?

gjn:
Yes, you are right. They can be removed from the translations as they are always the same for each topic.

@ponceta Could you do that?

ponceta:
Yes of course, we just need to be sure that all these links are not used anymore anywhere else @davidoesch @mariokeusen @cedricmoullet 

Meaning the one of 
 - are_help_link_href
 - bafu_...
 - blw
 - ech
 - funksender
 - geol
 - inspire
 - ivs
 - kgs
 - nga
 - sachplan
 - swisstopo
 - luftbilder
 - aviation
 - mgdi
 - gewiss
 - wildruhezonen
 - vu

gjn:
The *_help_link_href are always the same for all topics. So we can remove them and put the link directly in the index.mako.html (see reference above) are topic independant.

AFAIK, the translation spreadsheet is only used by this project.

Issue #1991, opened 2014-12-18
[Testing] Add WMS availability Service


ponceta:
@ltclm @cedricmoullet @gjn 

If I understood well what Marcel told me, we want to be able to predict wms status a bit like the status tool for api/geoadmin http://status.geo.admin.ch/

Is it right?

PR #2261, opened 2015-04-02
Add WMS / KML separated test

ponceta:
Add WMS layer to the browserstack print test.

Improves testing in the #1991 way

### DIFF
--- a/test/selenium/kml_test.js
+++ b/test/selenium/kml_test.js
@@ -1,0 +1,40 @@
+// KML import test using browserstack
+
+var webdriver = require('browserstack-webdriver');
+var assert = require('assert');
+
+var QUERYSTRING_KML = "KML%7C%7Chttp:%2F%2Fopendata.utou.ch%2Furbanproto%2Fgeneva%2Fgeo%2Fkml%2FRoutes.kml";
+
+var runTest = function(cap, driver, target) {
+  //We maximize our window to be sure to be in full resolution
+  driver.manage().window().maximize();
+  // Goto the travis deployed site.
+  driver.get(target + '/?lang=de');
+  //wait until topics related stuff is loaded. We know this when catalog is there
+  driver.findElement(webdriver.By.xpath("//a[contains(text(), 'Grundlagen und Planung')]"));
+
+  // Click on "Werkzeuge"
+  driver.findElement(webdriver.By.xpath("//a[@id='toolsHeading']")).click();
+  // Click on "KML Import"
+  driver.findElement(webdriver.By.xpath("//*[contains(text(), 'KML Import')]")).click();
+  // Click on "URL"
+  driver.findElement(webdriver.By.xpath("//a[contains(text(), 'URL')]")).click();
+  // Write URL of the chosen KML
+  driver.findElement(webdriver.By.xpath("//*[@id='import-kml-popup']//input[@placeholder='URL einfügen']")).sendKeys('http://opendata.utou.ch/urbanproto/geneva/geo/kml/Routes.kml');
+  // Click on "KML Laden"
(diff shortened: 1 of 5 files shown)


gjn:
I like the idea! But those are some ugly selectors....are there no other ways to select them (id or class?)

Or: Why not put it directly in a) separate test and b) put the wms directly in the link? (I've proposed in some PR on how to possibly check for the alert message.)

gjn:
Another thing to consider. Now if the print fails, the reasons might be many-fold: print itself, kml loading, wms loading, etc. I also tend to think the standard print (without wms or kml imports) are more important than kml/wms prints.

Maybe a little separation of concerns would make sense. Just thoughts here.

procrastinatio:
I aggree with @gjn, do not over engineer the print test. The result is also visual, and unless you also test the resulting PDF, the test is not complete.

ponceta:
jenkins test

test/selenium/kml_test.js
```diff
@@ -0,0 +1,40 @@
...
+  driver.findElement(webdriver.By.xpath("//a[contains(text(), 'Grundlagen und Planung')]"));
+
+  // Click on "Werkzeuge"
+  driver.findElement(webdriver.By.xpath("//a[@id='toolsHeading']")).click();
+  // Click on "KML Import"
+  driver.findElement(webdriver.By.xpath("//*[contains(text(), 'KML Import')]")).click();
+  // Click on "URL"
+  driver.findElement(webdriver.By.xpath("//a[contains(text(), 'URL')]")).click();
+  // Write URL of the chosen KML
+  driver.findElement(webdriver.By.xpath("//*[@id='import-kml-popup']/div[2]/div/div/div[1]/div/div[2]/form/input[1]")).sendKeys('http://opendata.utou.ch/urbanproto/geneva/geo/kml/Routes.kml');
```
ponceta: look for better selector

ponceta:
jenkins test

ponceta:
jenkins test

ponceta:
Please review ;)

test/selenium/wms_test.js
```diff
@@ -0,0 +1,46 @@
...
+  // Write URL of the chosen WMS
+  driver.findElement(webdriver.By.xpath("//*[@id='import-wms-popup']//input[@placeholder='URL']")).sendKeys('http://wms.geo.admin.ch/');
+  // Click on "Verbinden"
+  driver.findElement(webdriver.By.xpath("//*[@id='import-wms-popup']//button[contains(text(),'Verbinden')]")).click();
+  // Click on "AGNES"
+  driver.findElement(webdriver.By.xpath("//*[@id='import-wms-popup']//div[contains(text(),'AGNES')]")).click();
+  // Click on "Layer hinzufügen"
+  driver.findElement(webdriver.By.xpath("//*[@id='import-wms-popup']//button[contains(text(),'Layer hinzufügen')]")).click();
+  // Accept the alert
+  driver.switchTo().alert().accept();
```
gjn: If we wait for https://github.com/geoadmin/mf-geoadmin3/pull/2282, we could remove this, right?

gjn:
I like this alot! Good work.

It needs rebase, and maybe we should also wait for https://github.com/geoadmin/mf-geoadmin3/pull/2282

test/selenium/wms_test.js
gjn: Maybe not...

ponceta: The alert is the way we assess the user that the layer is now loaded. I do not like alerts but this won't disapears with 3rd party data agreement.

ponceta:
:) This is just separation of concerns for testing, it will not be influenced a lot by #2282.

After #2282 I'll add tests with kml/wms permalinks.

I'll rebase it.

gjn:
Very good improvement. Thanks alot.

PR #2262, opened 2015-04-02
Add agnostic link to help

ponceta:
Contributes to #2238

Does not work for RM since https://help.geo.admin.ch/?lang=rm does not exist

@oterral / @gjn Do I create a new translation for that or do you have an angular magic trick for this?

like if `rm` redirect to `de`.


ponceta:
I'll remove directly the tokens once redirection of rm wil be fixed.

src/index.mako.html
```diff
@@ -173,7 +173,7 @@
           <a href="" ng-click="globals.feedbackPopupShown = !globals.feedbackPopupShown">
             <span ng-class="{'selected': globals.feedbackPopupShown}" translate>problem_announcement</span>
           </a>&nbsp;&nbsp;
-          <a target="_blank" ng-href="{{topicId + '_help_link_href' | translate}}" translate>help_label</a>&nbsp;&nbsp;
+          <a target="_blank" ng-href="//help.geo.admin.ch/?lang={{langId}}" translate>help_label</a>&nbsp;&nbsp;
```
oterral: I'm not 100% sure but I think we don't need to use `ng-href`

ponceta: Really? But if you use a std href, {{langId}} won't be interpreted correctly! No?

oterral: ng-href allow two way data binding (read/write) in our case we need just read.

ponceta: You're right :+1:

ponceta:
search bern failed

jenkins test

oterral:
jenkins test

ponceta:
jenkins test

gjn:
Very nice. Thanks alot.

PR #2294, opened 2015-04-16
Optimize src images

ponceta:
Optimized src images using debian optimization functions :

`jpegoptim` and `optipng`

http://www.webpagetest.org/result/150416_KB_GWM/1/performance_optimization/#compress_images

Wins some precious kbytes


cedricmoullet:
@pauloamado @AFoletti : could you integrate this step when you create new images ?

Issue #1401, opened 2014-05-08
Popup: not resizable 


oterral:
I think we could something at least with the css3 property `resize`.

https://developer.mozilla.org/de/docs/Web/CSS/resize

ponceta:
Note: Internet Explorer and Opera do not support the resize property.

oterral:
only IE : http://caniuse.com/#search=resize

ponceta:
[Testlink](http://mf-geoadmin3.dev.bgdi.ch/ltpoa/)

For me there is no great plus-value since our interface is responsive and adapts to the user resolution and device and bring to the user the best experience possible (which includes already the best popups sizes possible) It should not be the user job to have to resize the popups (if he does, that means that we failed bringing him the resolutions he expected)

cedricmoullet:
Please try with several lubis features. 

![image](https://cloud.githubusercontent.com/assets/319676/7247745/13c7a45e-e80d-11e4-8640-88925c2924af.png)

the content is not resized

ponceta:
This comes from our defined max height and max width popups content containers.

If we want to go further with this, we should probably list first the popups we WANT to be resizable and the ones we do not want to, right?

gjn:
Despite my +1 at the start, I tend to support Arnaud's view: we should make the content to fit the pre-defined size. Resizing complexifies the UI...not sure we really want this with our usability efforts we want to make.

oterral:
I close this since the resizable thing introduce too much css complexity and the dynamic size of poup has been improved with  https://github.com/geoadmin/mf-geoadmin3/pull/728

PR #2300, opened 2015-04-21
Add resizable popups

ponceta:
Add the possibility to the user to resize the popups (not supported on IE)

Enhancement proposal #1401

[Testlink](http://mf-geoadmin3.dev.bgdi.ch/ltpoa/)

### DIFF
--- a/src/components/popup/style/popup.less
+++ b/src/components/popup/style/popup.less
@@ -1,6 +1,7 @@
 @ga-popup-max-height: 440px;
 @ga-popup-reduced-height: 40px;
 @ga-popup-max-width: 620px;
+@ga-popup-resize: both;
 
 [ga-popup] {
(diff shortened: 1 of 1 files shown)


ponceta:
jenkins test

gjn:
What is the change?

loicgasser:
Intresting I didn't know this css property. http://www.w3schools.com/cssref/css3_pr_resize.asp

Not supported on IE and you might need the -moz suffix for Firefox.

oterral:
I've already tried this solution once and  it's really not perfect because sometimes the content of the popup doesn't fit correctly the new size of the popup. To be sure, we should verify each popup content on each browser and fix it if the display is not correct, obviously that can lead to unwanted css bugs.

PR #2303, opened 2015-04-21
Improving makefile

loicgasser:
Before, one needed to use `make cleanall` and then rebuild the whole thing and finally make ol to update ol3 if ol3 folder already existed in `.build-artefacts` before hand.

This PR makes it now possible to update ol3 by simply changing the OL3_VERSION (It can be a commit, a tag reference etc...) variable or by editing ` scripts/ol-geoadmin.json`.

I also added localforage to package.json similarly to what has been done with angularjs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -157,12 +157,6 @@ fastclick: .build-artefacts/fastclick .build-artefacts/closure-compiler/compiler
 	cp .build-artefacts/fastclick/lib/fastclick.js src/lib/fastclick.js
 	java -jar .build-artefacts/closure-compiler/compiler.jar src/lib/fastclick.js --compilation_level SIMPLE_OPTIMIZATIONS --js_output_file  src/lib/fastclick.min.js
 
-.PHONY: localforage
-localforage: .build-artefacts/localforage
-	cd .build-artefacts/localforage && git reset HEAD --hard
-	cp .build-artefacts/localforage/dist/localforage.js src/lib/localforage.js
-	cp .build-artefacts/localforage/dist/localforage.min.js src/lib/localforage.min.js
-
 .PHONY: filesaver
 filesaver: .build-artefacts/filesaver
 	cp .build-artefacts/filesaver/FileSaver.js src/lib/filesaver.js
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -273,6 +269,7 @@ node_modules: package.json
 	cp $(addprefix node_modules/angular/,$(ANGULAR_JS)) src/lib/;
 	cp $(addprefix node_modules/angular-translate/dist/,$(ANGULAR_TRANSLATE_JS)) src/lib/;
 	cp $(addprefix node_modules/angular-translate/dist/angular-translate-loader-static-files/,$(ANGULAR_TRANSLATE_LOADER_JS)) src/lib/;
+	cp node_modules/localforage/dist/localforage.min.js src/lib;
```
oterral: we need localforage.js also

oterral:
Do you think we could override the param OL3_version simply passing the commit in parameter like this:

` make ol  2f602cd4663aba111ff6e53399e27d978c28f186`

loicgasser:
yes probably or in rc file export OL3_VERSION = ... and adapt make file to detect if variable is already set. Would be cool with rc_branch

Makefile
loicgasser: ok will do

loicgasser:
@oterral I made the correction for localforage and I added the command you suggested.
You can now use:
`make OL3_VERSION="632205d902f8dcc1f03eb1dd1736d26a1b3ac2a3" ol` to update the project to a given version.

loicgasser:
jenkins test

Makefile
```diff
@@ -140,6 +141,11 @@ ol: OL_JS = ol.js ol-debug.js
 ol: scripts/ol-geoadmin.json .build-artefacts/ol3 .build-artefacts/ol-requirements-installation.timestamp
 	cd .build-artefacts/ol3; \
 	git reset HEAD --hard; \
+	ifdef ${OL3_VERSION} \
+	git checkout ${OL3_VERSION}; \
+	else \
+	git checkout $(OL3_VERSION); \
+	enif; \
```
oterral: the `if`and `else ` do the same thing, is it normal?

loicgasser: not the same:
` ${OL3_VERSION}` refers to an argument 
`$(OL3_VERSION)` refers to a variable in makefile

gjn: You could use `OL3_VERSION ?= tags/v3.4.0` in the definition part. It will set OL3_VERSION varaible to OL3_VERSION parameter when given, otherwise to tags/v3.4.0 (as is already done for the DEPLOY_TARGET variable)

loicgasser: yup nice one

loicgasser: Are you sure it works when passing arguments directly in a make command. DEPLOY_TARGET uses rc files.

gjn: Not tested, but if not, you could do
`OL3_VERSION=some_hash && make ol`

oterral: @loicgasser ah yes thanks for the explanation

gjn: @loicgasser Can we use the other approach, I think it would make the code easier.

loicgasser: yes no problem, but it makes the command harder.

loicgasser: it actually works the same way, you can still use `make OL3_VERSION=bla ol` good to know. making additional tests

loicgasser:
@gin adapted according to your comment. Nice!

gjn:
Nice one.

ponceta:
Yes much better that way :)

PR #2309, opened 2015-04-23
Add close title to popups

ponceta:
image

Small fix

### DIFF
--- a/src/components/contextpopup/ContextPopupDirective.js
+++ b/src/components/contextpopup/ContextPopupDirective.js
@@ -6,12 +6,13 @@
 
   var module = angular.module('ga_contextpopup_directive', [
     'ga_networkstatus_service',
-    'ga_permalink'
+    'ga_permalink',
+    'pascalprecht.translate'
   ]);
 
   module.directive('gaContextPopup',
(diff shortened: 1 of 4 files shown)


ponceta:
jenkins test

src/components/contextpopup/ContextPopupDirective.js
```diff
@@ -6,12 +6,13 @@
```
gjn: Why do it in the contextpopup directive? Having it in the popup directive alone should be enough. No?

src/components/popup/PopupDirective.js
```diff
@@ -149,6 +150,7 @@
             scope.titlePrint = $translate.instant('print_action');
             scope.titleHelp = $translate.instant('help_label');
             scope.titleReduce = $translate.instant('reduce_label');
+            scope.titleClose = $translate.instant('close');
```
gjn: Does not react to translation changes.

gjn: Please ignore both my comments. I was being stupid.

oterral:
LGTM but I suggest in the future  we try to use th new version of angular-translate see https://github.com/geoadmin/mf-geoadmin3/issues/2310

PR #2311, opened 2015-04-23
Only very basic tests for IE9

gjn:
As written in the mail yesterday, > 95% of failed tests that were cause by Browserstack hiccups were cause by IE9 tests.

This PR removes most IE9 tests that fail on a regular basis. It only keeps the basic loading test. Hopefully, this improves robustness of BS.


ponceta:
:) I was going to propose the same! 

+1

PR #2326, opened 2015-04-30
Search only enabled when topic is loaded

gjn:
This PR assures that search requests are only send when a topic is loaded. Without this fix, the client
sometimes sends search requests with topic 'undefined', resulting in uneeded traffic and 400 response from back-end

See also: https://logs.bgdi.ch/kibana/#/dashboard/elasticsearch/undefined%20topic


oterral:
jenkins test

PR #2330, opened 2015-05-01
Makes Profile independent from Measure

oterral:
Changing the design of the draw functionnality implies a little refactoring of Profile and Measure  directives. 

With this PR the Profile accepts an ol Feature as parameter, listen every change on this feature and redraw the profile automatically.

This PR makes also the width of the profile dynamic. No need to specify a width by default anymore, the Profile takes the size of the parent div.

[TEST](https://mf-geoadmin3.dev.bgdi.ch/teo_prof/)

### DIFF
--- a/src/components/profile/ProfileService.js
+++ b/src/components/profile/ProfileService.js
@@ -1,77 +1,70 @@
 (function() {
   goog.provide('ga_profile_service');
 
+  goog.require('ga_urlutils_service');
+
   var module = angular.module('ga_profile_service', [
+    'ga_urlutils_service',
     'pascalprecht.translate'
   ]);
 
-  module.provider('gaProfileService', function() {
+  module.provider('gaProfile', function() {
+    var d3;
 
-    this.$get = function($timeout, $translate, $window) {
+    // Utils functions
+    var createArea = function(domain, height, elevationModel) {
+      return d3.svg.area().x(function(d) {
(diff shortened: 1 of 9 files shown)


src/components/profile/ProfileService.js
```diff
@@ -260,10 +278,19 @@
...
-      var d3LibUrl = this.d3libUrl;
       return function(options, lazyLoadCB) {
-        return new ProfileChart(options, lazyLoadCB, d3LibUrl);
+        // Lazy load of D3 library
+        var onD3Loaded = function() {
+          d3 = $window.d3;
+          lazyLoadCB();
+        };
+        if (!$window.d3) {
+          window.console.log('la');
```
gjn: debug code...

src/js/MeasureController.js
```diff
@@ -23,7 +22,8 @@
                 left: 60
               },
               height: 188,
-              elevationModel: 'COMB'
+              elevationModel: 'COMB',
+              profileFeature: null
```
gjn: can't see where profileFeature is needed. I don't think it is.

gjn:
Very nice re-factoring. Added 2 minor comments, otherwise, I think this can be merged.

oterral:
Thx @gjn I fix the 2 issue and then merge

PR #2336, opened 2015-05-05
Add feedback test

### LINKED ISSUE #2335: [Feedback Tool] doesn't allows to browse/upload files on chrome/firefox
It works without any issue on IE but clicking on the browse button doesn't do anything on chrome/FF AND spartan

ponceta:
Preliminary version, needs to fix #2335 before going further

Test is actually only on IE 10 to avoid too many messages to the webgis mailbox

### DIFF
--- a/test/selenium/feedback_test.js
+++ b/test/selenium/feedback_test.js
@@ -3,15 +3,14 @@
 var webdriver = require('browserstack-webdriver');
 
 var runTest = function(cap, driver, target) {
-  if (cap.browser == 'chrome' && cap.browser_version == '37') {
+  if (cap.browser == 'IE' && cap.browser_version == '10.0') {
     //We maximize our window to be sure to be in full resolution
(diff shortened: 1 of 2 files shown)


gjn:
We should think about avoiding sending the messages to the help desk. Hmm...how?

Can we filter them out before they appear in the inbox? (Remember, soon those emails will have to be treated by the SPOC, which is outside IWI/IKP.

ponceta:
I tried an outlook rule to make these directly moved into a specific folder but I did not manage to get it to the destination folder. We could just delete theses but in this case it's maybe not worth? Who is our outlook server manager? :)

cedricmoullet:
I think that we could avoid to send emails.

gjn:
In light of SPOC, I think it's too cumbersome to send emails with every PR and every test run of jenkins. I'll close this.

PR #2340, opened 2015-05-06
Add phantomjs to package.json so it is installed automatically

Jenselme:


gjn:
Nice contribution. Thanks alot.

PR #2341, opened 2015-05-06
Add configuration options

Jenselme:
Add options for:
- default topic
- translation fallback
- default bounding box
- default zoom level

### DIFF
--- a/rc_dev
+++ b/rc_dev
@@ -4,3 +4,7 @@ export API_URL=//mf-chsdi3.dev.bgdi.ch
 export APACHE_BASE_PATH=
 export BROWSERSTACK_TARGETURL=https://mf-geoadmin3.dev.bgdi.ch
 export VARNISH_HOSTS=()
+export DEFAULT_TOPIC_ID=ech
+export TRANSLATION_FALLBACK_CODE=de
+export DEFAULT_EXTENT='[485869.5728, 76443.1884, 837076.5648, 299941.7864]'
+export DEFAULT_RESOLUTION=500.0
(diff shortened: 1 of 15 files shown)


src/js/TranslationController.js
```diff
@@ -4,7 +4,7 @@
   var module = angular.module('ga_translation_controller', []);
 
   module.controller('GaTranslationController',
-      function($scope) {
+      function(gaGlobalOptions, $scope) {
```
gjn: Please put gaGlobalOptions as second parameter.

rc_dev
```diff
@@ -4,3 +4,7 @@ export API_URL=//mf-chsdi3.dev.bgdi.ch
 export APACHE_BASE_PATH=
 export BROWSERSTACK_TARGETURL=https://mf-geoadmin3.dev.bgdi.ch
 export VARNISH_HOSTS=()
+export DEFAULT_TOPIC_ID=ech
+export TRANSLATION_FALLBACK_CODE=de
+export DEFAULT_EXTENT='[485869.5728, 76443.1884, 837076.5648, 299941.7864]'
+export DEFAULT_RESOLUTION=500.0
```
gjn: Please put the default values directly in the Makefile, where it could/can be overwritten withe environment variabls; similar to https://github.com/geoadmin/mf-geoadmin3/blob/master/Makefile#L18. Otherwise, we can't build without these variables being defined. Also, like this, you don't need to put default values in any rc_* file.

You can overwrite those values using any rc_* file then.

gjn:
Thanks alot for your contribution.

I have added 2 comments that need correction before I can merge this. In addition to those 2 issues, please squash your commits so there's only one commit to merge.

Jenselme:
Just did it!

gjn:
Great. Thanks again for the contribution.

PR #2342, opened 2015-05-06
Change the way html files are generated from the templates

Jenselme:
- Switch to a mako (installed on the system) on python3 to avoid encoding problem. Your patched version of cmp.py is not needed any more.
- Use html-minifier (a nodejs module) to minify HTML instead of htmlmin (python)

If you want only parts of these modification, feel free to ask me to remove some commits.

### DIFF
--- a/src/index.mako.html
+++ b/src/index.mako.html
@@ -7,7 +7,7 @@ itemscope itemtype="http://schema.org/WebApplication"
 % endif
 >
   <head>
-    <!--![if !HTML5]>
+    <!--[if !HTML5]>
     <meta http-equiv="X-UA-Compatible" content="IE=9,IE=10,IE=edge,chrome=1"/>
     <![endif]-->
 % if device == 'desktop':
@@ -118,7 +118,7 @@ itemscope itemtype="http://schema.org/WebApplication"
           'embed': globals.embed
         }">
     <!-- The below conditional is ignored by >= IE10 and all other browsers -->
-    <!--![if IE]>
+    <!--[if IE]>
       <script>
        (function(){
        var msgDocumentMode = "";
@@ -618,7 +618,7 @@ itemscope itemtype="http://schema.org/WebApplication"
     <script src="${versionslashed}lib/build.js"></script>
     <!-- For <= IE9, we load it explicitely as XDomain does not support
          dynamic/lazy loading of javascript -->
-    <!--![if IE]>
+    <!--[if IE]>
     <script src="${versionslashed}lib/bootstrap-datetimepicker.min.js"></script>
     <script src="${versionslashed}lib/d3-3.3.1.min.js"></script>
     <script src="${versionslashed}lib/jQuery.XDomainRequest.js"></script>
--- a/src/index.mako.html
+++ b/src/index.mako.html
@@ -1,4 +1,4 @@
-﻿# -*- coding: utf-8 -*-
+# -*- coding: utf-8 -*-
 <!DOCTYPE html>
 <html ng-app="ga" ng-controller="GaMainController"
 itemscope itemtype="http://schema.org/WebApplication"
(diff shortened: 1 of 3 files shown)


gjn:
This PR would force us to change some of our development environments, so there should be good reasons to apply it. Could you describe the problems you want to solve with this?

Jenselme:
Despite your patch to the mako command line, we still encountered encoding issues. The switch to the python3 version allow us to completely avoid these problems and avoid to apply your custom patch (even if this patch is not a big deal). I guess a compromise would be to use all dependencies within a pyton3 venv.

Moreover, we would like to limit the number of technologies used. Because of the nature of the front-end, JS tools must be used. So, our mid-term goal is to get rid of the python venv. To do so, we must use node based tools wherever possible. For the few tools for which we don't have an equivalent yet (like the closure tools), we use system wide install.

Our long-term goal is to switch from the Makefile to a gulp based solution. It may or may not interest you since it will deeply change your workflow.

gjn:
What kind of encoding problems did you encounter?

System wide installs triggered by builds from our software are simply a no-go for us at this moment. All our systems are managed by puppet, and we will keep it that way. Also, we don't have python3 on those systems and such an addition would mean heavy testing of a couple of servers running a multitude of applications and tools. There is not plan to update to python3 anytime soon.

Anyhow, I don't think it would be problematic to use JS tools instead of python tools for specific tasks like template rendering. We simply used mako because it's the templating engine that is best known in our team.

As said in the other PR, we don't plan to switch to gulp or any other build tool. In fact, we want to replace our buildout build tool in mf-chdsi3 to GNU make.

Jenselme:
> What kind of encoding problems did you encounter?

I haven't written it down and sadly I am not able to reproduce it today. But it was something like: `UnicodeEncodeError: 'ascii' codec can't encode character u'CODE' in position POS: ordinal not in range(128)`. Now that I think of it, it was maybe just your patch which failed to install (on our system we had a problem with a missing symlink. I corrected that in 10b0f54) and we try to launch without noticing the failure.

> System wide installs triggered by builds from our software are simply a no-go for us at this moment. All our systems are managed by puppet, and we will keep it that way. Also, we don't have python3 on those systems and such an addition would mean heavy testing of a couple of servers running a multitude of applications and tools. There is not plan to update to python3 anytime soon.

Noted! Commit removed.

> Anyhow, I don't think it would be problematic to use JS tools instead of python tools for specific tasks like template rendering. We simply used mako because it's the templating engine that is best known in our team.

OK. Once we made the switch, I will let you know.

> As said in the other PR, we don't plan to switch to gulp or any other build tool. In fact, we want to replace our buildout build tool in mf-chdsi3 to GNU make.

This is was I understood from a comment on a PR in chdsi.

src/index.mako.html
```diff
@@ -1,4 +1,4 @@
-﻿# -*- coding: utf-8 -*-
+# -*- coding: utf-8 -*-
```
gjn: Did you add/remove the BOM?

src/index.mako.html
```diff
@@ -7,7 +7,7 @@
 % endif
 >
   <head>
-    <!--![if !HTML5]>
+    <!--[if !HTML5]>
```
gjn: Why?

gjn: I think this is needed for various IE versions.

src/index.mako.html
Jenselme: Yes. I thought my encoding problems might come from there. Is it a problem?

src/index.mako.html
Jenselme: It is. htmlmin (python) required this commenting format to preserve the comment after minification while html-minifier preserved them by default.

gjn: I understand.

src/index.mako.html
gjn: Not sure if it's a problem. Would like to keep it as it was if possible.

gjn:
I'll close this as we rather prefer to stay put. (I like the Makefile cleanups though...)

PR #2343, opened 2015-05-06
Rewrite translation script

Jenselme:
This is a rewrite of your translation script in a more readable and shorter way. I remove the support for google spreadsheet and the YAML configuration file (whose interests I was not able to see). If you are interested by this version of the script, I can help you add back the support of google spreadsheet.

The old script is not removed.


gjn:
I would be interested in this, but we need the google spreadsheet support, as the source of our translations.
The YAML is just used for configuration and could be removed and be put into the script directly.

Jenselme:
Good new! Before adding gsreads support, I would like to discuss how the script should work. From what I understand from your workflow, the translation script is only called within the Makefile. So for the script to work for both project, I should:
- Add two variables in the Makefile to easily select the locales and translation source. These variables could be overridden by environment variables.
- Modify the translate target to call the script with these variables.
- Add support in the script for gspread. If the argument ends with `.csv` assume it is a CSV file. If it doesn't, assume it is the key of a google spreadsheet.

Is it OK for you?

gjn:
Updating translations is _always_ a manual process, which involves calling `make translate` and creating a PR with the new translations in this github repository.

As long as our current hanlding of it doesn't change, I welcome those changes. This means that using _our_ google spreadsheet and all our languages should be the default options that will work with `make translate`

Jenselme:
OK, this is what I intended to do.

Jenselme:
I have updated my script to parse value from google spreadsheet. It uses your spreadsheet in the Makefile.

I had to use OAuth2 for authentication since I was not able to authenticate with the standard login. According to [this page](https://developers.google.com/identity/protocols/AuthForInstalledApps), this auth is not supported any more.

You can read [this](http://gspread.readthedocs.org/en/latest/oauth2.html?highlight=oauth) to access your document with OAuth2 and gspread. Please note that for this to work, I had to share the document with the email address `client_email` from the JSON file (see the documentation of gspread for the explanations).

Sadly, gspread doesn't support anonymous access for public document yet: https://github.com/burnash/gspread/issues/175. So I did my tests with a copy of your spreadsheet in my drive.

gjn:
Would it be possible to use the same access to the spreadsheet as in the existing spreadsheet? It was working just fine and we don't see any reason why we would want to change our approach on how we update the translations.

Jenselme:
Just did it. Please note that without a login/password for this drive, I was not able to test.

cedricmoullet:
@alcapat is this PR still useful ?

alcapat:
@Jenselme Thanks for your PR. But between you create it and now, we have changed the authentication method. We have replaced the USER/PASSWORD to oauth2client. 
Can you please update your code ?
You can have more information here : https://github.com/geoadmin/mf-geoadmin3/commit/a30524f6e3b6c5a95cc76be6dcbac475b5e08a88 
Best regards.

Jenselme:
PR updated

gjn:
@Jenselme I think this needs a rebase. And it should be ready to test/review, right?

Jenselme:
PR rebased

scripts/translation2js.yaml
```diff
@@ -1,2 +0,0 @@
-langs: [de,fr,en,it,rm]
-emptyFilename: empty.json
```
gjn: Where is this configuration handled now?

scripts/translation2json.py
```diff
@@ -0,0 +1,115 @@
+#!/usr/bin/env python3
+"""
+translation2js.py
+This script is used to create the {lang}.json files from the csv.
```
gjn: Doc should also include google spreadsheet source as possible way

scripts/translation2json.py
```diff
@@ -0,0 +1,115 @@
...
+                csv_reader = csv.DictReader(csv_file, delimiter=delimiter, quotechar=quotechar)
+                translations = _process_csv_file(csv_reader, translations)
+
+    _save_translations(translations, args.output_folder)
+
+
+def _get_csv_reader_from_gspread(key, key_file, delimiter=',', quotechar='"', **kwargs):
+    if 'DRIVE_USER' in os.environ.keys() and 'DRIVE_PWD' in os.environ.keys():
+        drive_user, drive_password = os.environ['DRIVE_USER'], os.environ['DRIVE_PWD']
+        gc = gspread.login()
```
gjn: When I try to use it with DRIVE_USER and DRIVE_PWD, I get the following error:

```
Traceback (most recent call last):
  File "scripts/translation2json.py", line 115, in <module>
    main(parser.parse_args())
  File "scripts/translation2json.py", line 33, in main
    quotechar=quotechar
  File "scripts/translation2json.py", line 49, in _get_csv_reader_from_gspread
    gc = gspread.login()
TypeError: login() takes exactly 2 arguments (0 given)
make: *** [translate] Error 1

```

Jenselme: In our version of the script, I changed how we get the document from the drive. Since your document is publicly available, I make a get on the export as CSV URL (https://docs.google.com/spreadsheets/d/1F3R46w4PODfsbJq7jd79sapy3B7TXhQcYM7SEaccOA0/export?format=csv&gid=0) to get the document. That allowed me to:
- Use your document without authentication
- Simplify the script.

Are you OK to drop authentication to access this document?

gjn: As long as it works, I'm fine with it.

But then we should drop code that isn't needed anymore.

scripts/translation2js.yaml
Jenselme: Currently this is not. I should add it. However, I see the point of the `empty.json` (not to include translations that are not ready if I'm right), I fail to see the utility of `langs`: langs are activated by topics so even if the translation for a lang is not ready in your spreadsheet it won't be activated on the topic.

Should I still support both?

gjn: yes. Basic idea is to only update languages that we want (translations are not always finished at the same time).

Jenselme:
Script updated.

gjn:
@Jenselme Could I ask you for a rebase? I'd like the update.

Jenselme:
@gjn Just done the rebase. Can you merge it quickly so I don't have to do another one?

gjn:
Thanks alot @Jenselme .

If I do a `make translate`, I get the following error:

```
.build-artefacts/python-venv/bin/python scripts/translation2json.py \
            --files "https://docs.google.com/spreadsheets/d/1F3R46w4PODfsbJq7jd79sapy3B7TXhQcYM7SEaccOA0/export?format=csv&gid=0" \
            --languages "'[\"de\", \"en\", \"fr\", \"it\", \"rm\"]'" \
            --empty-json-file src/locales/empty.json \
            --output-folder src/locales
Traceback (most recent call last):
  File "scripts/translation2json.py", line 182, in <module>
    main(parser.parse_args())
  File "scripts/translation2json.py", line 44, in main
    quotechar=quotechar)
  File "scripts/translation2json.py", line 75, in _process_files
    _process_csv_file(csv_reader, translations, empty_json)
  File "scripts/translation2json.py", line 104, in _process_csv_file
    json_key = row['key'] if 'key' in row else row['']
KeyError: ''
make: *** [translate] Error 1
```

Does it work on your side?

Jenselme:
> Does it work on your side?

It worked yesterday. I see that you added again a `msgid` column header to your gspread sheet. This wasn't handled correcly. I just updated the PR. It works now (with or without the `msgid` header).

gjn:
That's a good update. @Jenselme thanks alot for the improved script.

PR #2344, opened 2015-05-06
Makefile

Jenselme:
It includes:
- it forces the use of the `PHANTOMJS_BIN` variable before using karma. We experienced trouble to launch karma without this variable
- a reformat of long lines
- Use npm to install some dependencies instead of git
- a switch to a system wide install of closure compiler

If you want only parts of these modifications, feel free to ask me to remove some commits.

PS: we are thinking of a switch from make to gulp in a close future. If you are interested, let me know.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,18 @@ help:
 all: lint dev prod apache testdev testprod deploy/deploy-branch.cfg fixrights
 
 .PHONY: prod
-prod: prd/lib/ prd/lib/build.js prd/style/app.css prd/geoadmin.appcache prd/index.html prd/mobile.html prd/embed.html prd/img/ prd/style/font-awesome-3.2.1/font/ prd/locales/ prd/checker prd/robots.txt
+prod: prd/lib/ \
+	prd/lib/build.js \
+	prd/style/app.css \
+	prd/geoadmin.appcache \
+	prd/index.html \
+	prd/mobile.html \
+	prd/embed.html \
+	prd/img/ \
+	prd/style/font-awesome-3.2.1/font/ \
+	prd/locales/ \
+	prd/checker \
+	prd/robots.txt
 
 .PHONY: dev
 dev: src/deps.js src/style/app.css src/index.html src/mobile.html src/embed.html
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -308,12 +439,9 @@ $(addprefix .build-artefacts/annotated/, $(SRC_JS_FILES) src/TemplateCacheModule
 	${PYTHON_CMD} .build-artefacts/python-venv/bin/pip install "gspread==0.2.2"
 	touch $@
 
-.build-artefacts/ol-requirements-installation.timestamp: .build-artefacts/python-venv
```
gjn: Why was this target removed? Isn't the regex module needed?

Makefile
```diff
@@ -273,22 +381,42 @@ node_modules: package.json
...
 
-.build-artefacts/app.js: .build-artefacts/js-files .build-artefacts/closure-compiler/compiler.jar .build-artefacts/externs/angular.js .build-artefacts/externs/jquery.js
+.build-artefacts/app.js: .build-artefacts/js-files \
+	    .build-artefacts/externs/angular.js \
+	    .build-artefacts/externs/jquery.js
 	mkdir -p $(dir $@)
-	java -jar .build-artefacts/closure-compiler/compiler.jar $(SRC_JS_FILES_FOR_COMPILER) --compilation_level SIMPLE_OPTIMIZATIONS --jscomp_error checkVars --externs externs/ol.js --externs .build-artefacts/externs/angular.js --externs .build-artefacts/externs/jquery.js --js_output_file $@
-
-$(addprefix .build-artefacts/annotated/, $(SRC_JS_FILES) src/TemplateCacheModule.js): .build-artefacts/annotated/%.js: %.js node_modules
+	closure-compiler $(SRC_JS_FILES_FOR_COMPILER) \
```
gjn: This does not work anymore on our environment.

gjn:
As mentioned in the comments, I can't build the application anymore using this PR (closure compiler global installs are not possible on our system; we apply strict policies on our development machines), so I can't verify all parts of it.

In general, I like the improved readability by breaking down long lines into smaller ones.

I don't see a benefit by switching from direct git imports to git imports wrapped around npm. It adds another level/redirection that doesn't add substantial benefits, but another level that might fail.

As mentioned above, we strongly prefer local installs over global installs. The builds need to work on a variaty of systems that are under strict policies regarding global installs, among others, our production systems.

We also don't plan to switch to gulp. Make is working just fine.

Makefile
Jenselme: If it is, I wasn't able to find where. If you know, please tell me.

gjn: Might have been needed for the ol3 build in our original file, but now that you're using global closure install, is not needed anymore? I'm just speculating here..

Jenselme:
> As mentioned in the comments, I can't build the application anymore using this PR (closure compiler global installs are not possible on our system; we apply strict policies on our development machines), so I can't verify all parts of it.

I will add a commit to revert this. I don't want to squash yet since I am afraid it will delete the comments on the commits.

> I don't see a benefit by switching from direct git imports to git imports wrapped around npm. It adds another level/redirection that doesn't add substantial benefits, but another level that might fail.

I wanted to get rid of git clones. I think it makes sense for bootstrap since it is available on npm. I changed the other one because I thought it was more logical but I agree, this may create more problem this way. Maybe we should go back.

> As mentioned above, we strongly prefer local installs over global installs. The builds need to work on a variaty of systems that are under strict policies regarding global installs, among others, our production systems.

We don't have the same requirements. We prefer installing packages through the package manager of our distribution. I won't propose patches that implies such changes.

Makefile
Jenselme: I have just retested without the global closure install and everything seems to work fine without the regex module.

gjn:
Thanks alot for understanding.

Makefile
gjn: Than maybe it's not needed anymore. ol3 build system is quite in flux :) Thanks alot for the tests.

Jenselme:
Do you want me to squash or remove some commits (or maybe open a new PR to keep the history)?

Jenselme:
Any plans on merging it? The current presentation of your Makefile is annoying when I make PRs that add configuration variables.

gjn:
@Jenselme Can you rebase. I'll git a crack at reviewning then.

Jenselme:
Just rebased. I will wait for this PR to be merged before rebasing the others.

gjn:
It's working good. Thanks alot!

PR #2348, opened 2015-05-08
New background selector directive.

cedricmoullet:
Test link: https://mf-geoadmin3.dev.bgdi.ch/moc_layerselectordiv/
Comments: the background layer list has been hardcoded since the 4 bg layers are present in all topics. service could be adapted accordingly


gjn:
Ah. Same name as my new layer selector. But exactly the same name :=)

src/components/layerselector/style/layerselector.less
```diff
@@ -0,0 +1,138 @@
...
+    height: 32px;
+  }
+  
+  @media (max-width: @screen-phone) {
+    width: 48px;
+    height: 32px;
+  }
+}
+
+.bgLayer0 {
```
loicgasser: Regarding custom class names convention is: prefix with ga and don't use camel case unsless it's an id.
`ga-import-wms-container` for instance.

src/components/layerselector/LayerSelectorDirective.js
```diff
@@ -0,0 +1,147 @@
...
+        templateUrl: 'components/layerselector/partials/layerselector.html',
+        scope: {
+          map: '=gaLayerSelectorMap'
+        },
+        link: function(scope, elt, attrs) {
+          scope.isClosed = true;
+          var map = scope.map;
+          var firstLayerChangeEvent = true;
+          // Background layers are always the same
+          scope.backgroundLayers = [
```
loicgasser: `var backgroundLayers = gaLayers.getBackgroundLayers().getArray();` and then `backgroundLayers.push({id: 'voidLayer', label: 'void_layer'});` as it was done previously.
If we had/remove a bg layer we don't need to change the code.

src/components/layerselector/partials/layerselector.html
```diff
@@ -0,0 +1,5 @@
+<div ng-repeat="layer in backgroundLayers">
+  <div ng-click="onClick($event)" ng-show="!isClosed || $first" class="bgLayerDefault {{'bgLayer' + $index}} {{getClass(layer)}}">
```
loicgasser: you could `ng-click="onClick(layer.id)" it would simplify the code here:
`scope.onClick = function(layerID) {`
Additionnaly I would use a button instead of a div for a clickable element.

src/components/layerselector/LayerSelectorDirective.js
```diff
@@ -0,0 +1,148 @@
...
+              layer.background = true;
+              layer.displayInLayerManager = false;
+              if (layers[0] && layers[0].background) {
+                layers.setAt(0, layer);
+              } else {
+                layers.insertAt(0, layer);
+              }
+            }
+            gaPermalink.updateParams({bgLayer: layerid});
+            scope.currentLayer = layerid;
```
loicgasser: You don't need to attach it to the scope as it is not used in the html of the directive.

src/components/layerselector/LayerSelectorDirective.js
```diff
@@ -0,0 +1,148 @@
...
+            isOfflineToOnline = !offline;
+          });
+
+          scope.getClass = function(layerid) {
+            var splitLayer = layerid.id.split('.');
+            return splitLayer[splitLayer.length - 1];
+          };
+
+          // swissimage or pixelmap-color are always in the two first positions
+          scope.reorderBgLayer = function(layerid) {
```
loicgasser: you don't need to attach it to the scope.

src/components/layerselector/style/layerselector.less
```diff
@@ -0,0 +1,138 @@
...
+  }
+  
+  &:hover .bgLayer3-text {
+    display: block;
+    background-color: rgba(255, 255, 255, 1);
+  }
+
+}
+
+.pixelkarte-farbe {
```
loicgasser: you could factorize shared properties

src/components/layerselector/style/layerselector.less
```diff
@@ -0,0 +1,138 @@
...
+  
+  @media (max-width: @screen-tablet) {
+    bottom: 24px;
+  }  
+  
+  @media (max-width: @screen-phone) {
+    bottom: 24px;
+  } 
+
+  .bgLayer0-text {
```
loicgasser: some css factorization here as well

cedricmoullet:
@gjn @loicgasser @davidoesch ready for review and test. Branch has been updated.

loicgasser:
I believe Swissimage gets stuck at the beggining of the list. (at least on chrome mobile)

1. Open app
2. Open menu
3. Select swissimage
4. Open menu
5. Reselect Swissimage

Result:
Swissimage is the first result in the list. (e.g. displayed img when the menu is closed)

Expected Result:
Pixelfarbe is the first result.

Also, I think we should use the default bglayer of the topic for as a swissimage counterpart.

Do you want a hand to put the tests back? It should not be too hard if you are a little used too it.

cedricmoullet:
Thanks. nice bug ;-) I'll try to fix it.

cedricmoullet:
@loicgasser 's bug is now fixed. Happy to get some helps for the tests.

kandre:
I am testing it

davidoesch:
@kander: test enviro https://trello.com/c/8Iup3BkK/14-virtual-environment
@Cedricmoullet: mobile is not yet fixed I assume?

cedricmoullet:
@davidoesch should be ok on mobile. Do you see any issue ?

davidoesch:
@cedricmoullet we don't show the selector in the map on mobile. takes to much space. background is switched in the settings menue

there
unbenannt
fore

kandre:
It works really well ! 

Cannot find bugs. Tested on Win7 (all browsers), mac (all browsers), Ipad(Safari an chrome) and our samsung nexus (Chrome an FF)

cedricmoullet:
No David. If you want that, we duplicate code, duplicate risks and complexify the application. It works well on mobile and it's also an use case to switch background on mobile. 
We could make the size smaller on mobile. This would also remove the text.

ponceta:
Possible improvements :

 - darken the arrow ( It took me 30 sec to see it)
 - make it clickable
 - On click on the arrow expand the bg selector (reveal the other background choice options)
 - On click on the aerial imagery <b>switch to aerial</b> & expand the menu 

If menu is expanded - arrow is facing down - click on the arrow = reduce the bg selector

cedricmoullet:
jenkins test

loicgasser:
j'ai rebase et squash mais je vois plus le background layer selector.... pas push, t'as controllé de ton côté que ça marche aussi sur disons firefox

cedricmoullet:
@loicgasser Have you tried to click on the "switzerland full extent" icon ?

loicgasser:
Apart from the icon, is this PR subject to more changes? Ready for final review? I really like the new bg selector. good design!

cedricmoullet:
I just finalize the icon and then it will be ready for review.

2015-05-28 9:10 GMT+02:00 Gasser Loïc <notifications@github.com>:

> Apart from the icon, is this PR subject to more changes? Ready for final
> review? I really like the new bg selector. good design!
>
> —
> <https://github.com/geoadmin/mf-geoadmin3/pull/2348#issuecomment-106203641>
> .
>



-- 

cedricmoullet:
@loicgasser @gjn @davidoesch  ready for review. test link has been updated

src/components/backgroundselector/BackgroundSelectorDirective.js
```diff
@@ -0,0 +1,138 @@
...
+        templateUrl:
+            'components/backgroundselector/partials/backgroundselector.html',
+        scope: {
+          map: '=gaBackgroundSelectorMap'
+        },
+        link: function(scope, elt, attrs) {
+          scope.isBackgroundSelectorClosed = true;
+          scope.embed = gaBrowserSniffer.embed;
+          var map = scope.map;
+          var firstLayerChangeEvent = true;
```
gjn: Instead of this guard variable, you could use `scope.currentTopic` variable...and this currentTopic variable should be function variable and not attached to scope.

src/components/backgroundselector/BackgroundSelectorDirective.js
```diff
@@ -0,0 +1,138 @@
...
+          var isOfflineToOnline = false;
```
gjn: Good to think about offline!

src/components/backgroundselector/BackgroundSelectorDirective.js
```diff
@@ -0,0 +1,138 @@
...
+            scope.isBackgroundSelectorClosed = true;
+          });
+
+          scope.$watch('currentLayer', function(newVal, oldVal) {
+            if (oldVal !== newVal) {
+              setCurrentLayer(newVal);
+            }
+          });
+
+          scope.onClick = function(layerid) {
```
gjn: It's usually better to name functions on what they do, not on what might trigger them. So maybe something like 'activateBackgroundLayer'.

src/components/backgroundselector/BackgroundSelectorDirective.js
```diff
@@ -0,0 +1,138 @@
...
+              scope.backgroundLayers = defaultBgOrder.slice(0);
+            } else {
+              scope.isBackgroundSelectorClosed = true;
+              if (scope.currentLayer != layerid) {
+                scope.currentLayer = layerid;
+              }
+            }
+          };
+
+          scope.onClickBt = function() {
```
gjn: dito. Could be 'toggleMenu' or so.

src/index.mako.html
```diff
@@ -268,6 +265,7 @@
            ga-time-selector-options="options">
       </div>
     </div>
+    </div>
```
gjn: was this a missing div? (I don't see a corresponding <div>)

gjn:
Code looks good to me, good work! Just added a few comments.

Would be great if you could keep the unittests :)

Test on IE9 as well and it works fine.

src/components/backgroundselector/BackgroundSelectorDirective.js
```diff
@@ -0,0 +1,138 @@
...
+            // Specific use case when we go offline to online, in this use
+            // case we want to keep the current bg layer.
+
+            var bgLayer;
+            if (firstLayerChangeEvent) {
+              bgLayer = gaPermalink.getParams().bgLayer;
+              firstLayerChangeEvent = false;
+            }
+            if ((!bgLayer && !scope.currentLayer) ||
+              (scope.currentTopic && (scope.currentTopic != data.topicId))) {
```
gjn: currentTopic does not need to be attached to scope.

gjn:
Regarding UI, I don't like the thick border. While it is the same kind of tool, it could make sense to not have the border (and more visible real-estate to show the user). But it's a matter of taste.

src/components/backgroundselector/BackgroundSelectorDirective.js
cedricmoullet: Not sure that I understand. What is the "guard variable" ? I see that I currentTopic shouldn't be in the scope.

gjn: Slang, sorry. it's guarding against multiple runs of some code.

Anyway, it was the same in old code. I just said that you don't need it because currentTopic is basically doing exactly the same thing (currentTopic will be undefined when 'gaLayersChange' event is first triggered, after that it will be set. So below, instead of using `if (firstLayerChangeEvent)`, you could use `if (!scope.curentTopic)` thus saving one variable. It's minor.

loicgasser:
jenkins test

loicgasser:
browser stack tests are failing again...

src/components/backgroundselector/BackgroundSelectorDirective.js
cedricmoullet: Not really. A topic changes can also happen afterwards. In this case, we also need to change the background layer.

gjn: yeah, but then currentTopic will be set, and the part guarded by `if (scope.currentTopic)` won't be executed again. Only the first time. The idea is the remove firstLayerChangeEvent variable and us currentTopic to do the same thing.

src/components/backgroundselector/BackgroundSelectorDirective.js
cedricmoullet: ok

src/components/backgroundselector/BackgroundSelectorDirective.js
cedricmoullet: ok

src/index.mako.html
cedricmoullet: good catch

src/components/backgroundselector/BackgroundSelectorDirective.js
cedricmoullet: ok, i'll have a try

2015-05-28 20:32 GMT+02:00 Gilbert Jeiziner <notifications@github.com>:

> In src/components/backgroundselector/BackgroundSelectorDirective.js
> <https://github.com/geoadmin/mf-geoadmin3/pull/2348#discussion_r31263654>:
>
> > +
> > +  module.directive('gaBackgroundSelector',
> > +    function($document, gaPermalink, gaLayers, gaLayerFilters,
> > +      gaBrowserSniffer) {
> > +      return {
> > +        restrict: 'A',
> > +        templateUrl:
> > +            'components/backgroundselector/partials/backgroundselector.html',
> > +        scope: {
> > +          map: '=gaBackgroundSelectorMap'
> > +        },
> > +        link: function(scope, elt, attrs) {
> > +          scope.isBackgroundSelectorClosed = true;
> > +          scope.embed = gaBrowserSniffer.embed;
> > +          var map = scope.map;
> > +          var firstLayerChangeEvent = true;
>
> yeah, but then currentTopic will be set, and the part guarded by if
> (scope.currentTopic) won't be executed again. Only the first time. The
> idea is the remove firstLayerChangeEvent variable and us currentTopic to do
> the same thing.
>
> —
> <https://github.com/geoadmin/mf-geoadmin3/pull/2348/files#r31263654>.
>



-- 

cedricmoullet:
@loicgasser @gjn @davidoesch ready for final review. Waiting feedback from @jroland01

davidoesch:
ok

p1d1d1:
Sorry for jumping into the discussion. 
@cedricmoullet you'll get less "ah like in google maps" ;-)
I like a lot the new design, really great job.
I like a little less the strong border around the icon.
And it would also be great IMHO if the icon picture could change according to the selected background.
Just my 2 cents and IMHO.

tigerfoot:
After testing, I would have use the same layout for mobile and desktop (the round icon).
It's unify the experience, and see my comments below about how profan interpret it.
Having its background changing with the type of background would be the top notch feature :-)

I've shown the new thing to profan users, and they've believed that the desktop background selector was a kind of overview. I guess this confusion come to some mapfish portal layout where the overview is located at the bottom left.

davidoesch:
Dear all 

- icon desktop mobile :
According to usability tests : in desktop a larger icon is needed to make it obvious to the user that this is a background switcher.

- overview confusion : yes this is related to ancient function on map fish apps. However the majority of mapping apps like  gmaps HERE bing search.ch etc use now a bg switcher instead of an overview 

- changing the icon content based on the active bg: this would work if u only have to deal with 2 bg types.... We have four. Rearranging the order is disturbing....

cedricmoullet:
David's comment make sense.

2015-05-30 12:38 GMT+02:00 davidoesch <notifications@github.com>:

> Dear all
>
>    -
>
>    icon desktop mobile :
>    According to usability tests : in desktop a larger icon is needed to
>    make it obvious to the user that this is a background switcher.
>    -
>
>    overview confusion : yes this is related to ancient function on map
>    fish apps. However the majority of mapping apps like gmaps HERE bing
>    search.ch etc use now a bg switcher instead of an overview
>    -
>
>    changing the icon content based on the active bg: this would work if u
>    only have to deal with 2 bg types.... We have four. Rearranging the order
>    is disturbing....
>
> —
> <https://github.com/geoadmin/mf-geoadmin3/pull/2348#issuecomment-107019804>
> .
>



-- 

src/index.mako.html
```diff
@@ -214,6 +210,9 @@
         <div ga-geolocation ga-geolocation-map="map"></div>
   % endif
         <div id="zoomButtons"></div>
+  % if device == 'mobile' or device == 'embed':
+        <div ga-background-selector ga-background-selector-map="map" class="ga-bg-mobile"></div>
```
oterral: Adding a css class here suggest the css  is defined on the app.css, but it is not. I think this class should be added dynamically in the directive. The same for `ga-bg-desktop`.

oterral:
The js code lgtm but I think the css could be improved in order to match what was done for other map buttons (top, bottom managment).

and with a minor modification we could have a nice animation to show/hide the list of layers.

cedricmoullet:
jenkins test

cedricmoullet:
jenkins test

gjn:
Win.

PR #2351, opened 2015-05-11
Fix profile display on FF

loicgasser:
Because of a silghtly different implementation of the box model profile is hidden on FF.

https://github.com/geoadmin/mf-chsdi3/pull/1425


loicgasser:
trivial self merge

PR #2358, opened 2015-05-15
Rename layerselector to background selector

gjn:
This is a modes proposal to rename this layerselector to backgroundselector because I think the name fits better the purpose of the component.

:=)

@cedricmoullet This is a PR for your branch.


cedricmoullet:
I had in mind to propose other layers ... but it is fine for now.

PR #2361, opened 2015-05-19
Update to OL 3.5.0

ponceta:
Test link : https://mf-geoadmin3.dev.bgdi.ch/ltpoa/

Check all browsers, all functions.


ponceta:
@loicgasser 

ERR! No matching symbol found: ol.source.GeoJSON

Any idea?

Rebasing again seems too tricky, I let you update it for our new current master!

loicgasser:
ol.source.GeoJSON does not exist anymore. need to use ol.source.Vector and ol.format.GeoJSON instead.

PR #2382, opened 2015-06-02
Fix translate target

oterral:
Fix the new authetification to access to google drive api .

Mail is coming with new user/password

Info on oauth2 -> http://gspread.readthedocs.org/en/latest/oauth2.html


scripts/translation2js.py
```diff
@@ -7,7 +7,8 @@
 Usage: ./env/bin/python translation2js.py /home/ltmoc/mf-geoadmin3/src/locales/ 
 """
 
-import os, sys, codecs, re, gspread
+import os, sys, codecs, re, json, gspread
```
procrastinatio: unused `json`?

procrastinatio:
OK.

Issue #2389, opened 2015-06-04
Highlight of search not fully working


gjn:
When typing 'Bern' not all instances of bern in the result list are highlighted.

PR #2405, opened 2015-06-10
Fix highlight in search inside b tags

gjn:
This is to fix https://github.com/geoadmin/mf-geoadmin3/issues/2389

@oterral Please review for today's deploy.


oterral:
lgtm

oterral:
Nice tests

Issue #2388, opened 2015-06-03
Feedback widget broken


procrastinatio:
To reproduce:

1. Start geo.admin.ch with http://map.geo.admin.ch/?widgets=feedback
2. Try to select a gemometry --> warning 'veuillez renseigner ce champs'
3. Type something in the text box --> message sent. Over

Expected :

Click on the 'report problem' link (top right corner)

oterral:
it seems it's because the popup loose the focus.

PR #2407, opened 2015-06-10
Fix feedback startup

gjn:
This is a fix for. https://github.com/geoadmin/mf-geoadmin3/issues/2388

There were to problems: the type of the buttons in the draw dialog was not specified. The initialisation of the draw dialog was not correct and therefore the draw directive was not correctly initialized on the feedback dialog.

@oterral Last one for today's deploy. Thanks for the quick review/test/merge.

### DIFF
--- a/src/js/MainController.js
+++ b/src/js/MainController.js
@@ -97,6 +97,9 @@
       gaRealtimeLayersManager($scope.map);
 
       var initWithPrint = /print/g.test(gaPermalink.getParams().widgets);
+      var initWithFeedback = /feedback/g.test(gaPermalink.getParams().widgets);
+
+      gaPermalink.deleteParam('widgets');
 
       $rootScope.$on('gaTopicChange', function(event, topic) {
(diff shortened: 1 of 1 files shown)


gjn:
jenkins test

src/components/draw/partials/draw.html
```diff
@@ -53,6 +55,7 @@
             ng-if="options.showExport"
             ng-disabled="!canExport()"
             ng-click="exportKml($event)"
+            type="button"
```
oterral: all these type=""button" are really needed?

gjn: I think so, otherwise, when the draw directive is included in the feedback form, a submit is issued.

I verify again.

gjn: They were indeed not needed. I'll upate the PR.

gjn:
Adapted according to comments. Thanks alot.

oterral:
thx for the fix

PR #2415, opened 2015-06-12
[WIP] New layer selector

gjn:
Just for reference. And if someone wants to continue during my absence.

Wacked up css, only search added, etc.

I think overall appraoch to re-use current layermanager and current search in the new layer selector is probably not a good idea. We're probably way better off to start from scratch with those elements. :)

Ref: 
http://invis.io/DA2RHVRKS

### DIFF
--- a/src/components/ContextProposalService.js
+++ b/src/components/ContextProposalService.js
@@ -1,0 +1,52 @@
+(function() {
+  goog.provide('ga_context_proposal_service');
+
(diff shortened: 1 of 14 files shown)


oterral:
jenkins test

oterral:
@gjn can you send me the username/password for connect to invis.io ?

oterral:
@gjn forget I've found it

PR #2430, opened 2015-06-16
Do not use cached ur for legend 

procrastinatio:
`cachedApiUrl`is updated at every `deploy`and hold in cache for one year, whicle `apiUrl` is never cached.

Some legends (BAZL's _luftfahrthindernis_) use the BOD table `datenstand` which is updated and deployed every 4 hours. So its content may change.

See https://github.com/geoadmin/mf-chsdi3/issues/1511


procrastinatio:
jenkins test

PR #2404, opened 2015-06-09
Mapfish print v3

Jenselme:
Allow to print with mapfish print v3 with the mf-geoadmin3 interface. This component fits our needs. Let me know if I must change things so this PR can be merged.

- mapfish-print war used: https://github.com/Geoportail-Luxembourg/geoportailv3/blob/master/print/print-servlet.war
- mapfish-print configuration used: http://jenselme.perso.centrale-marseille.fr/visible/geojb.zip As of this moment, only the configuration for a4_portrait is complete (with northmap and legend page only if legend is asked). Based on: https://github.com/Geoportail-Luxembourg/geoportailv3/tree/master/print/print-apps/geoportailv3
- Based on components from [ngeo](https://github.com/camptocamp/ngeo):
  - https://github.com/camptocamp/ngeo/blob/master/src/services/print.js
  - https://github.com/camptocamp/ngeo/blob/master/src/services/printutils.js
- Pending PR at ngeo (applied on this version):
  - [Support map rotation for printing](https://github.com/camptocamp/ngeo/pull/231)
  - [Allow tiles from a TileWMS to be printed](https://github.com/camptocamp/ngeo/pull/226)

Side note: to make the war work correctly with java8, I had to delete `<deploy_print_root>/WEB-INF/lib/jdtcore-3.1.0.jar` and replace it by the 3.1.1 version of jdtcore: http://mvnrepository.com/artifact/org.eclipse.jdt/core/3.1.1

PR #2433, opened 2015-06-17
Update closure compiler to 20150609.0.0

### LINKED ISSUE #2363: Update closure-compiler
Hi,

I have just seen that despite the name lastest, the closure-compiler version you use is quite old: compiler-20131014.zip. I tried with the 20150505 version and it doesn't work. I have 305 errors of the same kind: `ERROR - Closure dependency methods(goog.provide, goog.require, etc) must be called at file scope.` Moving the `goog.provide` outside the anonymous function solves this.

Do you have any plans on updating it? I have no idea how much impact this migth have. I should also be able to help if needed.

Exact compiler version:
```
java -jar node_modules/google-closure-compiler/compiler.jar --version
Closure Compiler (http://github.com/google/closure-compiler)
Version: v20150505
Built on: 2015/05/06 16:19
```

Jenselme:
Close #2363


test/karma-conf.mako.js
```diff
@@ -26,7 +26,7 @@ files = [
   'lib/EPSG32631.js',
   'lib/EPSG32632.js',
   '../test/closure-loader-globals.js',
-  'lib/ol-debug.js',
+  'lib/ol.js',
```
oterral: leave `ol-debug.js` in dev mode

Jenselme: PR updated.

oterral:
I see you add node modules for closure but you still leave some use of .build-artefacts/closure in the makefile. I don't think we need 2 versions of closures (in .build-artefacts and in node_modules)

Have you try to just update the `.build-artefacts/closure-compiler/compiler-latest.zip ` target  to the laste version ? like this
```
.build-artefacts/closure-compiler/compiler-latest.zip:
  mkdir -p $(dir $@)
  wget -O $@ http://dl.google.com/closure-compiler/compiler-latest.zip
  touch $@
```

Jenselme:
I did not tried to update the closure-compiler in the Makefile. We rely on glup to build the software. When I make a pull request I try do update the Makefile but sometimes (like here), I may forgot.

I don't know what you prefer: the updated Makefile with a new version in `.build-artefacts` or in `node_modules`. Let me know and I will update the PR accordingly.

oterral:
I prefer to use the `.build-artefacts` it's closer to what ngeo does

Jenselme:
PR updated

oterral:
There is stil a use of google-closure-library with `node-modules`, can you remove it and use `.build-artefacts` ?

Jenselme:
Forgot that, sorry. PR updated.

Makefile
```diff
@@ -1,5 +1,5 @@
 SRC_JS_FILES := $(shell find src/components src/js -type f -name '*.js')
-SRC_JS_FILES_FOR_COMPILER = $(shell sed -e ':a' -e 'N' -e '$$!ba' -e 's/\n/ --js /g' .build-artefacts/js-files | sed 's/^.*base\.js //')
+SRC_JS_FILES_FOR_COMPILER = $(shell sed -e ':a' -e 'N' -e '$$!ba' -e 's/\n/ --js /g' .build-artefacts/js-files | sed -r 's/(.*)/ --js \1/g')
```
oterral: I don't think we don't  need this modification because that add  the base.js file to the single build file. And we don't need the base.js file.

Jenselme: I cannot remember if I had to do this for the update of closure compiler or to solve goog.require issue for print (#2404). I just tested it seems to work fine without. Maybe this is a particularity of our environment.

PR updated.

oterral:
Thx for your patience @jenselme it's really appreciated

Jenselme:
You are welcome.

PR #2437, opened 2015-06-19
Considering min- and max-resolution for geojson layer

procrastinatio:
In order to test:

Build `mf-geoadmin3`using your own `http://mf-chsdi3.dev.bgdi.ch/mom_geojson_scale` for service (export API_URL...). It is using the `bod_ltmom_master`DB. You may want change and edit values in `layerjs`if you want.

I have modified the `bod`DB (in `bod_ltmom_ master`) as follow:

layer    | min. res | max. res
----------|-----------|-------------
ch.bafu.hydroweb-messstationen_gefahren | 0.01  | 100
ch.bafu.hydroweb-messstationen_grundwasser  | 100  |  --
ch.bafu.hydroweb-messstationen_temperatur  | --  | 100
ch.bafu.hydroweb-messstationen_zustand  | -- | --

Note, It breaks the geo.admin.ch rule staing that a layer is visible at all scales

### DIFF
--- a/src/components/map/MapService.js
+++ b/src/components/map/MapService.js
@@ -918,6 +918,8 @@
             var fullUrl = gaGlobalOptions.ogcproxyUrl + layer.geojsonUrl;
             var olSource = new ol.source.Vector();
             olLayer = new ol.layer.Vector({
+              minResolution: layer.minResolution,
+              maxResolution: layer.maxResolution,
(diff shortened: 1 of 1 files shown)


AFoletti:
LGTM! Many thanks!
but I'll let a developer merge it ;)

PR #2440, opened 2015-06-19
[WIP] Use background layers provided by the current topic

Jenselme:
- Choice of background layers is not hard coded
- Topics without background layers don't create an error
- Add a signal which is broadcasted when layers are loaded. Necessary to know when we can access the list of background layers.

### DIFF
--- a/src/components/backgroundselector/BackgroundSelectorDirective.js
+++ b/src/components/backgroundselector/BackgroundSelectorDirective.js
@@ -33,6 +33,7 @@ goog.require('ga_topic_service');
             elt.addClass('ga-bg-desktop');
           }
 
+          var getLayersDeffered = $q.defer();
           var map = scope.map;
           var isOfflineToOnline = false;
 
@@ -93,11 +94,12 @@ goog.require('ga_topic_service');
           };
 
           var dereg = scope.$on('gaLayersChange', function(event, newLayers) {
+            getLayersDeffered.resolve();
             updateBgLayer();
             dereg();
           });
 
-          $q.all([gaTopic.getTopics(), gaLayers.getLayers()]).then(function() {
+          $q.all([gaTopic.getTopics(), getLayersDeffered.promise]).then(function() {
             // If the layers config is loaded
             if (gaLayers.getBackgroundLayers()) {
               defaultBgOrder = [];
--- a/src/components/backgroundselector/BackgroundSelectorDirective.js
+++ b/src/components/backgroundselector/BackgroundSelectorDirective.js
@@ -12,7 +12,7 @@ goog.require('ga_topic_service');
   ]);
 
   module.directive('gaBackgroundSelector',
-    function($document, gaPermalink, gaLayers, gaLayerFilters,
+    function($document, $q, gaPermalink, gaLayers, gaLayerFilters,
         gaBrowserSniffer, gaTopic) {
       return {
         restrict: 'A',
@@ -97,24 +97,22 @@ goog.require('ga_topic_service');
             dereg();
           });
 
-          gaTopic.getTopics().then(function() {
-            scope.$on('gaLayersChange', function() {
-              // If the layers config is loaded
-              if (gaLayers.getBackgroundLayers()) {
-                defaultBgOrder = [];
-                gaLayers.getBackgroundLayers().forEach(function(bgLayer) {
-                  defaultBgOrder.push({
-                    id: bgLayer.id,
-                    label: bgLayer.label
-                  });
-                });
+          $q.all([gaTopic.getTopics(), gaLayers.getLayers()]).then(function() {
+            // If the layers config is loaded
+            if (gaLayers.getBackgroundLayers()) {
+              defaultBgOrder = [];
+              gaLayers.getBackgroundLayers().forEach(function(bgLayer) {
                 defaultBgOrder.push({
-                  id: 'voidLayer',
-                  label: 'void_layer'
+                  id: bgLayer.id,
+                  label: bgLayer.label
                 });
-                updateBgLayer(true);
-              }
-            });
+              });
+              defaultBgOrder.push({
+                id: 'voidLayer',
+                label: 'void_layer'
+              });
+              updateBgLayer(true);
+            }
           });
 
           scope.$on('gaPermalinkChange', function(event) {
--- a/src/components/backgroundselector/BackgroundSelectorDirective.js
+++ b/src/components/backgroundselector/BackgroundSelectorDirective.js
@@ -25,6 +25,7 @@ goog.require('ga_topic_service');
           scope.isBackgroundSelectorClosed = true;
           var mobile = gaBrowserSniffer.mobile;
           scope.desktop = !gaBrowserSniffer.embed && !mobile;
+          scope.backgroundLayers = [];
 
           if (mobile) {
             elt.addClass('ga-bg-mobile');
@@ -35,11 +36,7 @@ goog.require('ga_topic_service');
           var map = scope.map;
           var isOfflineToOnline = false;
 
-          var defaultBgOrder = [
-              {id: 'ch.swisstopo.swissimage', label: 'bg_luftbild'},
-              {id: 'ch.swisstopo.pixelkarte-farbe', label: 'bg_pixel_color'},
-              {id: 'ch.swisstopo.pixelkarte-grau', label: 'bg_pixel_grey'},
-              {id: 'voidLayer', label: 'void_layer'}];
+          var defaultBgOrder = [];
 
           scope.backgroundLayers = defaultBgOrder.slice(0);
 
@@ -65,7 +62,7 @@ goog.require('ga_topic_service');
             }
           }
 
-          var updateBgLayer = function(topic) {
+          var updateBgLayer = function(onGaTopicChange) {
             // Determine the current background layer. Strategy:
             //
             // On gaLayersChange event we receive then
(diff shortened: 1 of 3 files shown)


cedricmoullet:
Thanks for this PR.
Here some points to discuss:
1. @davidoesch this change reimplements the dynamic loading of the topic list from the catalog (it was hardcoded since it's always the same....). This implies that the order of the bg layer is topic dependent. The swissimage is not always the same. What do you think of this ?
2. @Jenselme Could you explain me how you plan to use it ? Since the css is almost hardcoded (swissimage, pixelkarte), how will you add other bg layers ?
3. Unit tests are failing. Will need to be fixed if we want to merge it.

davidoesch:
1) changing the bg order dependent on topic is hard to understand . As discussed previously in the UI session - we keep it as.

Jenselme:
1. In your use case, I guess that the order is (currently?) always the same. However, the current implementation of topics allows to change the background layers (order and layers) when changing topic. This is a functionality we need: we have many topics and they don't all have the same background layers. Moreover, the background layers are listed in an array. So in my opinion, it makes sense to rely on the order of the array instead of hard-coding it. Hard-coding seems also strange since you cannot use one functionality any more.
2. We will have to either adapt the CSS or further modify the code and do something like the images of topic which are not hard-coded.
3. I should get the habit of launching and fixing them before opening a PR

I guess I should have precise this is a work in progress. Since it changes how this works, I wanted feedback as early as possible from your part (to know if it has a chance to be merged or none since some patches may need some work to be truly generic).

davidoesch:
For all mf-geoadmin3 application we keep it in the same order. We have topics with a grey background.even then : for the user the order in the flyout is the same as for a topic with color map. This is user friendly and we can expand with many bg as needed.

So again : for mf-geoadmin3 there will no change be made.

Jenselme:
Even if you want to always keep the same order, I think it is possible to make sure that the API always returns the bg layers in the proper order. That would fit your and our needs.

Jenselme:
Fix tests

loicgasser:
Have you tried to use gaLayersChange event in loadLayersConfig instead?

oterral:
since #2471 is merged this PR needs refactoring

Jenselme:
I have just updated the PR to take into account #2471. I use `gaTopicChange` instead of creating a now event.

gjn:
The background selector does not work anymore. When I try to open it, it simply disappears...@Jenselme: any idea why?

Jenselme:
@gjn I have the same behavior when a topic doesn't define background layers (we have some of those). So my guess is that `gaLayers.getBackgroundLayers()` returns something falsey and so the layers are never added. See https://github.com/geoadmin/mf-geoadmin3/pull/2440/files#diff-5e15dcf3cc8d419d3f15c3e878955964R102 I don't know why you wouldn't get bg layers thought.

Maybe I should change the code to always add the void layer?

gjn:
hmm. Not sure it's that. Even the getBackgroundLayers returns something, the selector directive does not work anymore with this PR.

Jenselme:
Got it. It is again the problem I describe in #2520. I didn't see the effect because I still rely on the `$timeout` solution. Just rebased and updated the PR. This time, it should work.

src/components/backgroundselector/BackgroundSelectorDirective.js
```diff
@@ -95,11 +97,24 @@ goog.require('ga_topic_service');
             dereg();
           });
 
-          scope.$on('gaTopicChange', function(event, newTopic) {
-            // If the layers config is loaded
-            if (gaLayers.getBackgroundLayers()) {
-              updateBgLayer(newTopic);
-            }
+          gaTopic.getTopics().then(function() {
+            scope.$on('gaLayersChange', function() {
```
gjn: We don't have garantueed order of exection between the getTopics promise and the gaLayersChange event; so I guess that this implementation might break when topics take longer to load than the layers configuration. Also, gaLayersChange will be called on every language change, not only once at startup. But it won't get called on topic change!

Not sure how to address this. I need to understand more. What's the difference between the gaLayersChange part here and the one above?

Jenselme: > We don't have garantueed order of exection between the getTopics promise and the gaLayersChange event

Indeed. The change I proposed requires work not to encounter #2520 again.

> Also, gaLayersChange will be called on every language change, not only once at startup. But it won't get called on topic change!

So this correction is very incomplete.

> What's the difference between the gaLayersChange part here and the one above?

From what I understand from your code, here we should react on topic change to force the change of background layers when the topic change. The one above is there to use the background layer defined in the permalink once the layers are loaded. I just added the `gaLayersChange` event to be "sure" that the background layers are loaded once we have the topic.

How can we make this work then? To populate `defaultBgOrder` I need:
- the topic which defines which layers are background layer
- the layers

I think that for this to work properly, we must: on `gaTopicChange` wait for the new promise which gets the topic to resolve. In this promise wait for a promise which gets the layers to resolve. This implies to have a new promise for the layers. This is what I intended to do at first before being lazy and just using the existing `gaLayersChange`. Let the reaction to `gaLayersChange` untounched.

What do you think?

gjn: Yes, I was thinking about a new promise as well (for gaLayers)...

Anyway, AFAIK (to be verified) when gaTopicChange is fired, we already know that the getTopics() promise is resolved. So we can either use gaTopicChange or getTopics() promise. But this does not solve the overall problem about the order of execution.

We could try to create local promises, one (p1) resolved for first gaTopciChange event, another (p2) resolved for first gaLayersChange event, add those topics in $q.all([p1, p2]) construct (see https://docs.angularjs.org/api/ng/service/$q), this would assure that both gaTopicChange and gaLayersChange would have beend called at least once.

Jenselme: > Anyway, AFAIK (to be verified) when gaTopicChange is fired, we already know that the getTopics() promise is resolved. So we can either use gaTopicChange or getTopics() promise

I think so.

> We could try to create local promises, one (p1) resolved for first gaTopciChange event, another (p2) resolved for first gaLayersChange event, add those topics in $q.all([p1, p2]) 

I will try this solution.

Jenselme:
I updated the PR with a new promise. I took inspiration from https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/topic/TopicService.js#L19.

- Nothing squashed yet
- test of ga_backgroundselector_directive out of date

What do you think of this new solution?

src/components/map/MapService.js
```diff
@@ -793,7 +794,7 @@ goog.require('ga_urlutils_service');
             layers = undefined;
           });
         };
-        loadLayersConfig();
+        layersPromise = loadLayersConfig();
```
gjn: Is this really an actual promise returned?

src/components/map/MapService.js
```diff
@@ -827,6 +828,10 @@ goog.require('ga_urlutils_service');
           }
         };
 
+        this.getLayers = function() {
+          return layersPromise;
+        };
+
```
gjn: I would not introduce this promise. I would instead create a local promise in backgrounddirective, that will be resolved on the first gaLayersChange event.

Jenselme:
Would something like this suit you better?

src/components/map/MapService.js
oterral: Maybe this promise could be useful in gaLayersPermalinkManger too. Using $q.all that simplify a lot what happens on first load.

gjn:
Yes, this looks good. I'll review and test in detail. Thanks for your patience.

src/components/map/MapService.js
Jenselme: It seems so. @gjn what is your opinion on this?

gjn:
btw: I think @oterral is probably correct that the promise in gaLayers might serve other places and might in fact facilitate the initial loading of the application. So I'm not purely against it.

But then we should adress this globally, identify places for application init is critical and adress all those places at once.

gjn:
I think we could say that a) promises should take care of application loading and b) events should take care of changes, trigger by UI interaction

Border case: permalink parameters, which are relevant at application loading only, so should be solved with promises as well.

@oterral Do you think it's doable in a reasonable way? What are the events and components that would be affected?

Jenselme:
> then we should adress this globally, identify places for application init is critical and adress all those places at once.

Maybe the current patch is enough for this PR and you should open a new one to address more globally. I let you decide which solution to choose for this PR. Let me know.

gjn:
I would keep it as it is for this PR.

src/components/backgroundselector/BackgroundSelectorDirective.js
```diff
@@ -91,14 +94,26 @@ goog.require('ga_topic_service');
           };
 
           var dereg = scope.$on('gaLayersChange', function(event, newLayers) {
+            getLayersDeffered.resolve();
             updateBgLayer();
             dereg();
```
gjn: We dereg here, so this function is called only once. Is this wanted? if yes, why is updateGbLayer() still called here?

I'm a little confused about what this directive wants to achieve.

oterral:
In this PR what happens on TopicChange , bgLayer is changed?

oterral:
@gjn think with $q.all trick we could resolve our problems of inital load.  Problems comes from component which needs Layers and Topics at the same time, I think about gaLayersPermalinkManager and BackgroundSelector. I don't think there is other components which needs the 2.

oterral:
I forget Search and Seo directives as well.

gjn:
@Jenselme just discussed with @oterral . We will create a separate PR adressing application init and using your idea of a gaLayers promise. Once this is merged, you can then adapt this PR to have what you need.

Is this ok for you?

src/components/backgroundselector/BackgroundSelectorDirective.js
Jenselme: This code blocks come from your (not you personally) current implementation. I think that the goal is to override the default bg layer from the topic with the one the user selected (which only happens once, hence the dereg). However, with my current implementation, it is always the default bg layer which is chosen since I can only be sure that the layers are loaded once `gaLayersChange` is emitted.

I will wait on your PR addressing init to fix this correctly and clear the code of this PR.

oterral:
@Jenselme and @oterral the PR is there #2539 .

While I'm doing refactor this thing. It appeared we are stil dependent of the background selector directive (I mean we can't have a bg layer if the bg selector directive is not in th HTML page).

I think we should have a gaBackgroundLayer service which manage the permalink and the list of bg layers available. This service will trigger gaBackgroundChange. A bit like the Topic service. What do you think @gjn @Jenselme ?

gjn:
Probably yes. Could be done after #2539.

Jenselme:
@oterral it seems to be a good idea.

Jenselme:
See #2561 for rework. It was easy for me to restart from scratch and adapt the patches than to do a rebase.

PR #2443, opened 2015-06-20
Dynamic BG Selector

cedricmoullet:


src/components/map/MapService.js
```diff
@@ -757,6 +757,7 @@
 
           var promise = $http.get(url).then(function(response) {
             layers = response.data;
+            $rootScope.$broadcast('gaLayersLoaded');
```
oterral: What is the difference between `gaLayersLoaded` and the existing `gaLayersChange` event ?

PR #2455, opened 2015-06-30
Add missing translation

alcapat:
One translation is missing for the tooltip (topic gewis)


procrastinatio:
OK, doesn't break anything

Issue #2450, opened 2015-06-30
Actuality date not shown in iFrame


cedricmoullet:
![image](https://cloud.githubusercontent.com/assets/319676/8424174/9d6c8b28-1ef6-11e5-8423-fc971bb6ac7f.png)

The actualization date should be shown within the iFrame: https://map.geo.admin.ch/?X=190000.00&Y=660000.00&zoom=1&lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.bafu.hydroweb-messstationen_zustand

davidoesch:
might be related as well to https://github.com/geoadmin/mf-geoadmin3/issues/2369

oterral:
Fixed on prod

PR #2462, opened 2015-07-01
Timestamp control should also be part of the embed template

loicgasser:
https://github.com/geoadmin/mf-geoadmin3/issues/2450


gjn:
Thank you for this.

PR #2466, opened 2015-07-03
Print azimuth

oterral:
Missing commit of new draw branch. Thx `git reflog`


oterral:
Tested with all knid of vector layers and overlay marker

PR #2472, opened 2015-07-07
Fix translate - usability improvements

cedricmoullet:


cedricmoullet:
Self merge - trivial trranslations

PR #2475, opened 2015-07-07
Fix kml tests

gjn:
This fixes the kml tests. We use the wait condition now _before_ getting the URL.


gjn:
jenkins test

gjn:
review with @alcapat

Issue #2410, opened 2015-06-10
Throttle service test fails


gjn:
Throttle service test fails from time to time. This happened today twice during deploy.

https://github.com/geoadmin/mf-geoadmin3/blob/master/test/specs/ThrottleService.spec.js

We should try to make it more robust. Assigend to @kandre as he's the original author.

kandre:
Have you the log of the error?

oterral:
In fact the problem is if the server is a little bit busy that slowdown the execution of the Throttle function.

gjn:
I think anyway, that the tests could be made more robust (introduce a more fuzzy like test). I can do a proposal if @kandre doesn't want to do it :)

oterral:
I've made a correction https://github.com/geoadmin/mf-geoadmin3/commit/ec5f044a541e2f8105a47f18acddfd50c5a5f172

Hope that fixes the problem

PR #2476, opened 2015-07-07
Improve throttle unit test

gjn:
https://github.com/geoadmin/mf-geoadmin3/issues/2410

This should be a little more stable.


gjn:
@oterral That do you think?

oterral:
Yes this test is better, the last condition should fix the problem we have

PR #2483, opened 2015-07-07
Fix #2417: Update the IE detection of typeahead

### LINKED ISSUE #2417: WMS import on IE11
1. Start WMS import
2. Select a WMS server

The selection list doesn't dissapear after the selection of a WMS server.

oterral:
Fix #2417 updating the IE detection.

Since we want to remove typeahead . I don't think we must loose time to try to update the typeahead version.

[Test](https://mf-geoadmin3.dev.bgdi.ch/teo_fix_2417/?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&X=190000.00&Y=660000.00&zoom=1)


cedricmoullet:
Thanks, works fine.

PR #2494, opened 2015-07-14
Use publicUrl to match files from our storage service

Jenselme:
I directly use publicUrl so it can only match one domain. I don't know if you have the case where files can come from both public.admin.ch and public.bgdi.ch. If so, let me know, I will update the pull request.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,8 @@ LAST_APACHE_BASE_PATH := $(shell if [ -f .build-artefacts/last-apache-base-path 
 API_URL ?= //mf-chsdi3.dev.bgdi.ch
 LAST_API_URL := $(shell if [ -f .build-artefacts/last-api-url ]; then cat .build-artefacts/last-api-url 2> /dev/null; else echo '-none-'; fi)
 PUBLIC_URL ?= //public.dev.bgdi.ch
+PUBLIC_URL_REGEXP ?= https?:\/\/public\..*(bgdi|admin)\.ch.*
+ADMIN_URL_REGEXP ?= ^(ftp|http|https):\/\/(.*(\.bgdi|\.geo\.admin)\.ch)
 LESS_PARAMETERS ?= '-ru'
 KEEP_VERSION ?= 'false'
 LAST_VERSION := $(shell if [ -f .build-artefacts/last-version ]; then cat .build-artefacts/last-version 2> /dev/null; else echo '-none-'; fi)
@@ -280,6 +282,8 @@ define buildpage
 		--var "resolutions"="$(RESOLUTIONS)" \
 		--var "public_url=$(PUBLIC_URL)" \
 		--var "default_elevation_model=${DEFAULT_ELEVATION_MODEL}" \
+		--var "admin_url_regexp=$(ADMIN_URL_REGEXP)" \
+		--var "public_url_regexp=$(PUBLIC_URL_REGEXP)" \
 		--var "default_epsg"="$(DEFAULT_EPSG)" \
 		--var "default_epsg_extend"="$(DEFAULT_EPSG_EXTEND)" $< > $@
 endef
--- a/src/components/map/MapService.js
+++ b/src/components/map/MapService.js
@@ -1239,6 +1239,13 @@ goog.require('ga_urlutils_service');
           if (id instanceof ol.layer.Layer) {
             id = olLayerOrId.id;
           }
+
+          // id for KML layers is like KML||<url> We must remove KML|| before
+          // testing the validity of the url.
+          if (id && id.indexOf('KML||') == 0) {
+            id = id.substring(5);
+          }
+
           return this.isKmlLayer(olLayerOrId) &&
                   gaUrlUtils.isPublicValid(id);
         },
--- a/src/components/map/MapService.js
+++ b/src/components/map/MapService.js
@@ -1239,8 +1239,8 @@ goog.require('ga_urlutils_service');
           if (id instanceof ol.layer.Layer) {
             id = olLayerOrId.id;
           }
-          var regex = /https?:\/\/public\..*(\.admin\.ch|\.bgdi\.ch)\/.*/;
-          return this.isKmlLayer(olLayerOrId) && regex.test(id);
+          return this.isKmlLayer(olLayerOrId) &&
+                  gaUrlUtils.isPublicValid(id);
         },
 
         // Test if a layer is an external WMS layer added by th ImportWMS tool
--- a/src/index.mako.html
+++ b/src/index.mako.html
@@ -686,6 +686,8 @@ itemscope itemtype="http://schema.org/WebApplication"
           mapUrl : location.origin + '${apache_base_path}',
           apiUrl : location.protocol + '${api_url}',
           publicUrl : location.protocol + '${public_url}',
+          publicUrlRegexp: new RegExp('${public_url_regexp}'),
+          adminUrlRegexp: new RegExp('${admin_url_regexp}'),
           cachedApiUrl: location.protocol + '${api_url}' + cacheAdd,
           resourceUrl: location.origin + pathname + '${versionslashed}',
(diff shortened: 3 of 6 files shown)


oterral:
yes we have this use case where files can come from both public.admin.ch and public.bgdi.ch

Jenselme:
I see two possibility to cover this case:
1. Add an `publicAllowedUrlRegexp` option in configuration. In your case it would be `https?:\/\/public\..*(\.admin\.ch|\.bgdi\.ch)\/.*`. This is the closest to the current code.
2. Add a `publicAllowedUrls` option in configuration which would contain an array of allowed urls, in your case `['public.admin.ch', 'public.bgdi.ch']`.

I don't know which one is the best. @oterral Do you have an opinion?

oterral:
@jenselme a regex is more powerful :)

Jenselme:
A regexp it is then. PR updated.

Jenselme:
I just updated this PR to take into account the new hard coded URLs from  src/components/UrlUtilsService.js

Makefile
```diff
@@ -9,6 +9,9 @@ LAST_APACHE_BASE_PATH := $(shell if [ -f .build-artefacts/last-apache-base-path
 API_URL ?= //mf-chsdi3.dev.bgdi.ch
 LAST_API_URL := $(shell if [ -f .build-artefacts/last-api-url ]; then cat .build-artefacts/last-api-url 2> /dev/null; else echo '-none-'; fi)
 PUBLIC_URL ?= //public.dev.bgdi.ch
+PUBLIC_ALLOWED_URL_REGEXP ?= https?:\/\/public\..*(\.admin\.ch|\.bgdi\.ch)\/.*
+PUBLIC_URL_REGEXP ?= ^https?:\/\/public\..*(bgdi|geo\.admin)\.ch.*
```
oterral: `PUBLIC_ALLOWED_URL_REGEXP`  and `PUBLIC_URL_REGEXP` do the same thing . Can you remove `PUBLIC_ALLOWED_URL_REGEXP` pls ?

src/components/map/MapService.js
```diff
@@ -1263,8 +1263,8 @@ goog.require('ga_urlutils_service');
           if (id instanceof ol.layer.Layer) {
             id = olLayerOrId.id;
           }
-          var regex = /https?:\/\/public\..*(\.admin\.ch|\.bgdi\.ch)\/.*/;
-          return this.isKmlLayer(olLayerOrId) && regex.test(id);
+          return this.isKmlLayer(olLayerOrId) &&
+                  gaGlobalOptions.publicAllowedUrlRegexp.test(id);
```
oterral: Can you add a function `isPublicValid` in `gaUrlUtils` pls then use it here.

src/index.mako.html
```diff
@@ -686,6 +686,8 @@
           mapUrl : location.origin + '${apache_base_path}',
           apiUrl : location.protocol + '${api_url}',
           publicUrl : location.protocol + '${public_url}',
+          publicAllowedUrlRegexp: new RegExp('${public_allowed_url_regexp}'),
```
oterral: To remove

test/specs/Loader.spec.js
```diff
@@ -33,8 +33,10 @@ beforeEach(function() {
       ],
       defaultTopicId: 'sometopic',
       translationFallbackCode: 'somelang',
-      defaultExtent: [420000, 30000,900000, 350000],
```
oterral: Why removing `defaultExtent`?

test/specs/Loader.spec.js
```diff
@@ -33,8 +33,10 @@ beforeEach(function() {
...
-      defaultResolution: 500.0
+      defaultResolution: 500.0,
+      publicAllowedUrlRegexp: /https?:\/\/public\..*(\.admin\.ch|\.bgdi\.ch)\/.*/,
```
oterral: to remove

Makefile
Jenselme: Maybe that both regexp should do the same thing but they currently don't. The first will match `https://public.bgdi.ch` but not the latter. Maybe you can give me the regexp that matches all your cases so I can easily use your test suite to test this PR.

test/specs/Loader.spec.js
Jenselme: It's a mistake, I had `defaultExtent` twice in my branch. I will cancel the deletion.

Jenselme:
@oterral  I will wait for #2344 before rebasing and taking into account your suggestions.

Jenselme:
While removing `PUBLIC_ALLOWED_URL_REGEXP ?= https?:\/\/public\..*(\.admin\.ch|\.bgdi\.ch)\/.*` I compared `PUBLIC_URL_REGEXP ?= ^https?:\/\/public\..*(bgdi|geo\.admin)\.ch.*` and `ADMIN_URL_REGEXP ?= ^(ftp|http|https):\/\/(.*(\.bgdi|\.geo\.admin)\.ch)` and they look pretty similar to me.

Can/should they be merged in one regexp?

Jenselme:
@oterral PR updated with your remarks. `ga_map_service gaMapUtils tests isStoredKmlLayer` fail because the two regexps are not exactly the same. Since you know better than me the URLs that should match, can you write it?

oterral:
To summarize, we need to detect if a external file comes from our domain (*.admin.ch and *.bgdi.ch) and if it comes from a third party domain ( all hostnames + public.*.admin.ch + public.*.bgdi.ch)

Jenselme:
In `public..admin.ch` and `public..bgdi.ch` is the double dot `..` intentional or is a star missing?

oterral:
A star is missing :)

oterral:
The test failed [here](https://github.com/geoadmin/mf-geoadmin3/pull/2494/files#diff-66ac86f99915855fde57c0a20d8f6d6bL1267) because id is like `KML||https://public.geo.admin.ch` and not `https://public.geo.admin.ch`

oterral:
I will fix it in another PR

Jenselme:
Just updated the regexp and fix `ga_map_service gaMapUtils tests isStoredKmlLayer`

src/index.mako.html
```diff
@@ -686,6 +686,7 @@
           mapUrl : location.origin + '${apache_base_path}',
           apiUrl : location.protocol + '${api_url}',
           publicUrl : location.protocol + '${public_url}',
+          adminUrlRegexp: new RegExp('${admin_url_regexp}'),
```
oterral: publicUrlRegExp is missing

Jenselme: Corrected.

cedricmoullet:
@Jenselme could you please resolve the conflicts ?

Jenselme:
Conflicts resolved!

src/index.mako.html
oterral: Forget to push ?

Jenselme: So it seems. Just pushed the modifications.

oterral:
@Jenselme it seems you p have pushed an old version of the PR because the isPublicValid function has disappeared.

Jenselme:
This time it should be good.

test/specs/Loader.spec.js
oterral: Can you re-add defaultExtent pls

oterral:
@Jenselme we are close to the end still 2 things:

 - [ ] another rebase because of https://github.com/geoadmin/mf-geoadmin3/pull/2541
- [ ] add defaultExtent in loader.spec.js

Jenselme:
>  add defaultExtent in loader.spec.js 

If I look ad loader.spec.js, line 24 I see `defaultExtent`. So I guess that this key was present two times before. Should I readd it anyway?

PR rebased.

oterral:
Ah yes you're right so it's good .

oterral:
Thx for your patience.

PR #2495, opened 2015-07-14
[WIP] Save drawings on a webdav server

Jenselme:
I am working on a way to save/load drawings on/from a webdav server (eg owncloud). It is saved/loaded from a KML file. The work is not yet completely done. I am still opening this PR to know if you are interested in this functionality and if you are so we can discuss how to complete the development. This work is based on the current drawing tools.

# What works and how it works

- User can choose not to save the drawing, to save the drawing on the default server or to save the drawing on a webdav server. By default, drawing are saved on server (eg S3)

draw-no-save

- User can load a drawing from a webdav instance. Currently, the drawing can only be loaded from the *draw and measure on map* screen.

draw-webdav-save

- Drawing from a webdav instance is modified if the user modify the drawing through the interface.
- File is deleted from webdav server if the user choose to delete all features of the drawing
- Error messages are displayed if the file cannot be found or if the requests fail (CORS problem, bad authentication, …)
- Share links are only displayed if the user saves the file on the default server.

draw-server-save-links

# What remains to be done

1. Translations: 
   - Text are currently only available in English
   - The error messages are not yet handled to be translated
2. Allow the user to load automatically (ask for username and password if necessary) a KML imported this way
3. Allow the user to import a KML from a webdav server with the *KML import* tool

# Points that require input

1. General layout of the functionality. I tried to place the elements logically but you have more experience in this domain than I have. Moreover, I currently display webdav related messages under the webdav connection block. I don't know if it would be better to place them all with the "saving" messages.
2. Should the drawing be loaded from webdav server only from the *KML import* tool? I think not: in my opinion it makes sense to be able to load KML from both draw tools and *KML import*:
   - I want to draw
   - If I forgot to load it before, I can load it here
   - Can enter my credentials only once
3. Should the filename/filepath be entered in another input field? Currently, the user must provide the full URL of the file in the URL field and not just the webdav endpoint. I think it can be disturbing for people who are familiar with webdav. On the other hand, that will make quite a lot of fields.
4. Should I add a button *Create* to create an empty file. It is not necessary as the file will be created if it doesn't exist but it may be more obvious and clear than an explanation that the file will be created automatically.
5. Should we add a way to browse files present on webdav server? Only KMLs? Problem: certainly long to do correctly (unless there is already an angular plugin for that).

# Not tested

- How it works on mobile/small screen
- Modify the KML from the geoadmin interface and somewhere else. I guess one file should erase the other. Should we add conflict resolution.

Let me know what you think.


gjn:
I'll close this as discussed off-line.

PR #2496, opened 2015-07-15
Dev browserstack improvement

alcapat:
Like discuss : 
- Only start_test for all browser (3 lasts versions of each browsers) and full test for the browser Chrome 43 Win 7.
- Getcapabilities of browser updated with last version available in browserstack.

### DIFF
--- a/test/selenium/tests.js
+++ b/test/selenium/tests.js
@@ -33,9 +33,16 @@ var printTest = require('./print_test.js');
 //okay we will start the script!
 console.log("Starting Browserstack script!");
 
-//for every browser config we want to run all the tests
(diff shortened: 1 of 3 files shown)


test/selenium/tests.js
```diff
@@ -33,9 +33,16 @@ var printTest = require('./print_test.js');
 //okay we will start the script!
 console.log("Starting Browserstack script!");
 
-//for every browser config we want to run all the tests
-browsers.capabilities.forEach(function(cap){
+// Start full test only for one browser (use Chrome for the print)
+var RunFullTests = function(cap) {
```
gjn: Please use lowercase for first character of function name.

PR #2499, opened 2015-07-17
Limit layers extent to the extent of the portal

Jenselme:
It prevents to load tile that are outside the scope of the portal. It
saves bandwidth, reduces load time, avoids to display unuseful
information.

### DIFF
--- a/src/components/map/MapService.js
+++ b/src/components/map/MapService.js
@@ -356,6 +356,7 @@ goog.require('ga_urlutils_service');
             opacity: options.opacity,
             visible: options.visible,
             attribution: options.attribution,
+            extent: options.extent,
             source: source
           });
           gaDefinePropertiesForLayer(layer);
@@ -375,7 +376,7 @@ goog.require('ga_urlutils_service');
           var wmsOptions = {
             url: getCapLayer.wmsUrl,
             label: getCapLayer.Title,
-            extent: getCapLayer.extent,
+            extent: gaMapUtils.mergeWithDefaultExtent(getCapLayer.extent),
             attribution: getCapLayer.wmsUrl
           };
           return createWmsLayer(wmsParams, wmsOptions);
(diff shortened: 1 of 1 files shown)


src/components/map/MapService.js
```diff
@@ -1308,6 +1313,18 @@ goog.require('ga_urlutils_service');
             easing: ol.easing.easeOut
           }));
           map.getView().setRotation(0);
+        },
+
+        adaptExtentPortalExtent: function(extent) {
```
gjn: Maybe adapt function name to reflect better what it does? Like `mergeWithDefaultExtent`.

gjn:
This is nice. Please rebase and probably adapt a function name.

Jenselme:
- rebased
- rename `adaptExtentPortalExtent` into `mergeWithDefaultExtent`

gjn:
Thanks alot for this!

src/components/map/MapService.js
```diff
@@ -375,7 +376,7 @@ goog.require('ga_urlutils_service');
           var wmsOptions = {
             url: getCapLayer.wmsUrl,
             label: getCapLayer.Title,
-            extent: getCapLayer.extent,
+            extent: gaMapUtils.mergeWithDefaultExtent(getCapLayer.extent),
```
oterral: Hmm I'm not sure of this one, if the layer's extent define by the GetCap is 10km aroud Zurich for example. I don't think we want the layer was requested outside this extent

oterral: Forget this I hadn't  read what the function does

gjn: AFAIK, the mergeWith is actually an intersectWith, @Jenselme ?

Jenselme: From what I understand of your comment, this line solves this.

Jenselme: > AFAIK, the mergeWith is actually an intersectWith, @Jenselme ?

Indeed, this name looks better. Should I open a new PR?

gjn: I would merge it in a second :=)

PR #2501, opened 2015-07-21
Add selected layers on topic change

oterral:
With tests 

@gjn can you review/merge for tomorrow pls?


gjn:
Test failure is probably unrelated, but we need a fix for this as well. It's because of delayed loading of print configuration which makes the print test fail now. (not always, but most of the time).

gjn:
Good one. Thanks alot.

PR #2506, opened 2015-07-21
Print: enable button when config is loaded. Adapt tests

gjn:
This assures that you can print only after configuration has been loaded. In addition, it adapts the tests accordingly, which should render them more stable.

@oterral Prefarably for tomorrow.

### DIFF
--- a/test/selenium/print_test.js
+++ b/test/selenium/print_test.js
@@ -13,6 +13,8 @@ var runTest = function(cap, driver, target) {
   driver.findElement(webdriver.By.xpath("//a[@id='printHeading']")).click();
   // Wait until print is opened and animation is finished
   driver.findElement(webdriver.By.xpath("//div[@id='print' and contains(@class, 'collapse in')]"));
+  // Wait until configuration is loaded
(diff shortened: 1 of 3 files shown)


test/selenium/print_test.js
```diff
@@ -11,6 +11,8 @@ var runTest = function(cap, driver, target) {
   driver.findElement(webdriver.By.xpath("//a[contains(text(), 'Grundlagen und Planung')]"));
   // Click on "Drucken"
   driver.findElement(webdriver.By.xpath("//a[@id='printHeading']")).click();
+  // Wait until configuration is loaded
+  driver.findElement(webdriver.By.xpath("//span[@ng-if='printConfigLoaded']"));
```
oterral: Couldn't we juste test if there is options in  one of the selectbox? 

I'm not a fan of added scope property only for test

gjn: Yes, you are correct.

I'll try.

gjn:
@oterral I've adapted.

oterral:
lgtm thx

PR #2508, opened 2015-07-21
Fix delete feature action when the user is writing in an input

oterral:
This PR disallows the removal of selected feature with delete key when the user is writing.

It also add a confirm message when the user remove a feature.

@gjn for tomorrow pls it's a nasty bug

@daguer thx for finding this


gjn:
Works find. Good find.

PR #2509, opened 2015-07-22
Fix css for icons

pfanguin:


pfanguin:
Test link : https://mf-geoadmin3.dev.bgdi.ch/dev_css_icons

oterral:
Can't you do it these modifications in the font awesome files ?

cedricmoullet:
@pfanguin Could you please answer oli's question ?

pfanguin:
The best solution that I have found

oterral:
I think, adding a font-size different for each icon is really bad,  we loose the purpose of using font-awesome.  We already do that in special case and it's not good.

@pfanguin Can you really not modify  the SVG to avoid using the `font-size` rules?

pfanguin:
Better for you ?

pfanguin:
jenkins test

pfanguin:
jenkins test

src/style/app.less
```diff
@@ -749,8 +756,11 @@ a .ga-icon-separator {
 
   .icon-reorder, .icon-remove {
     display: none;
+    vertical-align: middle;
```
oterral: icon-remove don't need the vertical-align

oterral:
There is something that still bother me. A font-aweosme icon must be aligned with a text by default (default css is vertical-align: baseline) here it's not the case. 

Are you able to provide an HTML test page with all the icons we have when you build the font?. Like this one http://fortawesome.github.io/Font-Awesome/3.2.1/cheatsheet/

With this kind of page we could easily see if icons are well-aligned and well-sized easily.

oterral:
jenkins test

PR #2516, opened 2015-07-23
Cancel previous catalog request if antother is launched

### LINKED ISSUE #2512: Catalog items are not translated when fall back to default language
Only Chrome (IE and FF OK)

1/ Link https://map.geo.admin.ch/?topic=gewiss&lang=it
2/ As 'lang=it' doesnt exist for 'topic=gewiss', switch to German
3/ First level in catalog is missing (if you go directly with 'lang=de' it's OK)

gewiss

Expected: First level in tree should be translated in German.

oterral:
DE catalog was not loaded when we opens Gewiss in IT, this PR cancel previous catalog requests

Fix #2512 

@procrastinatio does it seems ok for you?

https://mf-geoadmin3.dev.bgdi.ch/teo_catalog/?topic=gewiss&lang=it

### DIFF
--- a/src/components/catalogtree/CatalogtreeDirective.js
+++ b/src/components/catalogtree/CatalogtreeDirective.js
@@ -103,6 +103,7 @@ goog.require('ga_translation_service');
               }
             };
             var lastUrlUsed;
+            var canceller;
             var updateCatalogTree = function(topic, lang) {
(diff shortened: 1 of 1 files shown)


gjn:
Looks good to me. Thanks alot.

PR #2517, opened 2015-07-23
Don't erase local topics variable

Jenselme:


oterral:
yes better thx @Jenselme

PR #2520, opened 2015-07-24
[WIP] Prevent broadcasting while some directives or controller may not be initialized

Jenselme:
We must wait for all the directive to be completely initialized before broadcasting. Otherwise, we may broadcast while nothing is initialized to receive the message.

The $broadcast which is done in TopicService may happen before the directives are initialized. We add a $timeout to wait for complete DOM initialization to be sure all components that must listen to $broadcast('gaTopicChange') are ready.

### DIFF
--- a/src/components/topic/TopicService.js
+++ b/src/components/topic/TopicService.js
@@ -11,7 +11,8 @@ goog.require('ga_permalink');
    * Topics manager
    */
   module.provider('gaTopic', function() {
-    this.$get = function($rootScope, $http, $q, gaPermalink, gaGlobalOptions) {
+    this.$get = function($rootScope, $http, $q, $timeout, gaPermalink,
+        gaGlobalOptions) {
       var topic; // The current topic
       var topics = []; // The list of topics available
 
@@ -66,7 +67,10 @@ goog.require('ga_permalink');
           topics = topics;
           topic = getTopicById(gaPermalink.getParams().topic, true);
           if (topic) {
-            broadcast();
+            // We must wait for all the directive to be completely initialized
+            // before broadcasting. Otherwise, we may broadcast while nothing
+            // is initialized to recieve the message.
+            $timeout(broadcast);
           }
         });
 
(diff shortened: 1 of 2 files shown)


gjn:
What is the observed problem?

Not a fan of timeout, as it might be good for some, not good for others (might be browser specific, connection specific at given moment, etc). And increasing it slows down the app for all to help the few.

Are there no other ways to address this? 

Maybe we could make the depending directives/services/controllers get a promise from the topic service and act on that on initialisation, in other words, separate initialisation from standard topic change messages.

Jenselme:
The problem we saw was: the list of topics was not loaded.

I did some digging and found out that TopicService can get its list of topics and broadcast the `gaTopicChange` event before the TopicDirective is ready (which is logical since services are initialized before directives). The code about `$on('gaTopicChange')` was not yet executed. So the directive didn't work as expected (didn't have its initial list of topics).

I guess that you always get your list of topics from the database and that this add enough latency for the directive to register its `$on('gaTopicChange')` before the promise which gets the topics resolves.

> Maybe we could make the depending directives/services/controllers get a promise from the topic service and act on that on initialisation, in other words, separate initialisation from standard topic  change messages.

I guess that could do trick. I choose the `$timeout` as I spent lot of time to correctly figure out what was going on and how to fix it. Moreover, it required not that much work. In order for the promise solution to work, it would require more work and changes. Do you have an idea on how to do it right? And also should I or should you do these changes?

gjn:
We (me or @oterral ) could give it a try. It's indeed not trivial and requires some changes, but it adresses a flaw in our design. (Usage of timeouts is very often a code smell for bad design). I think it's worth it.

gjn:
I think for the case you are describing, it's enought to let `gaTopic.getTopics()` return a promise, then use that promise to initialize the topic directive. This should be possible without much refactoring.

I'll give it a go.

gjn:
@Jenselme I've created an alternative version. https://github.com/geoadmin/mf-geoadmin3/pull/2521 Could you try if it works for you?

Jenselme:
Solved by #2521

PR #2521, opened 2015-07-24
Different handling of topic loading at application start

gjn:
This is an alternative to https://github.com/geoadmin/mf-geoadmin3/pull/2520

Before, we handled 2 concerns (application init when topics are loaded and topic changes) in one broddcast. This could lead to timing issues as described by @Jenselme in his PR.

This PR tries to separate those 2 concerns. It distinguishes application init and topic changes.

### DIFF
--- a/src/components/topic/TopicDirective.js
+++ b/src/components/topic/TopicDirective.js
@@ -43,15 +43,17 @@ goog.require('ga_topic_service');
               }
             });
 
-            scope.$on('gaTopicChange', function(evt, newTopic) {
-              if (!scope.activeTopic) {
-                scope.topics = gaTopic.getTopics();
-                scope.$applyAsync(function() {
-                  element.find('.ga-topic-item').tooltip({
-                    placement: 'bottom'
-                  });
+            gaTopic.getTopics().then(function(topics) {
+              scope.topics = topics;
+              scope.activeTopic = gaTopic.get();
+              scope.$applyAsync(function() {
+                element.find('.ga-topic-item').tooltip({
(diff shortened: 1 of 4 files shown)


Jenselme:
Just tested, it works fine. Please merge.

gjn:
@oterral could you review?

src/components/topic/TopicDirective.js
```diff
@@ -43,15 +43,16 @@ goog.require('ga_topic_service');
...
 
-            scope.$on('gaTopicChange', function(evt, newTopic) {
-              if (!scope.activeTopic) {
-                scope.topics = gaTopic.getTopics();
-                scope.$applyAsync(function() {
-                  element.find('.ga-topic-item').tooltip({
-                    placement: 'bottom'
-                  });
+            gaTopic.getTopics().then(function(topics) {
+              scope.topics = topics;
```
oterral: I think you also need to set the `activeTopic` variable otherwise the topic will not be selected in the list, if the `gaTopicChange ` event is triggered before the loading of the directive. I've reproduced once.

oterral:
+1 I like the promise way. Just a comment of setting the `activeTopic` variable otherwse lgtm

oterral:
Maybe,  to be sure how the things work in this directive we should register the `gaTopicChange` event. in the success callback of the promise.

oterral:
I've added the activeTopi stuff.

gjn:
Thanks alot.

Jenselme:
Thanks, I will update our code soon with this fix!

Issue #2511, opened 2015-07-22
Hide search results when switching to draw view


Jenselme:
If you type a search (eg mout), don't select any result and then activate the "draw and measure" tool, you will still see your search results even if the search box is hidden (see screenshot).

![search-draw](https://cloud.githubusercontent.com/assets/6643536/8825507/012b42e0-3080-11e5-98b1-555fab540365.png)

I think it would be better to hide those search results.

PR #2528, opened 2015-07-28
Close drop down on blur

loicgasser:
This PR is introduced mainly to fix https://github.com/geoadmin/mf-geoadmin3/issues/25111

Note that I had to use a timeout for the click on the result items as the blur event on the input is fired before the click. (If you have other suggestions..)

Behavior is the same as before for swissearch and keyboard nav, the only change is that the dropdown list now closes when the input loses focus. In other words, we don't need to clear the query or select an entry for the dropdown list to disappear.

### DIFF
--- a/src/components/search/SearchDirective.js
+++ b/src/components/search/SearchDirective.js
@@ -67,6 +67,7 @@ goog.require('ga_search_type_directives');
     function($scope, $rootScope, $translate, $sce, $timeout, gaPermalink,
              gaUrlUtils, gaSearchGetCoordinate, gaMapUtils, gaMarkerOverlay,
              gaKml, gaPreviewLayers) {
+      var keyboardNav = false;
       var blockQuery = false;
       var restat = new ResultStats();
       $scope.restat = restat;
@@ -117,14 +118,6 @@ goog.require('ga_search_type_directives');
         evt.preventDefault();
       };
 
-      $scope.lostFocus = function(evt) {
-        $scope.searchFocused = false;
-      };
-
-      $scope.onFocus = function(evt) {
-        $scope.searchFocused = true;
-      };
-
       var startQuery = function(q) {
         restat.reset();
 
@@ -165,7 +158,7 @@ goog.require('ga_search_type_directives');
         startQuery($scope.query);
       };
 
-      $scope.$watch('query', function(newval, oldval) {
+      $scope.$watch('query', function() {
         $scope.trigger();
       });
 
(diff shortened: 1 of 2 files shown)


src/components/search/SearchDirective.js
```diff
@@ -183,10 +176,42 @@ goog.require('ga_search_type_directives');
...
+      };
+
+      $scope.onFocus = function() {
+        losingFocus = false;
+        $scope.setKeyboardNav(false);
+        $scope.searchFocused = true;
+      };
+
+      $scope.showDropDown = function() {
+        var hasResults = restat.sets() > 0 ? true : false;
```
gjn: Isn't `restat.sets()` enough? Should be truthy when there are results and falsy when not.

gjn:
Can't we fix #2511 with css only? @oterral ?

oterral:
I was looking into it. I think the best is to listen for onclick event on the doucment then hide the list of results if the click is outside the search list or input.

gjn:
But we don't want usually. We only want that when drawing is opened.

Maybe having events broadcasted when start/end drawing? Not ideal...

gjn:
As it is in this PR, the search gets closed when zooming or panning the map. I don't think that's a good idea.

oterral:
Or when you click on topic selector

oterral:
If it's only when draw is active yes we can do by css

loicgasser:
Yes it's kind of tricky. I wanted the dropdown to close when it looses focus, which is probably a "normal" behavior. It works fine if I replace the click event with the a mousedown (as it is triggered before the blur event on the input), but then things get messy on mobile.
Then CSS solution could do the trick...
@gjn if you don't click the map the search dropdown doesn t disappear.

loicgasser:
So you guys are happy with the fact that the input doesn't care about being focused. Cause we have a searchFocused on the scope but it is used nowhere.

gjn:
Well, Search results disappear here when I click on the map (or anywhere else that is not search).

Anyway, I think we should play with css. A simple `display: none` on the div with `id=search-container` hides everything search related. This could even be done on MainController level, so we don't pollute the component with this.

loicgasser:
ok, we're happy with the current behavior then. I'll close.

loicgasser:
FYI: http://www.local.ch/ same behavior 
Search results disappear here when you click anywhere else that is not search

gjn:
I did a PR to fix #2511 using just css: https://github.com/geoadmin/mf-geoadmin3/pull/2529

gjn:
`So you guys are happy with the fact that the input doesn't care about being focused. Cause we have a searchFocused on the scope but it is used nowhere.`

Well, this could be corrected if you think it makes sense.

I believe keeping the search results open and being able to navigate the map is a big +. Especially because we have feature and result highlighting.

loicgasser:
ok will clean up then

loicgasser:
What about the CSS of the result items? It changed from grey boxes with a white outiline, to white. Is this also wanted?

gjn:
But the searchFocused is used. It's in the index.thml (https://github.com/geoadmin/mf-geoadmin3/blob/master/src/index.mako.html#L111) And a css rule is attached to it: https://github.com/geoadmin/mf-geoadmin3/blob/master/src/style/app.less#L408

gjn:
`What about the CSS of the result items?`

I don't see any CSS changes in this PR...

gjn:
Forgot to mentiong. searchFocused is passed to directive here: https://github.com/geoadmin/mf-geoadmin3/blob/master/src/index.mako.html#L154

loicgasser:
not in this PR in general, didn't change cause I didn't know if it was wanted.

loicgasser:
yes, I saw that, but I fail to understand the point of it. don't think it's needed. Agree?

gjn:
It's needed. The global variable is bound to the local scope variable of the directive.

https://github.com/geoadmin/mf-geoadmin3/blob/master/src/index.mako.html#L154 results in the $scope.searchFocused variable to be set to the global variable via https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/search/SearchDirective.js#L225

That's why when you change $scope.searchFocused, the global searchFocused is changed as well (as it's acutally the same variable). Angular magic.

loicgasser:
ok got that but what does the local variable?

loicgasser:
Or the global variable? They don't seem to influence anything.

gjn:
The global variable is used in an ng-class directive: https://github.com/geoadmin/mf-geoadmin3/blob/master/src/index.mako.html#L111 The class is then used in app.less (css).

The local variable (and thus the global as well) is changed in the lostFocus and onFocus functions of the directive. Those functions are called by angular on the ng-blur and ng-focus events of the partial here: https://github.com/geoadmin/mf-geoadmin3/blob/0f4b3fbaaf27cb5ad3998f716ba36f568b07ffb9/src/components/search/partials/search.html#L8

loicgasser:
ok got it, which is then used on mobile to change the search width / hide menu btn. Thank you for your explanations.

gjn:
No worries. Levels of indirection in angular are sometimes overwhelming...

PR #2529, opened 2015-07-28
Hide search when draw is active

gjn:
This is an alternative fix for #2511.

### DIFF
--- a/src/style/app.less
+++ b/src/style/app.less
@@ -218,7 +218,7 @@ button.ga-btn, .ol-control button {
     top: unit(@small-header-height, px);
   }
 
-  #pulldown-content .panel-heading, .theme-toggle {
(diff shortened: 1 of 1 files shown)


oterral:
easier :)

src/style/app.less
```diff
@@ -205,6 +205,10 @@ button.ga-btn, .ol-control button {
 }
 
 .ga-draw-active {
+
+  #search-container {
+    display: none;
+  }
```
oterral: Can you add ` #search-container` with other components list which are already hidden when draw is  active see [line 225](https://github.com/geoadmin/mf-geoadmin3/blob/gjn_search_draw/src/style/app.less#L225)

gjn:
Good idea. Done.

oterral:
perfect

PR #2530, opened 2015-07-28
Use gaLang and gaTopic in search + other improvements

oterral:
[Test](https://mf-geoadmin3.dev.bgdi.ch/teo_fix_2511/?X=190000.00&Y=660000.00&zoom=1&lang=fr&topic=luftbilder&bgLayer=ch.swisstopo.pixelkarte-grau&layers=ch.swisstopo.lubis-luftbilder_schwarzweiss,ch.swisstopo.lubis-luftbilder_farbe&layers_timestamp=99991231,99991231&catalogNodes=1179,1180,1186)


src/components/search/SearchDirective.js
```diff
@@ -149,40 +153,31 @@ goog.require('ga_search_type_directives');
...
-                $scope.childoptions.currentTopic);
+            url = gaUrlUtils.append(url, 'lang=' + gaLang.get());
+            url = url.replace('{Topic}', gaTopic.get().id);
 
             $scope.childoptions.baseUrl = url;
             $scope.childoptions.query = q;
           }
         } else {
           blockQuery = false;
+          $scope.$broadcast('gaSearchReset');
```
gjn: Using `$scope.childoptions.query = ''` here would have the same effect and we would not need `gaSearchReset` event at all (we do this in the clearInput function already).

src/components/search/SearchDirective.js
```diff
@@ -149,40 +153,31 @@ goog.require('ga_search_type_directives');
...
-      $scope.updateHref = function() {
-        $scope.encodedPermalinkHref = encodeURIComponent(gaPermalink.getHref());
-      };
-
-      $rootScope.$on('$translateChangeEnd', function() {
-        $scope.lang = $translate.use();
+      // We begin to watch the query only when topics are loaded
+      gaTopic.getTopics().then(function() {
+        $scope.topicLoaded = true;
+        if ($scope.query) {
```
gjn: Is it even possible that 'query' contains something here? With swisssearch parameter?

oterral: Yes with the swisssearch parameter

gjn: ok.

src/components/search/SearchDirective.js
oterral: Ah yes nice, I'll use this. Do you think it's not a problem if `scope.query` and `scope.childoptions.query` have not the same value?

gjn: No. It shouldn't be a problem. In main directive, it's reacting to changes and will reset child query to actual value.

But it's maybe not ideal as it is right now.

gjn:
Just found out during testing that on IE9, the swissearch parameter works once out of 5 tries. Looks like timing issue. Note that it's the same bevahiour in current prod, so unrelated to the PR. But giving 5 seconds thought, you might come up with something.

gjn:
jenkins test

gjn:
jenkins does not seem to work with ga3 projects. @alcapat any idea why mf-geoadmin3.ci.bgdi.ch can't be reached?

oterral:
@gjn ther is nothing on mf-geoadmin3.ci.bgdi.ch , the geoadmin code has disappeared from the instance.

gjn:
hmm. The conf directory is gone in /var/www/mf-geoadmin3. We need this so apache finds the correct code.

Why did it disappear?

Code is there at /var/lib/jenkins/jobs/PR_geoadmin3/workspace/

gjn:
Maybe @danduk82 knows why (we are currently migrating jenkins...I believe)

danduk82:
C2C has not yet started the migration. Shall I increase the priority? It's now set to "low" because of the production problems with varnish...

gjn:
No, it's ok. VArnish has top priority.

Still would be great to know why /var/www/vhosts/mf-geoadmin3/conf directory disappeared completely...(And I can't create it by hand).

gjn:
I've run the tests manually. There is a problem in IE9 that is unrelated to this PR (I'll open an issue). Otherwise, it works well.

PR #2535, opened 2015-07-29
[WIP] Add arrow up under dropdown menu

loicgasser:
Small cosmetic change. I'm adding an arrow up in the draw menu.

arrow_up

### DIFF
--- a/src/components/draw/DrawDirective.js
+++ b/src/components/draw/DrawDirective.js
@@ -174,6 +174,7 @@ goog.require('ga_permalink');
           scope.isMeasureActive = false;
           scope.options.isProfileActive = false;
           scope.statusMsgId = '';
+          scope.dropDownOpen = false;
 
           // Add select interaction
           var select = new ol.interaction.Select({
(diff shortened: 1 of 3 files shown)


oterral:
Can't we just add an element with the `class="arrow"`  like in bootstrap popover ?
http://getbootstrap.com/javascript/#popovers-examples

loicgasser:
With the actual html it might require a lot more changes.

loicgasser:
http://codepen.io/chriscoyier/pen/lotjh cool link

oterral:
Ahah enorme

De : Gasser Loïc [mailto:notifications@github.com]
Envoyé : mercredi 29 juillet 2015 14:34
À : geoadmin/mf-geoadmin3
Cc : Terral Olivier swisstopo
Objet : Re: [mf-geoadmin3] Add arrow up under dropdown menu (#2535)


http://codepen.io/chriscoyier/pen/lotjh cool link

loicgasser:
So it's basically 2 triangles on top of each other (with 1 px offset on top) other with a different border-bottom-color. Can't do anything easier. Google does it for gmail profile. ;)

loicgasser:
On a side note we could change the dropdown position on scroll up and down (and even resize)

oterral:
My concern is only to have an arrow without need of javascript. The positionment of the dropdown menu is already a mess.  That's why if we can position an arrow without js it will be easier to maintain in the future.

oterral:
I think with a simple html like this and an appropriate css that could work. 
```
<div class="dropdown-menu"
  <div class="arrow"></div>
  <ul class="old-dropdownmenu">
```
what do you think ?

loicgasser:
Yes we can wrap it in a div, it would simplify JS indeed.

If you use only one element you'll need the CSS selectors before/after. I don't think you can simply add the arrow class to do the trick, but I may be wrong.
Looking at http://getbootstrap.com/2.3.2/javascript.html#dropdowns
they do have an example with navbar menus, but don't have the same with simple dropdowns.

oterral:
yes the arrow class is effective only with dropodown but we can copy the css to use it with dropdown menu

PR #2539, opened 2015-07-30
Use q.all to know when config files are available

oterral:
SImplify the loading page workflow.

[Test](https://mf-geoadmin3.dev.bgdi.ch/teo_q_all/?topic=funksender)

### DIFF
--- a/src/components/catalogtree/example/index.html
+++ b/src/components/catalogtree/example/index.html
@@ -111,7 +111,7 @@
                'https://mf-chsdi3.dev.bgdi.ch/rest/services/all/MapServer/layersConfig' +
                '?lang={Lang}';
            gaLayersProvider.legendUrlTemplate =
-               'https://mf-chsdi3.dev.bgdi.ch/rest/services/{Topic}/MapServer/{Layer}/legend' +
+               'https://mf-chsdi3.dev.bgdi.ch/rest/services/all/MapServer/{Layer}/legend' +
                '?lang={Lang}';
         }]);
 
@@ -152,7 +152,7 @@
             $scope.map = map;
             var loadedBackground = false;
 
-            $scope.$on('gaLayersChange', function(evt, newLayers) {
+            gaLayers.loadConfig().then(function() {
               if (!loadedBackground) {
                 var layer = gaLayers.getOlLayerById('ch.swisstopo.pixelkarte-farbe');
                 layer.background = true;
--- a/src/components/map/MapService.js
+++ b/src/components/map/MapService.js
@@ -766,35 +766,11 @@ goog.require('ga_urlutils_service');
               .replace('{Lang}', lang);
         };
 
-        var getMetaDataUrl = function(topic, layer, lang) {
+        var getMetaDataUrl = function(layer, lang) {
           return legendUrlTemplate
-              .replace('{Topic}', topic)
               .replace('{Layer}', layer)
               .replace('{Lang}', lang);
         };
-
-        // Load layers for a given topic and language. Return a promise.
-        var lastUrlUsed;
-        var loadLayersConfig = function() {
-          var url = getLayersConfigUrl(gaLang.get());
-          // Avoid loading twice the same layers config (happens on page load)
-          if (lastUrlUsed == url) {
-            return;
-          }
-          lastUrlUsed = url;
-          return $http.get(url).then(function(response) {
-            var isLabelsOnly = angular.isDefined(layers);
-            layers = response.data;
-            if (isLabelsOnly) {
-              $rootScope.$broadcast('gaLayersTranslationChange', layers);
-            } else {
-              $rootScope.$broadcast('gaLayersChange', layers);
-            }
-          }, function(response) {
-            layers = undefined;
-          });
-        };
-        loadLayersConfig();
 
         // Function to remove the blob url from memory.
         var revokeBlob = function() {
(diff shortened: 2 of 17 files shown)


src/components/catalogtree/example/index.html
```diff
@@ -152,7 +152,7 @@
             $scope.map = map;
             var loadedBackground = false;
 
-            $scope.$on('gaLayersChange', function(evt, newLayers) {
+            gaLayers.loadCOnfig().then(function() {
```
gjn: typo

src/components/map/MapService.js
```diff
@@ -828,13 +804,48 @@ goog.require('ga_urlutils_service');
...
+        };
+
         /**
          * Reurn an array of pre-selected bodId from current topic
          */
         this.getSelectedLayers = function() {
-          if (!layers || !gaTopic.get()) {
-            return;
-          }
           return gaTopic.get().selectedLayers;
```
gjn: I think we should remove this function here, as (pre-) selected layers depend on topic. It should not be part of gaLayers.

In any case, the gaTopic.loadConfig() promise must be resolved before people can call this...this could be indicated in a comment.

src/components/map/MapService.js
```diff
@@ -843,13 +854,12 @@ goog.require('ga_urlutils_service');
...
-            return;
-          }
+          var that = this;
           return $.map(gaTopic.get().backgroundLayers, function(bodId) {
-            var retVal = {id: bodId,
-                          label: self.getLayerProperty(bodId, 'label')};
+            var retVal = {
+              id: bodId,
+              label: that.getLayerProperty(bodId, 'label')
+            };
```
gjn: Maybe mention in comment that this can only be used after gaTopic.loadConfig and gaLayers.loadConfig are both resolved.

src/components/topic/TopicService.js
```diff
@@ -71,10 +71,15 @@ goog.require('ga_permalink');
         });
 
         // Returns a promise that is resolved when topics are loaded
-        this.getTopics = function() {
+        this.loadConfig = function() {
           return topicsP;
         };
 
+        // Returns a promise that is resolved when topics are loaded
```
gjn: Is `topics` really a promise?

src/components/catalogtree/example/index.html
oterral: mein gott!!!!

src/components/map/MapService.js
oterral: I fully agree I don't think this function should be there. And it's only used once in all our code I guess.

src/components/map/MapService.js
oterral: This function is a bit like getSelectedLayer, I don't think it should be there.

+1 for the comment

src/components/topic/TopicService.js
oterral: not anymore

src/components/topic/TopicService.js
```diff
@@ -71,10 +71,15 @@ goog.require('ga_permalink');
         });
 
         // Returns a promise that is resolved when topics are loaded
-        this.getTopics = function() {
+        this.loadConfig = function() {
           return topicsP;
```
gjn: We probably have a race condition here. I try to explain.

When the loadConfig promise is resolved, 2 things happen:
A) init the gaTopic service (set topics and topic variables on line 66), which leads to the first `gaTopicChange` event being triggered
B) init other components. In those 'then' functions, we attach to `gaTopicChange` event.

If the 'then' of A is called before the 'then' of B, the component in B will miss the first `gaTopicChange` event, no?

src/components/map/MapService.js
gjn: This here is tricky. It depends on topic, but as well on gaLayers (for the getLayerProperty function). So it can be argued that it belongs here...

gjn:
Code looks good to me. Will do some tests and in-depth review in a couple of places I don't fully understand yet.

src/components/topic/TopicService.js
oterral: yes but I thnik  we don't care because the  call of  'then' is like the first `gaTopicChange` event it miss.

Hard to explain:)

src/components/map/MapService.js
oterral: I've digged a little bit yesterday and it seems the 'that.getLayerProperty(bodId, 'label')' is not used by the background selector directive. but I'm not 100% sure.

src/components/topic/TopicService.js
gjn: If components are aware of it, then ok.

But there's another race condition :=)

If a component uses `gaTopic.getTopics` in their 'then' function before the service's 'then' function is called, it will be undefined, no? (same for gaTopic.get())

src/components/map/MapService.js
```diff
@@ -828,14 +804,42 @@ goog.require('ga_urlutils_service');
...
+        // Load layers config
+        var lastLangUsed;
+        var loadLayersConfig = function(lang) {
+          if (lastLangUsed == lang) {
             return;
           }
-          return gaTopic.get().selectedLayers;
+          lastLangUsed = lang;
+          var url = getLayersConfigUrl(lang);
+          return $http.get(url).then(function(response) {
```
gjn: I think we need to use the same machanism in gaTopic as is used here. Here, we publish the promise returned by the `then` function. In gaTopic, publish the promise returned by the request.

src/components/topic/TopicService.js
oterral: It will be impossible if wereturn the promise of the `then` function of the service not the promise of the `$http.get` so in th $q.all `then `calls the topics and topic are setted.

So there is a problem yes :)

src/components/map/MapService.js
oterral: yes just saw it

src/components/map/MapService.js
gjn: That might be true. Initially, background layers were named the same way in layermanager/catalog and in background selector; but that was not wished anymore. So we created separate translations for background layers. Not sure it's valid for every background layer we have...most probably yes.

src/components/topic/TopicService.js
gjn: Lines 65-71 should not be done with the topicsP promise. It should be done in the loadConfig functions on success, before resolving the topicsP promise. Then I think we avoid the race condition. (and it's more proper anyway).

src/components/map/MapService.js
oterral: I thnik things have changed since we have the new bg layer selector. We don't care (don't know if this is wanted) of the list of bg layers of the topic, we only care of the first element to determine which one is displayed by default.

The list of bg layers are currently always the same (hardcoded) and so the translation id are defined in this list.

gjn: So let's remove what is not needed anymore.

oterral:
Ready for antoher useful review @gjn

src/components/map/MapService.js
```diff
@@ -1777,6 +1723,21 @@ goog.require('ga_urlutils_service');
             });
           }
         };
+
+        // Add permalink layers when topics and layers config are loaded
+        $q.all([gaTopic.loadConfig(), gaLayers.loadConfig()]).then(function() {
+          if (!layerSpecs.length) {
+            // We idd topic selected layers if no layers parameters provided
```
gjn: Typo

src/components/topic/TopicService.js
```diff
@@ -60,21 +59,19 @@ goog.require('ga_permalink');
...
-        });
+        var topicsP = loadTopicsConfig(topicsUrl);
 
         // Returns a promise that is resolved when topics are loaded
-        this.getTopics = function() {
+        this.loadConfig = function() {
           return topicsP;
         };
 
+        // Returns the topics loaded. Should be used only when fter the load
```
gjn: Typoe

gjn:
Only found 2 typos in the comments. Otherwise I'm quite happy with it.

I think it's good that application initialisation is handled using promises (happens once, depends on external events) and other events that are user trigger is handled by triggering events (can happen anytime, depends on user interaction).

oterral:
typos fix 
I squash then merge when you want .

oterral:
Squashed

oterral:
Thanks a lot for the inspirations and review @gjn and @Jenselme

gjn:
It's well done. Thanks alot.

PR #2540, opened 2015-07-30
Add new test with wordforms

alcapat:
rte berne -> route de berne

### DIFF
--- a/test/selenium/swisssearch_test.js
+++ b/test/selenium/swisssearch_test.js
@@ -4,6 +4,7 @@ var webdriver = require('browserstack-webdriver');
 var assert = require('assert');
 
 var QUERYSTRING_OF_RARON = "X=128114.80&Y=629758.13&zoom=10";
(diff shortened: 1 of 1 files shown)


gjn:
Why remove an existing test?

alcapat:
@gjn sorry it's a mistake. i only want to add a new test, not replace. i will correct that soon

gjn:
I think you can add it in the same file, same function. Just start load a new url with `driver.get`, it's like typing a new address in the address bar.

The first test is useful I think. It checks if a click on the search result does indeed navigate to the selected point.

alcapat:
@gjn it's done.

gjn:
Very nice. Thanks alot.

alcapat:
@gjn i don't know if we need test other wordforms and/or in other language ?

gjn:
As it's run only in one browser, it might make sense to add more. But it's also not a top feature of map.geo...I'll let you decide

All wordforms: https://github.com/geoadmin/service-sphinxsearch/blob/master/conf/wordforms_main.txt

PR #2550, opened 2015-08-05
Add new test for wordforms (DE + IT)

alcapat:
New wordforms tests, Platz (DE) and Piazza (IT), french already done with Route


cedricmoullet:
@alcapat can we merge ?

alcapat:
@cedricmoullet now you can

PR #2554, opened 2015-08-06
[WIP] Add new terrain layer in the backgroundselector

fredj:


cedricmoullet:
Shouldn't we have a "3dMode" boolean somwhere in order to avoid that this BG appears already in map.geo.admin.ch ?

cedricmoullet:
@pfanguin could you please prepare a terrain-rect.png file symbolizing a 3D view (what about showing the DTM of the Cervin ?)

gjn:
Yes, we should have a gaGlobalOptions.3ddev parameter that is false on standard builds and true when developing 3d. This parameter will be removed once we go live with 3d.

It should be activated either a) by permalink parameter or b) inside files specifically created for 3d or c) both. I will prepare the general setup of this (in case @oterral @fredj don't already work on this, let me know)

Also, we should make sure to extent unittests and selenium tests to test more thoroughly all the components that will be affected by the 3ddev parameter. We need to assure that the work we do on 3d does in no way affect what we currently have.

Regarding this PR: backgroundlayers are driven by topic, so this should be ok even without the 3ddev switch. Of course, unittests or e2e tests to assure this are needed.

oterral:
@gjn and @fredj there is one in my branch: https://github.com/geoadmin/mf-geoadmin3/compare/teo_reality?expand=1#diff-8c464debc76779ec5061991bb4161492R158

gjn:
@oterral isn't this a switch for the app wether to show 3d view or 2d view?  This show3D parameter will be set to true when 3d background is selected. 

The 3ddev parameter is a little different as it toggles 3d integration (3d background layer is not shown, other 3d relevant code is excluded, etc)

oterral:
Yes you're right @gjn

cedricmoullet:
I'm not sure it's always topic dependent, since all topics have all bglayers.

gjn:
Global dev3d switch is here: https://github.com/geoadmin/mf-geoadmin3/pull/2556/

fredj:
I've updated the code to take the `gaGlobalOptions.dev3d` variable into account

gjn:
Could you rebase?

gjn:
@cedricmoullet We need translations for the labels as well.

fredj:
rebased

gjn:
@oterral You want this in master already so you can attach activation of 3d to background layer selector (instead of checkbox) in your branch? Or wait for picture/translations here?

fredj:
Please note that this code is largely untested ...

gjn:
It works fine.

fredj:
that's right that it can be merged to attach the 3d activation

oterral:
Yes we can merge it

PR #2556, opened 2015-08-06
Adding 3d dev switch

gjn:
This adds a global dev3d switch indicating that the page is loaded with 3d features enabled. See discussion https://github.com/geoadmin/mf-geoadmin3/pull/2554

It also adds a ALPHA ribbon in top left corner to indicate status.

To load page with dev3d switch, you can do (replace user specifc url part accordingly):
https://mf-geoadmin3.dev.bgdi.ch/ltjeg/3d
https://mf-geoadmin3.dev.bgdi.ch/ltjeg/?dev3d=true

For source mode
https://mf-geoadmin3.dev.bgdi.ch/ltjeg/src/3d
https://mf-geoadmin3.dev.bgdi.ch/ltjeg/src/?dev3d=true

Switch should be used to exclude 3d specific code from current page. Use either `gaGlobalOptions.dev3d` or `scope.globals.dev3d` to check state in-code.

### DIFF
--- a/apache/app.mako-dot-conf
+++ b/apache/app.mako-dot-conf
@@ -26,6 +26,12 @@ Header set Access-Control-Allow-Origin "*"
 Header always set Access-Control-Allow-Methods "POST, GET, OPTIONS"
 Header always set Access-Control-Allow-Headers "x-requested-with, Content-Type, origin, authorization, accept, client-security-token"
 
+# To be removed once 3d goes live
+RedirectMatch ^${apache_base_path}/3d$ ${apache_base_path}/3d/
+RedirectMatch ^${apache_base_path}/3d/src$ ${apache_base_path}/src/?dev3d=true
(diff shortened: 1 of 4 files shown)


oterral:
Don't you think we should have also a variable in makefile ? I think about custom compilations/loading  for 3d

gjn:
I can make it a Makefile variable.

But I think we will have custom compilation/loading after it's released as well, so it should not be this variable. Or maybe I don't fully understand.

oterral:
yes I can add it myself in my branch

gjn:
2D (current app) should not be affected by this new parameter. Cesium/ol3-cesium should be loaded only when 3d background/topic is selected.

So we'll merge this and you rebase your branches, ok?

gjn:
dev3d variable is the price we pay to have everything in master. If we would develop 3d on a branch, there would be no dev3d.

PR #2557, opened 2015-08-06
[3D] Add ol3cesium

oterral:
Ok I've succeded make it work on src and prod mode :)

This PR adds:
      - an `ol3cesium` target  which dl the project recursively and then build cesium, ol and ol3cesium, then copy files we need in `src/lib`

[https://mf-geoadmin3.dev.bgdi.ch/teo_reality/?bgLayer=ch.swisstopo.terrain.3d&dev3d=true](https://mf-geoadmin3.dev.bgdi.ch/teo_reality/?bgLayer=ch.swisstopo.terrain.3d&dev3d=true)


Makefile
```diff
@@ -223,9 +239,11 @@ prd/robots.txt: scripts/robots.mako-dot-txt .build-artefacts/last-deploy-target
 prd/lib/: src/lib/d3-3.3.1.min.js \
 	    src/lib/bootstrap-datetimepicker.min.js  \
 	    src/lib/IE9Fixes.js \
-	    src/lib/jQuery.XDomainRequest.js
+	    src/lib/jQuery.XDomainRequest.js \
+			src/lib/ol3cesium.js \
+			src/lib/Cesium
```
elemoine: Indent is not the same as the previous lines.

src/components/map/MapDirective.js
```diff
@@ -79,6 +80,38 @@ goog.require('ga_permalink');
...
+                  ol3d.setEnabled(false);
+                  var baseUrl = '//api3.geo.admin.ch/mapproxy/service';
+                  var ip = new Cesium.WebMapServiceImageryProvider({
+                    url: baseUrl,
+                    layers: 'ch.swisstopo.swissimage'
+                  });
+                  var layer = scene.imageryLayers.addImageryProvider(ip);
+                  layer.show = true;
+                }
+                if (ol3d) {
```
elemoine: May `ol3d` still be `undefined` at this point?

elemoine:
I just added a few minor comments. Otherwise looks good to me.

src/components/map/MapDirective.js
oterral: No

oterral:
PR updated with lazyload and activation the permalink paramter `dev3d=true`

gberaudo:
This PR looks reasonable to me and I agree we could merge it.
As a side note:
- compiling OL3 and OL3-Cesium together would avoid duplicating the
  goog library payload. However it is probably not significant relative to
  the size of Cesium.js.
- compiling app3D code and OL3-Cesium together would be more flexible
  by allowing to extend OL3-Cesium with custom behaviour. It may become
  a problem later (eg. we may want to extend olcs.FeatureConverter).
  Additionally, a minor implovement is that the app3D code could be lazyloaded
  only when 3D is required by the user.

See the POC using `--module` parameter at https://github.com/gberaudo/experiment-closure-modules

elemoine:
> compiling app3D code and OL3-Cesium together would be more flexible by allowing to extend OL3-Cesium with custom behaviour.

Note that this would require a lot of changes to the mf-geoadmin3 js code.

Makefile
```diff
@@ -29,6 +30,8 @@ RESOLUTIONS ?= '[650.0, 500.0, 250.0, 100.0, 50.0, 20.0, 10.0, 5.0, 2.5, 2.0, 1.
 DEFAULT_EPSG ?= EPSG:21781
 DEFAULT_EPSG_EXTEND ?= '[420000, 30000, 900000, 350000]'
 DEFAULT_ELEVATION_MODEL ?= COMB
+DEV3D ?= 'false'
```
gjn: Not needed anymore, no?

gjn:
Very nice work!

PR #2559, opened 2015-08-07
[3D] Adapt to new official urls

gjn:
* Use new official URL for terrain files (https://github.com/geoadmin/3d-forge/pull/111)
* Use tlm as basic layer (there are reasons for this...)

@oterral @gberaudo Let's use [3D] tag for 3D related PR's. Guillaume, you should have access to this repo now. So you can work directly on this instead of a fork.


elemoine:
LGTM.

PR #2560, opened 2015-08-10
[3D] Add a new Controls3d components

fredj:
à la veloland


fredj:
depends on #2568

fredj:
The styling has to be done (images) but the code is ready to be reviewed

src/components/controls3d/partials/controls3d.html
```diff
@@ -0,0 +1,12 @@
+<div class="ga-controls3d">
+  <div class="tilt">
```
oterral: can you use `ga-` prefixed class name please?

src/components/controls3d/Controls3dDirective.js
```diff
@@ -0,0 +1,61 @@
...
+    return {
+      restrict: 'A',
+      templateUrl: 'components/controls3d/partials/controls3d.html',
+      scope: {
+        ol3d: '=gaControls3dOl3d'
+      },
+      link: function(scope, element, attrs) {
+
+        var ol3d = scope.ol3d;
+        goog.asserts.assert(goog.isDef(ol3d));
```
oterral: we don't use `goog` stuff in geoadmin

src/components/controls3d/Controls3dDirective.js
```diff
@@ -0,0 +1,61 @@
+goog.provide('ga_controls3d_directive');
+
+goog.require('goog.asserts');
+goog.require('goog.math');
+(function() {
+
+  var css_rotate = function(element, angle) {
```
oterral: `camelCase`name pls

src/components/controls3d/Controls3dDirective.js
```diff
@@ -0,0 +1,61 @@
...
+        var rotateIndicator = element.find('.rotate .indicator');
+
+        camera.moveEnd.addEventListener(function() {
+          var tiltOnGlobe = olcs.core.computeSignedTiltAngleOnGlobe(scene);
+          css_rotate(tiltIndicator, -tiltOnGlobe);
+          css_rotate(rotateIndicator, -camera.heading);
+        });
+
+        scope.tilt = function(angle) {
+          angle = goog.math.toRadians(angle);
```
oterral: Same here.

src/index.mako.html
```diff
@@ -230,6 +230,8 @@
     </div>
 % endif
 
+    <div ga-controls3d ga-controls3d-ol3d="ol3d" ng-if="ol3d" ng-show="globals.is3dActive"></div>
```
oterral: Can't we use `ng-if="globals.is3dActive"`

fredj: The nice thing about `ng-if` is that it initialize the directive only when the condition is true.
That means that the `ga-controls3d` is sure that the `scope.ol3d` is correct in the constructor.

oterral: Sure,  so an all-in-one `ng-if="globals.is3dActive && ol3d"`

fredj: We can use `ng-if="globals.is3dActive && ol3d"` and no `nf-if` but the components will be re-created every time the 3d mode is activated.
I've updated the code.

gjn:
@fredj State of this PR?

fredj:
The UI (images for the tilt, heading and the directional buttons) needs to be done.

fredj:
demo: https://mf-geoadmin3.dev.bgdi.ch/controls3d/?dev3d=true

cedricmoullet:
@fredj when I test the demo in 3D, i don't see the navigation tools. Any help ?

fredj:
I need to update the demo link

cedricmoullet:
@oterral could you please review and merge ?

src/components/controls3d/Controls3dDirective.js
```diff
@@ -0,0 +1,60 @@
...
+
+        var ol3d = scope.ol3d;
+        var scene = ol3d.getCesiumScene();
+        var camera = scene.camera;
+
+        var tiltIndicator = element.find('.ga-tilt .ga-indicator');
+        var rotateIndicator = element.find('.ga-rotate .ga-indicator');
+
+        camera.moveEnd.addEventListener(function() {
+          scope.$apply(function() {
```
oterral: Do we really need to execute an $apply ? The callback executes only css modification

oterral:
Doesn't work in production mode there is js errors.

fredj:
> Doesn't work in production mode there is js errors.

Probably caused by the html minifier:
```
ng-if="globals.is3dActive && ol3d"
```
becomes:
```
ng-if="globals.is3dActive & ol3d"
```

fredj:
See https://github.com/mankyd/htmlmin/issues/26

fredj:
I've changed the attributes back to `ng-if="ol3d" ng-show="globals.is3dActive"` (instead of `ng-if="globals.is3dActive && ol3d"`) as a workaround for the htmlmin issue.

oterral:
here it should be also the problem https://github.com/geoadmin/mf-geoadmin3/blob/controls3d/src/index.mako.html#L292

fredj:
> here it should be also the problem https://github.com/geoadmin/mf-geoadmin3/blob/controls3d/src/index.mako.html#L292

Yes; compressed to `globals.catalogShown & globals.selectionShown` ...

oterral:
But why we don't have this problem on master ?

fredj:
A new version of htmlmin (https://pypi.python.org/pypi/htmlmin) today ! this is why I didn't catch this error while developing this component

fredj:
See #2624

oterral:
Thx it's merged

src/components/controls3d/Controls3dDirective.js
oterral: @fredj ?

fredj: I systematically use `$apply` when doing stuff outside angular; no matter what the function does. But it's true that it should not make a big difference here because (as you said) only css is changed.

I will remove it

fredj: done

mankyd:
Sorry about this. The issue is fixed in the 0.1.8 release of htmlmin.

oterral:
Thx @mankyd

PR #2566, opened 2015-08-11
Use the topic configuration promise to get the languages from initial topic

Jenselme:


Jenselme:
PR updated to fix failing tests.

src/components/translation/TranslationDirective.js
```diff
@@ -38,6 +39,10 @@ goog.require('ga_translation_service');
         scope.$on('gaTopicChange', function(event, newTopic) {
           scope.langs = newTopic.langs;
         });
+
+        gaTopic.loadConfig().then(function() {
+          scope.langs = gaTopic.get().langs;
+        });
```
oterral: if we use `loadConfig`, wee need to register the callback on gaTopicChange here too to avoid unwanted calls.

Jenselme: Indeed. PR updated.

src/components/translation/TranslationDirective.js
```diff
@@ -1,6 +1,7 @@
 goog.provide('ga_translation_directive');
 
 goog.require('ga_translation_service');
+goog.require('ga_topic_service');
```
oterral: "goog.require classes must be alphabetized" said the compiler

oterral:
I let you time to you fix the comment this time :)

Jenselme:
PR updated

oterral:
Thank you very much for this.

PR #2569, opened 2015-08-12
Add languages to global options

Jenselme:
TranslationService will use the list of languages if the topics are not
leaded yet. So we are sure to use the fallback language if the lang in
permalink or most likely the language we get from the web browser is not
supported. Without this, the loading of layerConfig.json migth fail and
thus the portail will be unsuable.

This assumes that the default topic is available for the fallback
language.

### DIFF
--- a/src/components/translation/TranslationService.js
+++ b/src/components/translation/TranslationService.js
@@ -19,6 +19,10 @@ goog.require('ga_topic_service');
           ($window.navigator.userLanguage ||
           $window.navigator.language).split('-')[0];
 
+      if (gaGlobalOptions.languages.indexOf(lang) === -1) {
+        lang = gaGlobalOptions.translationFallbackCode;
+      }
+
       // Verify if a language is supported by the current topic
       var isLangSupportedByTopic = function(lang, topic) {
         if (!topic) {
--- a/src/geoadmin.mako.appcache
+++ b/src/geoadmin.mako.appcache
@@ -1,6 +1,11 @@
+<%page args="languages" />
 <%
+  import json
   defaultLanguage = 'en'
-  languages = ('de', 'en', 'fr', 'it', 'rm')
+  # The languages variable must contain simple quote to be correctly parsed in
+  # index.html but here it prevents python to parse the JSON. So we manually
+  # remove them.
+  languages = json.loads(languages[1:-1])
 %>CACHE MANIFEST
 # Version ${version}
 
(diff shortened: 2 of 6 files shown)


test/specs/Loader.spec.js
```diff
@@ -36,6 +37,7 @@ beforeEach(function() {
       ],
       defaultTopicId: 'sometopic',
       translationFallbackCode: 'somelang',
+      languages: [navLang, 'rm', 'somelang'],
```
oterral: I don't think we need `navLang` in this list because it's not true all the time. Put 'de','en' and 'fr' instead.

src/components/translation/TranslationService.js
```diff
@@ -21,8 +21,10 @@ goog.require('ga_topic_service');
 
       // Verify if a language is supported by the current topic
       var isLangSupportedByTopic = function(lang, topic) {
-        if (!topic) {
+        if (!topic && gaGlobalOptions.languages.indexOf(lang) > -1) {
```
oterral: I think we need this test [here](https://github.com/ioda-net/mf-geoadmin3/blob/add-languages-globalVariables/src/components/translation/TranslationService.js#L18) instead  after we set the lang from the browser or permalink.

Jenselme: Yes, it may look better. PR updated.

test/specs/Loader.spec.js
Jenselme: Indeed, removed.

src/components/translation/TranslationService.js
```diff
@@ -19,10 +19,16 @@ goog.require('ga_topic_service');
...
+      if (gaGlobalOptions.languages.indexOf(lang) === -1) {
+        lang = gaGlobalOptions.translationFallbackCode;
+      }
+
       // Verify if a language is supported by the current topic
       var isLangSupportedByTopic = function(lang, topic) {
         if (!topic) {
           return true;
+        } else if (!topic) {
+          return false;
```
oterral: No need of this `if` anymore

Jenselme: Quite true. PR updated.

src/geoadmin.mako.appcache
```diff
@@ -1,6 +1,11 @@
+<%page args="languages" />
 <%
```
oterral: put `<%` on the first line otherwise the appcache file generated   is not valid

oterral:
I've merged you rchange in #2582 thx @Jenselme

Jenselme:
@oterral  thanks. You didn't give me the time to propose the correction for appcache ^^

oterral:
Yes sorry :)

PR #2583, opened 2015-08-17
[3D] Webgl detection

oterral:
We don't use `ol.has.WEBGL` variable because `ol.has.WEBGL` rely on `ol.ENABL_WEBGL` variable and in mf-geoadmin this viriable is `true` in debug mode but `false` in production.

### DIFF
--- a/src/components/BrowserSnifferService.js
+++ b/src/components/BrowserSnifferService.js
@@ -117,10 +117,9 @@ goog.require('ga_permalink');
         phone: mobile && testSize(480),
         events: eventsKeys,
         embed: embed,
-        isInFrame: ($window.location != $window.parent.location)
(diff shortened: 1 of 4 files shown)


oterral:
I change the target branch

elemoine:
See https://github.com/openlayers/ol3/pull/4016.

elemoine:
So, after discussions, openlayers/ol3#4016 won't be merged, as we don't want `ol.has.WEBGL` to evaluate to `true` when OpenLayers is compiled with `ol.ENABLE_WEBGL` set to `false`.

You can look at https://github.com/openlayers/ol3/blob/master/src/ol/has.js#L125-L139 if you want to improve the WebGL detection code.

PR #2584, opened 2015-08-17
[3D] Webgl detection 

oterral:
We don't use `ol.has.WEBGL` variable because `ol.has.WEBGL` rely on  `ol.ENABLE_WEBGL` variable and in mf-geoadmin this variable is `true` in debug mode but `false` in production.

### DIFF
--- a/src/components/BrowserSnifferService.js
+++ b/src/components/BrowserSnifferService.js
@@ -117,10 +117,9 @@ goog.require('ga_permalink');
         phone: mobile && testSize(480),
         events: eventsKeys,
         embed: embed,
-        isInFrame: ($window.location != $window.parent.location)
+        isInFrame: ($window.location != $window.parent.location),
+        webgl: !(typeof WebGLRenderingContext === 'undefined')
       };
     };
-
   });
-
 })();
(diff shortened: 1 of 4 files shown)


elemoine:
> viriable is true in debug mode but false in production.

Can you explain why? Is this a problem in OpenLayers?

oterral:
`ol.ENABLE_WEBGL` is false in produstion mode to avoid having the all webgl part in the compiled file.

elemoine:
> The problem in OpenLayers is that the detection of WEBGL is deactivated when ol.ENABLE_WEBGL=false but probably it's wanted.

Yes, I think it's a problem. Things in `ol.has` are about the browser capabilities. I think they should not depend on `ol.ENABLE_*` values. If we change ol you will be able to use `ol.has.WEBGL`, right?

gjn:
Code LGTM, but agree with @elemoine about the ol variables.

Side note: could we dynamically load the webgl stuff that is compiled in when ol.ENABLE_WEBGL is true? This would allow us to use webGL for 2D display as well.

oterral:
@elemoine yes I guess . It worked when `ol.ENABLE_WEBGL = true`

elemoine:
> Side note: could we dynamically load the webgl stuff that is compiled in when ol.ENABLE_WEBGL is true?

I don't think so. At least I don't see how this could be possible…

PR #2593, opened 2015-08-19
[3D] Add 3d params to the permalink

fredj:


src/components/map/MapDirective.js
```diff
@@ -80,6 +81,40 @@ goog.require('ga_permalink');
...
+                }
+                camera.setView({
+                  positionCartographic: position,
+                  heading: params.heading ? parseFloat(params.heading) : undefined,
+                  pitch: params.pitch ? parseFloat(params.pitch) : undefined,
+                  roll: 0.0
+                });
+
+                // update permalink
+                camera.moveEnd.addEventListener(function() {
```
gjn: I don't know how often this even is triggered. But we should probably debounce this update with gaDebounce (as is done with map parameters: https://github.com/geoadmin/mf-geoadmin3/blob/d88ca65a4b5592f375625c7264c597f6b070ee2b/src/components/map/MapDirective.js#L77)

gjn:
linting is not happy.

gjn:
Works well.

I propose to remove some digits. We have 13-15 digits per parameter. We should reduce them to 'readable' and still assure repeatability of position within a certain precision.

Also, I think we discussed to have the angles in degrees instead of radians for readability. Not sure, pinging @davidoesch .

fredj:
All angles are in radians and the digits are now limited:
 - 5 digits for the longitude and latitude
 - 3 digits for the heading and the pitch
 - 0 digits for the elevation

cedricmoullet:
any demo link ?

fredj:
dmo link:
https://mf-geoadmin3.dev.bgdi.ch/3d_permalink/src/?dev3d=true&X=171521.69&Y=631167.09&zoom=6&lang=en&topic=ech&bgLayer=ch.swisstopo.terrain.3d&lon=0.13577&lat=0.81431&elevation=2768&heading=54.131&pitch=-12.279

cedricmoullet:
- If I go back to 2D and move the map, Y and X are not updated.
- Do we want to have Y/X and lon/lat/elevation/heading/pitch always together in the permalink or should it be 2D/3D mode switch ?

gjn:
According to the doc, we either have 2D (y/x/zoom) or 3D (lon/lat...) in the permalink.

Also, we should use degrees for all angles.

davidoesch:
digits : 5 digits ( we are consistent with 2d) e g 7.85038, 46.67743 )
angle: yes use angle instead of radians (see google...make it easier to switch from google) use 2 digits

fredj:
Angles unit and digits are now ok

@davidoesch: do we want to remove the 2d parameters (X, Y, zoom) when the 3d view is active (and vice versa) ?

elemoine:
> @davidoesch: do we want to remove the 2d parameters (X, Y, zoom) when the 3d view is active (and vice versa) ?

I think that's what we said yes, but I'll let @davidoesch confirm.

davidoesch:
confirmed : show only 2d parameters in 2d and only 3d parameters in 3d

elemoine:
@fredj, is the [demo link](https://mf-geoadmin3.dev.bgdi.ch/3d_permalink/src/?dev3d=true&X=171521.69&Y=631167.09&zoom=6&lang=en&topic=ech&bgLayer=ch.swisstopo.terrain.3d&lon=0.13577&lat=0.81431&elevation=2768&heading=54.131&pitch=-12.279) still valid? Does it represent the latest state of your branch?

fredj:
@elemoine no, something is missing: the snapping to the nearest resolution when switching from 2d to 3d: https://github.com/geoadmin/mf-geoadmin3/blob/3d_transition/src/js/MainController.js#L116-L119

Without it; the 2d params are not updated because `view.getZoom()` returns `undefined`

src/index.mako.html
```diff
@@ -187,7 +187,7 @@
 
 % endif
     <div id="loader"></div>
-    <div tabindex="1" ga-map ga-map-map="map">
+    <div tabindex="1" ga-map ga-map-map="map" ga-map-ol3d="ol3d">
```
elemoine: I know there are currently no one-time binding expressions in `index.mako.html`, but you could use one here for `ol3d` (`ga-map-ol3d="::ol3d"`).

@gjn, would it make sense to start using one-time binding expressions to minimize the number of watchers?

fredj: +1

src/components/map/MapDirective.js
```diff
@@ -79,6 +88,49 @@ goog.require('ga_permalink');
...
+                  positionCartographic: position,
+                  heading: heading,
+                  pitch: pitch,
+                  roll: 0.0
+                });
+
+                // update permalink
+                camera.moveEnd.addEventListener(function() {
+                  // remove 2d params
+                  gaPermalink.deleteParams(['X', 'Y', 'zoom']);
```
elemoine: Super minor, but I'd have used three calls to `deleteParam` to avoid creating an array and adding a `deleteParams` function to `gaPermalink`.

src/components/map/MapDirective.js
elemoine: This is a good point. @fredj, have you verified how often the listener is called? `gaDebounce` is already used for the update of the 2D map parameters, so it probably makes sense to do the same for the 3D params.

src/components/map/MapDirective.js
```diff
@@ -79,6 +88,49 @@ goog.require('ga_permalink');
 
             map.setTarget(element[0]);
 
+            scope.$watch('::ol3d', function(ol3d) {
```
elemoine: +1 for the one-time binding! :)

elemoine:
> @elemoine no, something is missing: the snapping to the nearest resolution when switching from 2d to 3d: https://github.com/geoadmin/mf-geoadmin3/blob/3d_transition/src/js/MainController.js#L116-L119
> 
> Without it; the 2d params are not updated because view.getZoom() returns undefined

Are you able to update the `3d_transition` instance with the latest code?

elemoine:
I am done with the code review. I can do some testing on the instance when it's up-to-date.

src/components/map/MapDirective.js
fredj: The event is triggered at the end of the 'navigation': when the mouse button is released.

src/components/map/MapDirective.js
fredj: Suggested by @gberaudo !

elemoine: The use of `=gaMapOl3d` creates a watcher. And you create another watcher by calling `$watch` here. So to remove both watchers when `ol3d`  gets a defined value in the user scope, two one-time binding expressions should be used.

fredj:
> Are you able to update the `3d_transition` instance with the latest code?

cherry-pick `3d_transition` in this branch ?

elemoine:
> cherry-pick 3d_transition in this branch ?

Oops sorry. I was referring to this PR's branch (`3d_permalink`)…

elemoine:
This is good to merge in my opinion.

src/components/map/MapDirective.js
gjn: @fredj Is this done?

src/components/map/MapDirective.js
gjn: I really would like the have debouncing here to be consistent and on the save side.

Side note: is it worth to remove the listener when switching back the 2d?

src/components/map/MapDirective.js
fredj: I think so: the two expressions uses one time binding:
 - here with `scope.$watch('::ol3d', function(ol3d) {`
 - in `index.mako.html` with `<div ga-map ga-map-map="map" ga-map-ol3d="::ol3d">`

gjn: +1. Sorry, oversaw the `::`

gjn:
I've been testing this. IMO, there are 2things left:

a. lon/lat should be in degrees (46.x/7.x instead of 0.1 ...) as in [google]()
b. when opening a permalink with given lon/lat, we open in 3d, but we don't open in the correct position. In fact, I assume that the map is in 2d first at default x/y (because they're not in permalink) then transferes to 3D while ignoring the permalink values.

src/components/map/MapDirective.js
fredj: > I really would like the have debouncing here to be consistent and on the save side.

See https://github.com/geoadmin/mf-geoadmin3/commit/d06ff54f03f6f45e2b12651155aee67edf916f74

> Side note: is it worth to remove the listener when switching back the 2d?

It's not needed because the camera is not changed in 2d

fredj:
> a. lon/lat should be in degrees (46.x/7.x instead of 0.1 ...) as in google

very good point, fixed

> b. when opening a permalink with given lon/lat, we open in 3d, but we don't open in the correct position. In fact, I assume that the map is in 2d first at default x/y (because X/Y they're not in permalink) then transferes to 3D while ignoring the permalink values.

it was an isse with https://github.com/geoadmin/mf-geoadmin3/commit/a4bc7f315f7e291e019f5ea96ef54a447bac4819: I didn't test enough with the two PR combined (this one and #2595). sorry about that

src/components/map/MapDirective.js
```diff
@@ -63,22 +64,75 @@ goog.require('ga_permalink');
...
+                if (center && zoom !== undefined) {
+                  var x = center[1].toFixed(2);
+                  var y = center[0].toFixed(2);
+                  gaPermalink.updateParams({X: x, Y: y, zoom: zoom});
+                }
               }
             };
             view.on('propertychange', gaDebounce.debounce(updatePermalink,
                 1000, false));
-            updatePermalink();
```
gjn: This is a regression. 2D parameters arent' in PL anymore after page load and without user interaction.

But I think it's a good thing. And in 3D it's the same now.

gjn:
Nice work. Thanks alot.

@davidoesch This changed current 2D behaviour. We won't have X/Y/zoom in the PL after the page is loaded and the user didn't yet interact with it. They are added only after interaction with the map. I think this makes sense.

src/components/map/MapDirective.js
elemoine: A good thing? Why? The application has a state after initialization. So I don't see why the PL should not reflect that state.

gjn: The state is the default state.

davidoesch: +1 for "We won't have X/Y/zoom in the PL after the page is loaded and the
user didn't yet interact with it" makes sense (and it was like this in
RE2...)

PR #2594, opened 2015-08-20
Fix tests with implicit wait

gjn:
kml tests, which had implicit waits, made up 60% of our browserstack failures.

This PR tries to improve this by removing the implicit waits and instead to rely on a feature of the permalink service. On permalink change, the permalink service first adapts the URL in the address bar and then issues a `gaPermalinkChange` event.

We use this feature in our tests now by checking for the changed permalink value in the document first using seleniums standard `findElement` approach. After this is sucessfuly, we can then check the page URL directly without having to wait.

@oterral Care to take a quick look and merge. It would get rid of some annoying build failures.


oterral:
Lgtm thx

PR #2595, opened 2015-08-20
[3D] Add an animation when switch from or to 3d view

fredj:
demo: https://mf-geoadmin3.dev.bgdi.ch/3d_transition/?dev3d=true


fredj:
todo: snap to the nearest zoom level when switching from 3d to 2d

fredj:
Thanks for any review

fredj:
now in #2593

PR #2597, opened 2015-08-20
Add gaTime service

### LINKED ISSUE #2105: Permalink handling on application load
As discussed with @oterral , when loading the page, the permalink affects our application with it's parameter. We have several of those mechanics inside components. This means that the permalink loading would/will fail, when the component is not present for some reason (reduced page, etc).

We should refactor those.

oterral:
Makes time management independent from component.

That adds:

 - Priority of `time` parameter over `layer_timestamps` parameter
 - Manage only visible timeEnabled layers

Before:
https://map.geo.admin.ch/?time=1992&layers=ch.swisstopo.lubis-luftbilder_schwarzweiss,ch.swisstopo.lubis-luftbilder_schwarzweiss&layers_timestamp=19921231,20041231

After:
https://mf-geoadmin3.dev.bgdi.ch/teo_time/?time=1992&layers=ch.swisstopo.lubis-luftbilder_schwarzweiss,ch.swisstopo.lubis-luftbilder_schwarzweiss&layers_timestamp=19921231,20041231

Fix #2105  and all time permalink problem I've seen.

### DIFF
--- a/src/components/map/MapDirective.js
+++ b/src/components/map/MapDirective.js
@@ -16,7 +16,6 @@ goog.require('ga_permalink');
   module.directive('gaMap',
       function($window, $parse, $rootScope, $timeout, gaPermalink,
           gaBrowserSniffer, gaLayers, gaDebounce, gaOffline, gaGlobalOptions) {
-
         return {
           restrict: 'A',
           scope: {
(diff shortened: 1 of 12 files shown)


oterral:
jenkins test

gjn:
build failure was realted to wms test...we have owschecker tests for wms failing in chsdi3 as well. something fishy is going on with wms.
@oterral Do you happen to know something?

oterral:
Yes something went wrong on wms, Andrea has restarted jsut restarted the instance.

src/components/map/MapDirective.js
```diff
@@ -125,6 +124,60 @@ goog.require('ga_permalink');
...
+
+            var savedTimeStr = {};
+            scope.$on('gaTimeChange', function(evt, time, oldTime) {
+              var switchTimeActive = (!oldTime && time);
+              var switchTimeDeactive = (oldTime && !time);
+              var olLayers = scope.map.getLayers().getArray();
+              var singleModif = false;
+
+              // Detection the time change has been triggered by a layer's
+              // 'propertychange' event.
```
gjn: Why can this only happen when `switchTimeDeactive` is true?

gjn:
our 3ddev effort is tapping mapproxy/wms quite heavily...so this might be related to this. Certainly we should keep that under our eyes...

src/components/map/MapDirective.js
oterral: Each time you modify the time property of layer, we try to display/hide the time selector components if all the layers have the same time value or not.

Special case: when you modify the layer's time and then all the layer haven't the same time value anymore. 
In that specific case we don't want the new `undefined` value of the global time will be set to all the layers, we want the layers keep their last time value.

src/components/map/MapDirective.js
```diff
@@ -125,6 +124,60 @@ goog.require('ga_permalink');
...
+                    break;
+                  }
+                }
+              }
+
+              // In case the time change event has been triggered by a layer's
+              // 'propertychange' event, we do nothing more.
+              // (ex: using the layer manager)
+              if (singleModif) {
+                savedTimeStr = [];
```
gjn: Reset as empty object? (see line 128)

oterral: ah ja  good found

src/components/map/MapDirective.js
gjn: So, when changing the time of a layer in the layerManager and this change results in every layer having the same time, then `singleModif` will be true here, right?

I just can't see how the layermanager triggers a gaTimeChange event. I also don't see where it calls gaTime.set.

gjn: Found it now in the service, called by updateStatus there.

src/components/map/MapService.js
```diff
@@ -1744,6 +1752,8 @@ goog.require('ga_urlutils_service');
                 layerTimestamps);
           }
 
+          gaTime.allowStatusUpdate = true;
```
gjn: I think this activation should be done in the gaTime service itself, but that's probably hardly possible as we would face a race condition when we depend on the $q.all promise in both locations, right?

src/components/map/MapDirective.js
oterral: >So, when changing the time of a layer in the layerManager and this change results in every layer having the same time, then singleModif will be true here, right?

Yes

gjn:
kml test is failing now, I would have a PR for that :) https://github.com/geoadmin/mf-geoadmin3/pull/2594

This PR looks good to me.

gjn:
jenkins test

src/components/map/MapService.js
oterral: Yes exactly, I havn't tested but I think it could potentitally be the case.
Maybe we can trigger an event  gaLayersPermalinkLoaded.

gjn: hmm. Not sure if it's worthwile. I'll let you decide.

gjn:
Merge if you decide against moving `gaTime.allowStatusUpdate = true;`

oterral:
I let this as it is. I'm not 100% satisfied  but it will be easy to modify it in the future if we have better idea.

PR #2590, opened 2015-08-18
3D imagery layers synchronization through plugin

gberaudo:
Create and use an ol3-cesium plugin to synchronize imagery layers.
The plugin is compiled together with ol3-cesium code.

PR #2604, opened 2015-08-25
Migrate Search overlays to regular layer

gberaudo:
Should be rebased and merged after #2590.

[Test](https://mf-geoadmin3.dev.bgdi.ch/gberaudo_search_layer/src/?lang=en&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&dev3d=true&catalogNodes=532&layers=ch.swisstopo.vec200-names-namedlocation&lon=7.01749&lat=46.51861&elevation=76984&heading=359.927&pitch=-89.998)

### DIFF
--- a/src/components/MarkerOverlayService.js
+++ b/src/components/MarkerOverlayService.js
@@ -6,10 +6,30 @@ goog.provide('ga_marker_overlay_service');
   module.provider('gaMarkerOverlay', function() {
 
     this.$get = function(gaStyleFactory) {
-      var overlay, bbox, isAlwaysVisible;
-      var marker = $('<div></div>')
-          .addClass('ga-crosshair')
-          .addClass('marker');
+      var bbox, isAlwaysVisible;
+      var source, layer;
+      var feature = null;
+
+      /** @const */
+      var style = new ol.style.Style({
+        image: new ol.style.Icon({
+          anchor: [0.5, 46],
+          anchorXUnits: 'fraction',
+          anchorYUnits: 'pixels',
+          src: 'img/marker.png'
+        })
+      });
+
+      function initialize(map) {
+        if (!layer) {
+          source = new ol.source.Vector();
+          layer = new ol.layer.Vector({
+            source: source
+          });
+          layer.set('altitudeMode', 'clampToGround');
+          map.addLayer(layer);
+        }
+      }
 
       function isPointData() {
         return bbox && (bbox[2] - bbox[0]) <= 1 &&
@@ -17,38 +37,38 @@ goog.provide('ga_marker_overlay_service');
       }
 
       function addMarker(map, center, extent, visible) {
+        initialize(map);
         bbox = extent;
         removeMarker(map);
-        overlay = new ol.Overlay({
-          element: marker.get(0),
-          position: center,
-          stopEvent: false
+        feature = new ol.Feature({
+          geometry: new ol.geom.Point(center),
+          style: style
         });
         isAlwaysVisible = visible;
         setVisibility(map.getView().getZoom());
-        map.addOverlay(overlay);
+        source.addFeature(feature);
       }
 
       function setVisibility(zoom) {
         if (isAlwaysVisible) {
-          marker.removeClass('hide');
+          feature.setStyle(style);
           return;
         }
-        if (overlay) {
+        if (feature) {
           if (!isPointData()) {
             if (zoom > 6) {
-              marker.addClass('hide');
+              feature.setStyle(null);
             } else {
-              marker.removeClass('hide');
+              feature.setStyle(style);
             }
           }
         }
       }
 
       function removeMarker(map) {
-        if (overlay) {
-          map.removeOverlay(overlay);
-          overlay = null;
+        if (feature) {
+          source.clear();
+          feature = null;
         }
       }
 
--- a/src/components/MarkerOverlayService.js
+++ b/src/components/MarkerOverlayService.js
@@ -1,32 +1,24 @@
 goog.provide('ga_marker_overlay_service');
+
+goog.require('ga_map_service');
+goog.require('ga_styles_service');
+
 (function() {
 
-  var module = angular.module('ga_marker_overlay_service', []);
+  var module = angular.module('ga_marker_overlay_service', [
+    'ga_map_service',
+    'ga_styles_service'
+  ]);
 
   module.provider('gaMarkerOverlay', function() {
 
-    this.$get = function(gaStyleFactory) {
-      var bbox, isAlwaysVisible;
-      var source, layer;
-      var feature = null;
-
-      /** @const */
-      var style = new ol.style.Style({
-        image: new ol.style.Icon({
-          anchor: [0.5, 46],
-          anchorXUnits: 'fraction',
-          anchorYUnits: 'pixels',
-          src: 'img/marker.png'
-        })
-      });
+    this.$get = function(gaStyleFactory, gaMapUtils) {
+      var bbox, isAlwaysVisible, layer;
 
       function initialize(map) {
         if (!layer) {
-          source = new ol.source.Vector();
-          layer = new ol.layer.Vector({
-            source: source
-          });
-          layer.set('altitudeMode', 'clampToGround');
+          layer = gaMapUtils.getFeatureOverlay([],
+              gaStyleFactory.getStyle('marker'));
           map.addLayer(layer);
         }
       }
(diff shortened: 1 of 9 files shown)


gberaudo:
In this PR, a regular `ol.layer.Layer` is used instead of the `ol.Overlay`.
The layer is detected by OL3-Cesium and automatically synchronized to Cesium.

It looks good and I think we can merge it.

However, there is still substantial work to do for ensuring the layer stays on top of the other vector layers.

gjn:
The `always on top` need to be done before merging into master. Put to [WIP] because of it.

oterral:
The problem is we have unmanaged layer everywhere  (draw  interaction for example or geolocation). It was a nice feature to don't have to manage this `always on top` thing. 

Don't you think it will be better to have access of unmanaged layer from ol and use the `Vector Synchronizer` to display it on cesium ?

src/components/MarkerOverlayService.js
```diff
@@ -6,10 +6,30 @@ goog.provide('ga_marker_overlay_service');
...
+
+      /** @const */
+      var style = new ol.style.Style({
+        image: new ol.style.Icon({
+          anchor: [0.5, 46],
+          anchorXUnits: 'fraction',
+          anchorYUnits: 'pixels',
+          src: 'img/marker.png'
+        })
+      });
```
oterral: Usually we define style in `gaStyleFactory` service. See https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/StylesService.js#L138

gberaudo:
@oterral I opened an issue in OL3 so that we discuss what can be done:
https://github.com/openlayers/ol3/issues/4053

I propose several solutions; is the first one what you have in mind?

gjn:
From those proposals, I tend to think that 2 or 3 is the most reasonable. It fits well with css z-index approach (and ol3 could even read the z-index for layers from css....)

1) does not cover the ordering of multiple unmanaged layers, which would need to be managed by the user (rare use case, but might exists). Also, unmanaged layers are managed again.
2) doesn't break current usage of unmanaged layers
3) probably ideal, but will break current usage of unmanaged layers (clients need to adapt code)

oterral:
thx @gberaudo, nice summary.

Yes 1) seems the easisest way.

The problem of 2) and 3) is you have a Collection where the order doesn't reflect what you see and that can be a problem when you want to insert a Layer somewhere in the Collection or if you specify 2 layers with the same z-index or 2 layers without z-index. It seems to me that could be a mess to manage :).

Unmaniged layers are temporary layers, they are not displayed all the time. They are often display for a low period of time.

cedricmoullet:
@gberaudo could you please rebase ?

gberaudo:
The second proposal was merged in OL3.
Work was done on ol3-cesium to fix the blue flash when adding/updating/removing layers and using the Z-index available from OL3. The PR review https://github.com/openlayers/ol3-cesium/pull/241 is ongoing.
FlyTo 3D animation was added for locations and features.
The PR was rebased on master.

oterral:
Very nice!!! can you create a test link pls `make deploybranch`

gberaudo:
Here is a link: https://mf-geoadmin3.dev.bgdi.ch/gberaudo_search_layer/src/?lang=en&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&dev3d=true&catalogNodes=532&layers=ch.swisstopo.vec200-names-namedlocation&lon=7.01749&lat=46.51861&elevation=76984&heading=359.927&pitch=-89.998

gberaudo:
This PR is now ready for review.

cedricmoullet:
The markers are supposed to be on the terrain ?
I can see this issue:
image

gberaudo:
Which marker is it? All locations I have tried so far were displayed
correctly on terrain.

oterral:
@gberaudo I got one : search for `Alpiner Gipfel Stockhorn (VS) - Zermatt` then  if you move the map, the marker is displaying in the moutain.
stochhorn

cedricmoullet:
"Rue des Berges 37 Payerne"

gberaudo:
I was not able to reproduce the issue at Payerne and could not find the `Alpiner Gipfel Stockhorn (VS) - Zermatt` location.
However I could find a case, with `Gebaeude Einzelhaus Restaurant Alpenruh (BE) - Reichenbach im Kandertal`, where the symbol is cut by the mountain. For this case it is the expected behaviour as we want to have symbols behind a mountain hidden by it.

It might be the same cause in Payern if for some reason the depth was not updated correctly due to some glitch with the terrain or LOD race condition. @cedricmoullet are you able to consistently reproduce the issue?

payerne
through_mountain

cedricmoullet:
No i can't reproduce....

2015-09-09 17:59 GMT+02:00 gberaudo <notifications@github.com>:

> I was not able to reproduce the issue at Payerne and could not find the Alpiner
> Gipfel Stockhorn (VS) - Zermatt location.
> However I could find a case, with Gebaeude Einzelhaus Restaurant Alpenruh
> (BE) - Reichenbach im Kandertal, where the symbol is cut by the mountain.
> For this case it is the expected behaviour as we want to have symbols
> behind a mountain hidden by it.
>
> It might be the same cause in Payern if for some reason the depth was not
> updated correctly due to some glitch with the terrain or LOD race
> condition. @cedricmoullet <https://github.com/cedricmoullet> are you able
> to consistently reproduce the issue?
>
> [image: payerne]
> <https://cloud.githubusercontent.com/assets/7294662/9767041/759bdadc-571c-11e5-98c4-f274334e1dfd.png>
> [image: through_mountain]
> <https://cloud.githubusercontent.com/assets/7294662/9767042/75a73e04-571c-11e5-9727-7925a409b0b3.png>
>
> —
> <https://github.com/geoadmin/mf-geoadmin3/pull/2604#issuecomment-138956352>
> .
>



-- 

<http://t.sidekickopen16.com/e1t/c/5/f18dQhb0S7lC8dDMPbW2n0x6l2B9nMJN7t5XYgf6M1WN1qfVVMRYFRqW8qlQyH56dByMf1q8vLz02?t=http%3A%2F%2Ftwitter.com%2Fcedricmoullet&si=6057645868318720&pi=5b97c8f5-2860-4da3-904c-ee753ebe4e09>

cedricmoullet:
@gjn @oterral is is ready to merge ?

src/components/search/SearchTypesDirectives.js
```diff
@@ -443,8 +472,17 @@ goog.require('ga_urlutils_service');
...
-                gaPreviewFeatures.zoom($scope.map,
-                    geojsonParser.readFeature(f));
+                var feature = geojsonParser.readFeature(f);
+                var ol3d = $scope.ol3d;
+                if (ol3d && ol3d.getEnabled()) {
+                  var extent = feature.getGeometry().getExtent();
+                  extent = gaPreviewFeatures.getMinimalExtent(extent);
+                  ol3d.getCesiumScene().camera.flyTo({
+                    destination: extentToRectangle(extent)
+                  });
```
oterral: Why not putting this in the `gaPreviewFeatures.zoom`  adding ol3d as parameter of the function. We use this function in other place,  I guess.

src/components/search/SearchTypesDirectives.js
```diff
@@ -332,13 +344,29 @@ goog.require('ga_urlutils_service');
...
+                  gaMapUtils.moveTo($scope.map,
+                                    originToZoomLevel[res.attrs.origin],
+                                    [res.attrs.y, res.attrs.x]);
+                }
               } else {
-                gaMapUtils.zoomToExtent($scope.map, e);
+                if (ol3d && ol3d.getEnabled()) {
+                  ol3d.getCesiumScene().camera.flyTo({
+                    destination: extentToRectangle(e)
+                  });
```
oterral: This should be in  `gaMapUtils.zoomToExtent`

src/components/search/SearchTypesDirectives.js
```diff
@@ -332,13 +344,29 @@ goog.require('ga_urlutils_service');
...
-                gaMapUtils.moveTo($scope.map,
-                                  originToZoomLevel[res.attrs.origin],
-                                  [res.attrs.y, res.attrs.x]);
+                if (ol3d && ol3d.getEnabled()) {
+                  ol3d.getCesiumScene().camera.flyTo({
+                    destination: Cesium.Cartesian3.fromDegrees(
+                                   res.attrs.lon,
+                                   res.attrs.lat,
+                                   3000)
+                  });
```
oterral: This should be in `gaMapUtils.moveTo`

oterral:
I've added some comment to avoid duplicate code in other directives. @gjn does it make sense to you to put zoom/move logic of the globe in gaMapUtils functions?

gjn:
+1 for @oterral proposals.

Also, would like to review this today.

gjn:
Fantastic. I like it.

There's an issue with feature search. Add `ch.swisstopo.lubis-luftbilder_farbe` and search for `2010`. Select an object and it flies to this object as expected, but zooms in too much IMO.

Looking at code right now.

src/components/MarkerOverlayService.js
gjn: +1

gjn:
Last comment was for 3D only.

For 2D and 3D, the feature preview (when hovering over results in the dropdown menu) changed. It was standard marker before, now it's a blue circle filled with grey. I think it should be standard marker.

gjn:
A couple of things need adaption, otherwise, this is good stuff.

oterral:
jenkins test

oterral:
Ready to review/test

cedricmoullet:
@gjn could you please review and merge ?

gjn:
jenkins test

gjn:
This looks good, but the jenkins failure is indeed valid here. It's a regression. Compare prod to this. Open the link, press the x on the search box. On [prod](https://map.geo.admin.ch/?swisssearch=wasser&lang=de) and in [master](https://mf-geoadmin3.dev.bgdi.ch/?swisssearch=wasser&lang=de), the term disappears from the search box and from the permalink, on [this branch](https://mf-geoadmin3.dev.bgdi.ch/gberaudo_search_layer/?swisssearch=wasser&lang=de) it remains both in the url and in the search box.

oterral:
Tests are good!!  Fixed on the branch

gjn:
Perfect. That's fine work here.

PR #2609, opened 2015-08-27
[3D] 3d tilt

fredj:


fredj:
First look: https://mf-geoadmin3.dev.bgdi.ch/3d_tilt/?dev3d=true

fredj:
demo updated

fredj:
Thanks for any review.

I don't really like the new `disable3d` property for the swissimage layer. Any other proposal welcome !

src/components/tilt3d/Tilt3dDirective.js
```diff
@@ -0,0 +1,39 @@
...
+      templateUrl: 'components/tilt3d/partials/tilt3d.html',
+      link: function(scope, element, attrs) {
+
+        scope.supported = gaBrowserSniffer.webgl;
+
+        // FIXME: remove this
+        // true is the selected background layer is not 3d compatible.
+        scope.disabled = false;
+        scope.$on('gaBgChange', function(evt, value) {
+          scope.disabled = value.id == 'ch.swisstopo.swissimage';
```
elemoine: Why don't you use `value.disable3d` here?

src/index.mako.html
```diff
@@ -217,6 +217,7 @@
   % if device == 'mobile':
         <div ga-offline-bt></div>
   % endif
+        <div ga-tilt3d ng-if="globals.dev3d"></div>
```
elemoine: Don't you think that using `ng-show` would be a better choice here?

elemoine:
Question for everyone: shouldn't the mouse cursor change when hovering the 3D button and when 3D is disabled because the background layer is swissimage? Currently, the cursor is still a hand, which I think is confusing.

elemoine:
@fredj, when I switch to 3D I sometimes get some big jump with a blue background before actually seeing the expected background. Is this a known issue?

src/components/tilt3d/Tilt3dDirective.js
fredj: `disable3d` property is only available in the background selector directive, not here

elemoine:
> @fredj, when I switch to 3D I sometimes get some big jump with a blue background before actually seeing the expected background. Is this a known issue?

And when this occurs the map is not tilted after switching to 3D. I am not even sure the location is the same as the previous, 2D, location.

src/index.mako.html
fredj: The advantage of `ng-if` is that the directive is not initialized at all if the condition is false. (but I'm not opposed to change it)

fredj:
> @fredj, when I switch to 3D I sometimes get some big jump with a blue background before actually seeing the expected background. Is this a known issue?

yes, it's a known ol3-cesium issue

fredj:
> Question for everyone: shouldn't the mouse cursor change when hovering the 3D button and when 3D is disabled because the background layer is swissimage? Currently, the cursor is still a hand, which I think is confusing.

This is correct. The UI will by done by swisstopo.

src/components/tilt3d/Tilt3dDirective.js
oterral: it shouldn't, The `value`param should contains the `disable3d`property.

fredj:
I've fixed the css cursor when the feature is disabled or not available

src/components/tilt3d/Tilt3dDirective.js
fredj: oups, I was searching in the `evt` object ...

fredj: fixed, code updated

src/components/tilt3d/Tilt3dDirective.js
```diff
@@ -0,0 +1,35 @@
...
+      restrict: 'A',
+      templateUrl: 'components/tilt3d/partials/tilt3d.html',
+      link: function(scope, element, attrs) {
+
+        scope.supported = gaBrowserSniffer.webgl;
+
+        // true is the selected background layer is not 3d compatible.
+        scope.disabled = false;
+        scope.$on('gaBgChange', function(evt, value) {
+          scope.disabled = !!value.disable3d;
```
elemoine: Why `!!`? `value.disable3d` is a boolean value already, no?

src/index.mako.html
elemoine: Ok, It may be quite the same here. Personally I tend to use `ng-if` only when strictly necessary.

elemoine:
No more comments from me.

src/components/tilt3d/Tilt3dDirective.js
fredj: `boolean` or `undefned`

elemoine: Oh yes, true.

gjn:
LGTM. Artwork still remains to be done. But I'll merge this.

Thanks alot.

Issue #2610, opened 2015-08-27
[3D] Switch to 3D not possible with singleTile wms layers


gjn:
Testlink: `dev3d=true&lang=de&topic=aviation&bgLayer=ch.swisstopo.pixelkarte-grau&catalogNodes=1379,1381,1514&X=168670.41&Y=688454.08&zoom=6&layers=ch.bazl.luftfahrthindernis`

This is a singleTile WMS layer. Not that other WMS layer (e.g. ch.bfe.windenergieanlagen) don't suffer this bug.

@gberaudo I'm not sure, but I think this was introduce with https://github.com/geoadmin/mf-geoadmin3/pull/2590. Could you take a quick look? (But don't invest too much time as only one layer seems to be affected..)

The same behaviour is also in https://github.com/geoadmin/mf-geoadmin3/pull/2603

gjn:
The only difference I see in the configuration between the working wms and non-working wms is that the non-working wms is a single-tile wms. That's probably the cause.

[Here's a diff of the configurations. Left is working, right is causing this bug](https://www.diffchecker.com/kxomxqc6)

I could confirm that singleTile wms seems to be the cause of this behaviour. Other singleTile wms (e.g. ch.bazl.laermbelastungskataster-zivilflugplaetze_helikopter-maximalpegel) is causing the same issue. OP updated.

gberaudo:
I quickly checked Cesium support for singletile WMS but find nothing.
https://cesiumjs.org/Cesium/Build/Documentation/WebMapServiceImageryProvider.html

There is only a
https://cesiumjs.org/Cesium/Build/Documentation/SingleTileImageryProvider.html
which has nothing to do with WMS: a single image is loaded at creation of
the provider and never updated.

I guess the reason is that only tiled or small enough imagery are usable in
3D:
imagine you are close to the ground and tilted a lot, then Cesium would
have to request a high-precision single
image of the entire world. That would be huge and would not do.

Is it possible to have the 3D layer use plain WMS/WMTS?

gjn:
I think it should be possible to use the singleTile layers in tiled wms over mapproxy in 3D.

PR #2603, opened 2015-08-25
[3D] Add config3d property

oterral:
This PR is based on #2590. It adds a config3d property to the layer which contains what we nee to use the layer in 3d.

@gjn is it what you had in mind when we talk about client modification of the layersconfig?

PR #2611, opened 2015-08-28
[3D] Support singleTile WMS in 3d by switching to tiled wms

gjn:
This fixes https://github.com/geoadmin/mf-geoadmin3/issues/2610

SingleTile WMS are switched to tiled wms when going to 3D. It works for all singleTile WMS sources and thus we can now drape all our layers on 3D.

Note: the luftfahrthindernisse layers looks terrible like this because of the update date

image


gberaudo:
When we merge https://github.com/geoadmin/mf-geoadmin3/pull/2603 we should not need this anymore as the 3D configurations should never use singletiles.
+1 for merging this simple workaround in the meantime.

gjn:
Thanks for the review.

PR #2614, opened 2015-08-28
Some minor fixes

oterral:
[Test](https://mf-geoadmin3.dev.bgdi.ch/teo_export/)


fredj:
+1

src/components/popup/style/popup.less
```diff
@@ -108,10 +108,8 @@
     font-size: 12px;
 
     @media (max-width: @screen-tablet) {
-      position: absolute;
-      top: 36px;
-      bottom: 10px;
       width: 100%;
+      height: ~"calc(100% - 36px)";
```
gjn: What is this? less feature? or is this valid css accross all browsers?

oterral: ` ~"anything" ` is the syntax to avoid less to try to transform it, see http://lesscss.org/features/#features-overview-feature-escaping

PR #2624, opened 2015-09-08
Set the htmlmin version to v0.1.6

fredj:
version v0.1.7 is buggy ...

work in progress


oterral:
Thx for debugging this

PR #2629, opened 2015-09-09
[3D] Switch to 3d mode if the permalink contains lon and lat

### LINKED ISSUE #2628: [3D] Permalink and 3d mode
The 3d mode (versus the 2d mode) is not saved when reloading the page.

To reproduce:
 - load the page
 - switch to 3d
 - reload the page

-> the page loads in 2d instead of 3d.

Proposed solution: if the permalink contains `lon`, `lat` and `elevation` switch to 3d, otherwise stay in 2d.

fredj:
Fixes #2628

### DIFF
--- a/src/js/MainController.js
+++ b/src/js/MainController.js
@@ -152,6 +152,12 @@ goog.require('ga_storage_service');
           });
         }
       });
+      gaLayers.loadConfig().then(function() {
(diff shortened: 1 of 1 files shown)


fredj:
Thanks for any review

oterral:
lgtm

PR #2631, opened 2015-09-09
[3D] Update the tilt and rotate indicator

fredj:
Update the indicators as soon as the view changes.

Note: scope.watch can't be used because angular doesn't start a new digest cycle every time the camera is updated.


gberaudo:
LGTM

fredj:
Any other review ?

oterral:
Nice thx

PR #2644, opened 2015-09-15
Validate 3D permalink params

### LINKED ISSUE #2643: 3D permalink params allow bad value
Just saw, if we put bad value for `lon` and `lat` . The map stays black.

For `x` and `y` param  we use `isFinite` to test if the value is usable, may be we should do the same for 3d params

fredj:
Fixes #2643


oterral:
lgtm thx @fredj

Issue #2626, opened 2015-09-09
Controls 3D improvements


oterral:
- [x] Fix css. We can't drag the map at the bottom, the controls takes all the window width. 
- [x] Fix dynamic css when screen gets smaller. (oli)
- [x] UI (icons)

PR #2645, opened 2015-09-15
[3D] Reset the default tilt and rotation

fredj:
See #2626: "Click on indicator buttons should apply default values"

### DIFF
--- a/src/components/controls3d/Controls3dDirective.js
+++ b/src/components/controls3d/Controls3dDirective.js
@@ -48,11 +48,17 @@ goog.provide('ga_controls3d_directive');
 
         scope.tilt = function(angle) {
           angle = Cesium.Math.toRadians(angle);
-          var pivot = olcs.core.pickBottomPoint(scene);
-          if (pivot) {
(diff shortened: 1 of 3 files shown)


fredj:
Thanks for any review

src/components/controls3d/Controls3dDirective.js
```diff
@@ -63,6 +69,17 @@ goog.provide('ga_controls3d_directive');
...
 
+        scope.resetRotation = function() {
+          var angle = -camera.heading;
+          while (angle < -Math.PI) {
+            angle += Cesium.Math.TWO_PI;
+          }
+          while (angle > Math.PI) {
+            angle -= Cesium.Math.TWO_PI;
+          }
+          scope.rotate(Cesium.Math.toDegrees(angle));
```
oterral: Is it possible to use the exsting function https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/map/MapService.js#L1443  adding an ol3d param?

fredj: yes, we can create an angular service and put a function there

PR #2651, opened 2015-09-16
[3D] pegman mode

fredj:
demo: https://mf-geoadmin3.dev.bgdi.ch/pegman/?dev3d=true


fredj:
todo: update icons

fredj:
Thanks for any review

gjn:
Amazing. This is going to rule. I like the icons!

First thing: next to pegman, we need to have visbile help that shows keyboard bindings and how to use the pegman mode[s].

The mouse isn't used anymore, which I think takes aways from the exprience. It's not easy to 'look around' using keyboards...but would like to know input from others as well.

Keys to adapt 'height above ground' and 'speed' would be great as well (and show actual values next to icons).

gjn:
Code LGTM.

cedricmoullet:
[] Clicking on "f" activates the fly mode. Superpegman should also be visible after clicking on f
[] With shift, pegman moves/flies faster. Could it be more fast ? Or could we use ctrl-shift in order to have a "FA-18" speed ?
[] After clicking on pegman, please add a text box containing information on how to fly (keys etc...)
[] Rename SuperPegman.png to super_pegman.png
[] Mobile css

fredj:
updated todo list:
 - [ ] use mouse to look around in fps mode
 - [ ] left/right keys moves pegman instead of moving the camera

src/components/controls3d/fps.js
```diff
@@ -0,0 +1,431 @@
+goog.provide('fps');
+
+
+
+/**
+ * @param {!Cesium.Scene} scene
+ * @param {!Angular.Scope} scope
+ * @constructor
+ */
+function FPS(scene, scope) {
```
oterral: Shouldn't it be a service ? I'm pretty sure we will have the need to know which state is the FPS in other directives in the future.

fredj:
@gjn I've added the help div and updated the demo link: https://mf-geoadmin3.dev.bgdi.ch/pegman/?dev3d=true

See also #2762 to track down the improvements of the pegman mode

fredj:
Before merging, I need to fix the altitude in pegman mode.
See https://github.com/geoadmin/mf-geoadmin3/blob/master/src/js/GaCesium.js#L53

gjn:
Let ne know when you're done.

gjn:
oops. Ok, I'll review and merge.

gjn:
Very fine work. Thanks alot.

Issue #2675, opened 2015-09-17
3D quality loss


cedricmoullet:
https://mf-geoadmin3.dev.bgdi.ch/ltmoc/?dev3d=true&topic=ech&lang=fr&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.swissimage-product&lon=7.06684&lat=45.94749&elevation=3090&heading=360.000&pitch=-39.996

In 3D:
![image](https://cloud.githubusercontent.com/assets/319676/9932883/9b0f7572-5d47-11e5-9b6e-212f618fbc71.png)
in 2D:
![image](https://cloud.githubusercontent.com/assets/319676/9932898/b8c43364-5d47-11e5-8ffb-3b2bee5fa3ae.png)

How is it possible to explain the quality difference ?

gjn:
Is it similar on flat regions?

How is the view from top (and not at an angle as in the example above)?

gjn:
Also, requested LOD might make a difference. What happens if you zoom in?

cedricmoullet:
Yes, same issue in flat region and from top.
I tried to zoom in, without more success. 
Is there any setting with a kind of min resolution parameter in Cesium ?

gjn:
This might be related to the latest mapproxy adaptions to use the correct tilegrid. Maybe the tilegrid matching (from 4326 to lv03) is not optimal e.g. for level 12 it takes source 12 instead of source 21...or some such thing.

I try to see if the raw request is already with bad quality of if it's cesiums rendering.

gjn:
it's not mapproxy.

It's because it's using the pre-generated 4326 tiles that was done for testing by marcel. It's not complete and lacks higher zoom levels. Cesium is requestiong higher zoom levels but gets a 204 and therefore displays lower quality images from lower zoom levels.

Once we have full pyramid pre-generated, it should be good.

@ltclm When can we expect this for this layer? Or is it already done but we use the wrong address (we are currently using `00_to_delete_ch.swisstopo.swissimage-product` layer in 3d.

cedricmoullet:
Could we use mapproxy for this swissimage layer ?
I wouldn't generate the tiles since it will not be used in the first
release.

2015-09-18 8:46 GMT+02:00 Gilbert Jeiziner <notifications@github.com>:

> it's not mapproxy.
>
> It's because it's using the pre-generated 4326 tiles that was done for
[...]

PR #2678, opened 2015-09-18
Use official swissimage tiles

gjn:
This takes care of https://github.com/geoadmin/mf-geoadmin3/issues/2675 by using the official 3d tiles for swissimage product.


gjn:
https://github.com/geoadmin/mf-geoadmin3/issues/2675

Issue #2691, opened 2015-09-23
Draw annotation text not translated


cedricmoullet:
![image](https://cloud.githubusercontent.com/assets/319676/10038260/5c5b6630-61c3-11e5-80ac-200a4995f2c6.png)

gjn:
Like this in prod. I'll have a look.

gjn:
missing translations for annotation as well.

davidoesch:
draw_next_marker would fit

PR #2693, opened 2015-09-23
Fix drawing labels

gjn:
Fix for https://github.com/geoadmin/mf-geoadmin3/issues/2691

Trivial self merge.

### DIFF
--- a/src/components/draw/DrawDirective.js
+++ b/src/components/draw/DrawDirective.js
@@ -110,6 +110,9 @@ goog.require('ga_permalink');
       var updateHelpTooltip = function(overlay, type, drawStarted, onFirstPoint,
           onLastPoint) {
         var helpMsg = $translate.instant('draw_start_' + type);
+        if (type == 'marker' || type == 'annotation') {
+          helpMsg = $translate.instant('draw_next_' + type);
+        }
         if (drawStarted) {
           if (type != 'Point') {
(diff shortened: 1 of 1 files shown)


src/components/draw/DrawDirective.js
```diff
@@ -110,6 +110,9 @@ goog.require('ga_permalink');
       var updateHelpTooltip = function(overlay, type, drawStarted, onFirstPoint,
           onLastPoint) {
         var helpMsg = $translate.instant('draw_start_' + type);
+        if (type == 'marker' || type == 'annotation') {
+          helpMsg = $translate.instant('draw_next_' + type);
```
oterral: On the screenshot of Cedric the text not translated is 'draw_start_marker' . The draw next next is not used when drawing annotation and marker.

gjn: I know, but the translation is there for next and it makes sense.

Wasn't sure if I could just replace draw_next_marker with draw_start_marker in the translations.

oterral: When you add a point there is no next message,  the help message is always "Add a point". The translation exists but it's never used so we could remove it, I guess.

oterral: SO yes replacing draw_next_marker by draw_start_marker  should work

PR #2694, opened 2015-09-23
[3D] Adjust default extent

oterral:
This extent avoid loading too much tiles in 3D. This extent hqs been created with http://api3.geo.admin.ch/examples/geoadmin_rectangle.html


gjn:
For the Terrain, we will need to adapt the layer.json creation in the 3d-forge project. Usually, the layer.json is created based on the terrain files as present in the S3 bucket.

gjn:
New extent:
image

Old extent:
image

I think new extent is probably a bit too narrow.

@davidoesch @cedricmoullet Thoughts?

davidoesch:
+1 for default but when zooming out you need them? IMHO it does not make sense to look at 3d with a scale "old extent". if we can block cesium here we might improve user experience

gjn:
Not sure it's easy to have different extent for openlayers/2D and cesium/3D. @oterral ?

oterral:
I investigate

oterral:
After investigation the more logical solution should be to have the extent of a layer in the layersConfig.

With that we could override the extent for 3d if necessary and we could avoid loading a lot of empty tiles for layer like the ["Patrouillr des glaciers"](https://mf-geoadmin3.int.bgdi.ch/?topic=verteidigung&lang=fr&bgLayer=ch.swisstopo.pixelkarte-grau&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bafu.wrz-wildruhezonen_portal,ch.swisstopo.swisstlm3d-wanderwege,ch.vbs.patrouilledesglaciers-z_rennen&layers_visibility=false,false,false,false,true&layers_timestamp=18641231,,,,&catalogNodes=1579&layers_opacity=1,1,1,1,0.75&X=95587.67&Y=605805.55&zoom=3)

gjn:
Indeed, that would be a great solution!

It would be even better if it would be possible in 2D as well. It would eliminate a good part of the 204 requests (which is ~15-20% of total tile requests).

Maybe a demo with 1-2 layers, hardcoded client-side?

gjn:
@ltclm Would we have max-extent per layer easily available?

oterral:
Just see in the [wms GetCap](http://wms.geo.admin.ch/?request=getcapabilities&service=wms&version=1.3.0), we should have also good extent value there.

oterral:
@gjn I prepare a branch

gjn:
Talked with Marcel. No straight forward way to get max extent for each layer from bod currently.

But we can certainly add this info to the bod, set it to current extent per default and adapt them by hand for each layer.

oterral:
[Here is with a default extent](https://mf-geoadmin3.dev.bgdi.ch/teo_dfltextent/?topic=sachplan&lang=fr&bgLayer=voidLayer&dev3d=true&layers=ch.bfe.sachplan-geologie-tiefenlager,ch.vbs.patrouilledesglaciers-z_rennen&layers_opacity=0.75,0.75&X=189150.00&Y=643500.00&zoom=0&catalogNodes=419,420,421,426,427,430,1245,1654,424)  and [here without]( 
https://mf-geoadmin3.dev.bgdi.ch/?topic=sachplan&lang=fr&bgLayer=voidLayer&dev3d=true&layers=ch.bfe.sachplan-geologie-tiefenlager,ch.vbs.patrouilledesglaciers-z_rennen&layers_opacity=0.75,0.75&catalogNodes=419,420,421,426,427,430,1245,1654,424&X=189150.00&Y=643500.00&zoom=0)

The difference is significant in 3d

gjn:
I think this is very important for performance.

We'll do the needed changes in the bod. We'll do the following: if in the layersConfig, a layer does not have a `extent` defined, then we will use the default extent as defined in the Makefile. If there is a `extent` defined, then we will use the defined one.

I'll prepare a PR to prepare front-end for it.

Then we can gradually adapt the layers to have correct, minimal extent. This will be a simple new attribute in the BOD that can either be Null or have a defined extent.

gberaudo:
@oterral, we were not passing the `rectangle` on the Cesium imagery so I am wondering what
was the reason for the improvement you perceived in 3D.

The tests are now passing and we use rectangle for Cesium imageries.
Restricting to the Swiss rectangle does not seem to diminish the number of requested tiles.

An improvement would be to use EPSG:4326 coordinates for the 3D extents.

gjn:
in 2D, results are visible.

in 3D, there seems to be work left. @gberaudo , could you rebase on master (as 3d layer definitions change) and try to adapt default extent for 3D as you think it could be?

oterral:
@gberaudo we set the extent of ImageryLayer here ->  https://github.com/geoadmin/mf-geoadmin3/blob/master/ol3-cesium-plugin/garastersynchronizer.js#L53  so I guess it has the same effect if you set it on Provider.

gberaudo:
Thank you @oterral for pointing this to me.
I think the effect is slightly different and using the value from the 3D target layer
would make more sense.

TODO:
- [ ] use EPSG:4326 for 3D extent;
- [ ] rebase on master;
- [ ] stop setting the extent on the imagery layer.

oterral:
I'm updating the is PR .

oterral:
I don't think we should define the 3d extent in epsg:4326 (in the layersconfig I mean), because we use and compare projected extent everywhere else, and our data is in epsg:21781. I don't think we loose anything to convert the extent on the fly.

oterral:
@gjn this one is goot to review/merge I think

PR #2696, opened 2015-09-23
[3D] mouse position

fredj:


fredj:
demo: https://mf-geoadmin3.dev.bgdi.ch/mouseposition_3d/?dev3d=true

cedricmoullet:
When I come back in 2D, the mouse position tool is not activated

gjn:
I can confirm. The re-enabling of 2D Map coordinates is not working correctly.

src/components/mouseposition/MousePositionDirective.js
```diff
@@ -23,6 +25,45 @@ goog.provide('ga_mouseposition_directive');
...
         });
+
+
+        scope.$watch('::ol3d', function(ol3d) {
+          if (ol3d) {
+            var scene = ol3d.getCesiumScene();
+            var ellipsoid = scene.globe.ellipsoid;
+            var handler = new Cesium.ScreenSpaceEventHandler(scene.canvas);
+
+            handler.setInputAction(function(mouvement) {
```
gjn: This is called alot of times when moving the mouse. Maybe we should throttle this (we have a `gaThrottle` class for this) to 100ms or 200ms.

fredj:
Issues fixed, please test again

cedricmoullet:
Works fine.
@gjn: could you review and merge if ok ?

Issue #2356, opened 2015-05-13
[query tool] implement AND/OR operators in order to connect single queries


mariokeusen:
According to a discussion with @davidoesch  and the BAFU people responsible for gewiss and water topic, we have the requirement to connect several queries with other boolean operators than AND:

UC: open app, load a queriable layer, open the query tool, make two user-defined queries
Result: the user has no possibility to define how the queries are combined
Expected: user has a choice of different boolean operators (AND, OR, XOR ...) to combine his queries

loicgasser:
Technically not too hard.

davidoesch:
@cedricmoullet: FYI we need this function for the go live for the GEWISS migration

AFoletti:
It's planned to give the GEWISS ppl a link to the demo instance on the 19.06 (Friday next week). Do you have something against? Should we wait a couple of days more and deliver the link with this function already implemented?

cedricmoullet:
This will be without this function

2015-06-11 14:49 GMT+02:00 AFoletti <notifications@github.com>:

> It's planned to give the GEWISS ppl a link to the demo instance on the
> 19.06 (Friday next week). Do you have something against? Should we wait a
> couple of days more and deliver the link with this function already
> implemented?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/geoadmin/mf-geoadmin3/issues/2356#issuecomment-111121713>
> .
>



-- 
Twitter: http://twitter.com/cedricmoullet
Linked In: http://www.linkedin.com/in/cedricmoullet

<http://twitter.com/cedricmoullet>

cedricmoullet:
The proposal is to offer a global operator AND, OR, NOT and this will apply to all conditions.
@mariokeusen could you validate this with the customer ?

cedricmoullet:
@mariokeusen ?

mariokeusen:
Confirmed. We'll implement a global boolean operator.

cedricmoullet:
@loicgasser Could you please propose a mockup for this ?

loicgasser:
ok I will

loicgasser:
Keeping the same design we could per default, hide the option. (Nothing changes)
![query_1](https://cloud.githubusercontent.com/assets/3306154/8932885/1c12ce88-3540-11e5-80ca-373a49170e77.png)

Then when we have more than 1 filter, display a select box to select the operator:
[...]

PR #2723, opened 2015-09-29
Support OR in query tool

loicgasser:
[Test Link](https://mf-geoadmin3.dev.bgdi.ch/gal_and_or_query/?lang=fr&topic=aviation)

This is to address:
https://github.com/geoadmin/mf-geoadmin3/issues/2356

TODO: One translation

Relies on https://github.com/geoadmin/mf-chsdi3/pull/1688

### DIFF
--- a/src/components/query/QueryDirective.js
+++ b/src/components/query/QueryDirective.js
@@ -58,8 +58,9 @@ goog.require('ga_storage_service');
           list.push(paramsByLayer[filter.layer.bodId]);
         }
         var params = paramsByLayer[filter.layer.bodId].params;
+        var operator = $scope.selectedQueryOperator.value;
         // Where condition
-        var where = (params.where) ? params.where + ' and ' : '';
+        var where = (params.where) ? params.where + ' ' + operator + ' ' : '';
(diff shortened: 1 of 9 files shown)


gjn:
This looks fine to me. Need translations for `query_operator` in translation sheet.

cedricmoullet:
@oterral could you please review ?

src/components/query/QueryDirective.js
```diff
@@ -74,6 +75,16 @@ goog.require('ga_storage_service');
       $scope.updateOp(idx, filter);
     };
 
+    $scope.queryOperators = [{
+      label: 'AND',
+      value: 'and'
+    }, {
+      label: 'OR',
+      value: 'or'
```
oterral: Shouldn't we translate the labels?

oterral:
Missing translations otherwise lgtm.

src/components/query/QueryDirective.js
loicgasser: I think it's fine, what would you put instead of OR and AND?

oterral: "Et " et "Ou"

PR #2727, opened 2015-09-30
Faster rotate and tilt animation

fredj:
100 ms instead of 500ms


cedricmoullet:
For my understanding: why do we need this change ?

fredj:
To reduce the animation time for the tilt and rotate (3d controls).

Issue #2726, opened 2015-09-30
Switch directly to 3D if permalink is 3D


cedricmoullet:

oterral:
Fixed

Issue #2724, opened 2015-09-30
First 3D activation doesn't position the camera correctly


cedricmoullet:
all in the title

gjn:
I think I've found the reason for this. It's related to ol3-cesium initialisation, where the canvas size is not updated on initialisation.

gjn:
ol3-cesium patch is ready. See https://github.com/openlayers/ol3-cesium/pull/253

@gberaudo if you could have a look.

oterral:
Fixed [Test](https://mf-geoadmin3.dev.bgdi.ch/?lang=fr&topic=ech&dev3d=true&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bafu.wrz-wildruhezonen_portal,ch.swisstopo.swisstlm3d-wanderwege&layers_visibility=false,false,false,false&layers_timestamp=18641231,,,&lon=8.88086&lat=43.36545&elevation=389646&heading=360.000&pitch=-43.714)

PR #2729, opened 2015-09-30
Lazy loading and initialisation of cesium

gjn:
Contains a fix for https://github.com/geoadmin/mf-geoadmin3/issues/2726 
Contains a fix for https://github.com/geoadmin/mf-geoadmin3/issues/2724

Currently, cesium initialisation [1] is done once the 3d mode is activated by the user (clicking 3D button) or when application starts in 3d mode directly.

This PR adds  the initialisation process [1] after a given idle time [2]. This gets rid of the `First 2d->3d switch bug` described in the minutes document by assuring that the cesium viewer is ready when the user clicks the 3D button. Also, switching to 3D for the first time should be faster as the library and 3d view is already present.

As @gberaudo said, this could be extended to even load 3D terrain/tiles in the background. This is for a separate PR. Again, the hard part is the really define the 'idle' state [2].

[Testlink](https://mf-geoadmin3.dev.bgdi.ch/gjn_idle_load/?dev3d=true)

[1] cesium inistialisation is in fact 2 parts: loading of the script from the server and initialisation of the cesium view, which is synchronized with the ol3 map.
[2] Idle state is not easy to determine (you can find alot about it with google). We currently use angulars https interceptor to detect if all http/ajac requests are finished and wait 2 seconds to emit an idle event. Unfortunately, this check ignores the ol3 map tile requests.

Addendum: we don't use defer/async because we want to load cesium depending on browser capabilities.

Contains a fix for https://github.com/geoadmin/mf-geoadmin3/issues/2726 
Contains a fix for https://github.com/geoadmin/mf-geoadmin3/issues/2724


cedricmoullet:
Nice !
Two questions:
1. The initial 3D view is very zoomed. I expected to see the whole switzerland. 
2. We basically double the size of the library that will be loaded by every user (also for guys wanting only to play with 2D). I'd like to have a SWOT about that. @davidoesch ?

gjn:
Code is not very clean. I'll try an approach with promises.

1. Yes, that's bad. Will have a look.
2. Yes, 0.5MB is loaded by every use if he stays on the page longer than 5-10 seconds. All to speed up initial display. With a good connection, user gains 500ms for initial display, on a bad connection, it might rise to 2-3 seconds. We can also decide to not do it on mobile devices and only pre-load for desktop...but mobile (with worse connections) would gain the most by the pre-loading in the background.

gjn:
I've found the reason for 1) and, indeed, for https://github.com/geoadmin/mf-geoadmin3/issues/2724 . Seems to require a patch in ol3-cesium: https://github.com/openlayers/ol3-cesium/pull/253

davidoesch:
1; sth like this https://mf-geoadmin3.dev.bgdi.ch/gjn_idle_load/?dev3d=true&topic=ech&lang=en&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bafu.wrz-wildruhezonen_portal,ch.swisstopo.swisstlm3d-wanderwege&layers_visibility=false,false,false,false&layers_timestamp=18641231,,,&lon=8.06832&lat=45.10528&elevation=121853&heading=0.000&pitch=-33.166

2. Must under all conditions: keep 2d as fast at loading as it is right now (60% aller nutzer sind unter 10Mbit/s )
So my proposal: don't load 3d at all at start. load it
A) on click on 3d
and
B) after 5-10 sec of no requests made by the application ( user is not panning or using swisssearch)

doable?

gjn:
This PR does this, but we wait 2 seconds instead of 5-10.

gjn:
**Current behaviour (as in master)**:
good: 0.5MB loaded only when 3D is activated
good: users not using 3D don't load 0.5MB
good: 2D loading not impacted
bad: delay of first 3D view by the time loading will take (0.5 - 3 seconds, depending on connection speed)

** Behaviour of this PR**:
good: no delay because of library when activating 3D view after 5-10 seconds.
good: 2D loading is not impacted (for most users)
bad: some users (on very slow connections), _might_ have slower 2D view loaded
bad: after 5-10 seconds, everybody loads a library that might not be used. unneeded traffic.

cedricmoullet:
@oterral could you please review and merge if ok ?

cedricmoullet:
Please separate mobile (no load) from desktop (with preloading)

gjn:
@oterral I've adapted the PR a little bit, pre-loading after idle is now done only on desktop. Also working with promises now and cesium initailisation moved out of maincontroller.

gjn:
This PR now contains https://github.com/geoadmin/mf-geoadmin3/pull/2732 as well.

gjn:
I've also added a fix for https://github.com/geoadmin/mf-geoadmin3/issues/2726 as it's related to initial loading/setup of cesium.

@fredj @oterral If you could take a look.

oterral:
lgtm

gjn:
ok. thanks alot.

Issue #2731, opened 2015-10-01
ol3cesium.js never cached


fredj:
The file is never cached by the browser (or the proxies) because a `?_=xxx` suffix is added by jquery.
See https://api.jquery.com/jquery.getscript/ (Caching Responses)

PR #2735, opened 2015-10-03
Cache lazy loaded scripts

gjn:
This fixes https://github.com/geoadmin/mf-geoadmin3/issues/2731

Note: also the d3.js library was never cached (also lazy loaded). So this improves the caching for this file as well.


fredj:
+1

PR #2739, opened 2015-10-06
Save only visible layers

### LINKED ISSUE #2734: Offline mode downloads selected but not active layers
To reproduce:
- open map.geo.admin.ch with default settings on a mobile device (or desktop with mobile=true)
- select an area and download data

Result:
- not only pixelkarte, but the 4 selected (which are not visible on the map, since not active) layers are downloaded.
- Several pixelkarte tiles are missing, maybe because of limited memory (gobbled up by unwanted layers). To the user, it looks like the selected area is a white hole

oterral:
Fix #2734 

[Test](https://mf-geoadmin3.dev.bgdi.ch/teo_fix_2734/mobile.html?lang=fr&topic=ech&bgLayer=voidLayer&X=190000.00&Y=660000.00&dev3d=true&zoom=4&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bafu.wrz-wildruhezonen_portal,ch.swisstopo.swisstlm3d-wanderwege&layers_visibility=false,false,false,false&layers_timestamp=18641231,,,&mobile=true)

### DIFF
--- a/src/components/offline/OfflineService.js
+++ b/src/components/offline/OfflineService.js
@@ -95,10 +95,13 @@ goog.require('ga_styles_service');
 
 
       // Get cacheable layers of a map.
-      var getCacheableLayers = function(layers) {
+      var getCacheableLayers = function(layers, onlyVisible) {
         var cache = [];
         for (var i = 0, ii = layers.length; i < ii; i++) {
           var layer = layers[i];
+          if (onlyVisible && !layer.getVisible()) {
+            continue;
+          }
           if (layer instanceof ol.layer.Group) {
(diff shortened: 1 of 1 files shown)


gjn:
So an offline map would have layer in the permalink, the layer would appear in the layer-selector, but when activating the layer, nothing would be shown, right?

I'm ok with this, just to be sure I understand correctly.

oterral:
You can't active a layer when offline because the layer manager is hidden. But if you succeed yes nothing will be displayed.

Issue #2740, opened 2015-10-06
activate preselected layer via search


davidoesch:
all browsers

start app
 search for adresses "Gebäude- und Wohnungsregister"

-> layer is  loaded - but cant' be activated via all in one search

expected result:

if a layer is searched for and he is already in the availbale layer list: he can be activated via all in one search

gjn:
This affects all pre-selected layers.

The reason is probably that these layers are already added to the map, but invisible. So clicking them does nothing because usually a click adds a map with default visibility.

What should happen? Should we activate the layer on click? For all layers in the layer manager?

davidoesch:
expected result:

if a layer is searched for and he is already in the availbale layer list: he can be activated via all in one search

oterral:
Fixed by PR

PR #2743, opened 2015-10-07
Assure layer is made visible after search selection

gjn:
This is a fix for https://github.com/geoadmin/mf-geoadmin3/issues/2740


oterral:
The preview layer should be displayed to when we are hover it

oterral:
@gjn I'e updated the PR with to display the preview layer.

gjn:
Great.

Issue #2717, opened 2015-09-29
wms multi-url support


gjn:
We now have it and should therefore use it.

PR #2749, opened 2015-10-08
Add multi-domain support for wms layers in 3D

gjn:
Multi-domain support for wms requests in 3D.

Fix for https://github.com/geoadmin/mf-geoadmin3/issues/2717

Also, removes unneeded variable.

@cedricmoullet Shouldn't we have wms dynamically scaled before going public with this?

### DIFF
--- a/src/index.mako.html
+++ b/src/index.mako.html
@@ -751,8 +751,6 @@ itemscope itemtype="http://schema.org/WebApplication"
           
           gaLayersProvider.terrainTileUrlTemplate = '//3d.geo.admin.ch/1.0.0/{Layer}/default/{Time}/4326';
 
-          gaLayersProvider.wmsMapProxyUrl = '//wmts{s}.geo.admin.ch/mapproxy/service';
-          
           gaLayersProvider.layersConfigUrlTemplate =
               gaGlobalOptions.resourceUrl + 'layersConfig?lang={Lang}';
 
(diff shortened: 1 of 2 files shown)


src/index.mako.html
```diff
@@ -751,8 +751,6 @@
           
           gaLayersProvider.terrainTileUrlTemplate = '//3d.geo.admin.ch/1.0.0/{Layer}/default/{Time}/4326';
 
-          gaLayersProvider.wmsMapProxyUrl = '//wmts{s}.geo.admin.ch/mapproxy/service';
-
```
oterral: Why not using a `gaLayersProvider.wmsUrl` here and removing it from layersConfig ?

oterral:
We could use the template for 2d too, no?

src/index.mako.html
gjn: We could do that. But need to adapt 2D also. I see this in a next PR.

oterral:
Like you want for the common 2d/3d template otherwise lgtm

gjn:
Thanks for the review!

cedricmoullet:
@gjn: we'll have the same number of requets, now ? (but probably more concurrent request). I would observe it with the current infra.

gjn:
It's all about the peaks. Even second-only peaks. We certainly have more concurrent requests...in theory by factor 4-5.

PR #2754, opened 2015-10-08
Profile elevation differences (see 1912)

mariavkourouni:
[Demo](https://mf-geoadmin3.dev.bgdi.ch/kom_elevations/)


mariavkourouni:
de
en

cedricmoullet:
Very nice ! You could deploy a branch so that everyone can test it. Ask @oterral or @gjn how to do it.

cedricmoullet:
1. use this drawing: https://mf-geoadmin3.dev.bgdi.ch/ltmoc/?lang=de&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bafu.wrz-wildruhezonen_portal,ch.swisstopo.swisstlm3d-wanderwege,KML%7C%7Chttps:%2F%2Fpublic.dev.bgdi.ch%2FZc5HmyybRFOyU2FeUPJ1cw&layers_visibility=false,false,false,false,true&layers_timestamp=18641231,,,,&X=200940.05&Y=603834.54&zoom=9
2. Show one profile
3. Show the second profile:
image
-> difference should be visible and not NaN

gjn:
That's great stuff, Maria! Code is good.

Number formatting would be nice. I would keep meters, but add 1000 indicator: 20'000 m.

src/components/profile/ProfileService.js
```diff
@@ -104,6 +104,34 @@ goog.require('ga_urlutils_service');
...
+            var minElev = data[0].alts[elevationModel] || 0;
+            var diffel = maxElev - minElev;
+            return diffel;
+          }
+        };
+
+        //total positive elevation & total negative elevation
+        this.twoElevDiff = function(data) {
+          if (data.length != 0) {
+            var sum_down = 0;
```
oterral: Use camelCase variable name `sumDown` `sumUp`

src/components/profile/ProfileService.js
```diff
@@ -187,7 +215,15 @@ goog.require('ga_urlutils_service');
           var x = d3.scale.linear().range([0, width]);
           var y = d3.scale.linear().range([height, 0]);
 
+          //for the total elevation diff of the path
+          this.diff = this.elevDiff(data);
+          //for the total elevation up & total elevation down
+          var two_diff = this.twoElevDiff(data);
+          var diff_up = two_diff[0];
+          var diff_down = two_diff[1];
```
oterral: camelCase variables here too

src/components/profile/ProfileService.js
```diff
@@ -266,7 +302,35 @@ goog.require('ga_urlutils_service');
...
-           return element;
+          //Total Elevation Difference
+          group.append('text')
+              .attr('class', 'ga-profile-label ga-profile-elevation-difference')
+              .attr('x', 0)
+              .attr('y', height + options.margin.bottom - 2)
+              .style('text-anchor', 'left')
+              .attr('font-size', '0.9em')
+              .text('\u03A3 \u0394 ' + $translate.instant(options.yLabel) +
+                  ': ' + Math.round(this.diff * 100) / 100 + ' m');
```
oterral: To use the number formating suggested by @gjn, have a look to the `measure` filter [here](https://github.com/geoadmin/mf-geoadmin3/blob/kom_elevations/src/components/measure/MeasureService.js#L7)

oterral:
Very nice feature @mariavkourouni !!!

@cedricmoullet @gjn icons instead of text should be better (responsiveness, readibility), what do you think?

cedricmoullet:
I'm more for text. Icons are always hard to interpret.

2015-10-09 8:38 GMT+02:00 oterral <notifications@github.com>:

> Very nice feature @mariavkourouni <https://github.com/mariavkourouni> !!!
>
> @cedricmoullet <https://github.com/cedricmoullet> @gjn
> <https://github.com/gjn> icons instead of text should be better
> (responsiveness, readibility), what do you think?
>
> —
> <https://github.com/geoadmin/mf-geoadmin3/pull/2754#issuecomment-146770219>
> .
>



-- 

mariavkourouni:
jenkins test

src/components/profile/ProfileService.js
```diff
@@ -8,6 +8,48 @@ goog.require('ga_urlutils_service');
     'pascalprecht.translate'
   ]);
 
+  module.filter('indicator', function() {
```
oterral: I meant you can  use the filter which already exist : https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/measure/MeasureService.js#L7

src/components/profile/ProfileService.js
```diff
@@ -46,8 +88,8 @@ goog.require('ga_urlutils_service');
       };
     };
 
-    this.$get = function($q, $http, $timeout, $translate, $window, gaUrlUtils,
-                         gaBrowserSniffer, gaGlobalOptions) {
+    this.$get = function($q, $http, $timeout, $translate, indicatorFilter,
```
oterral: Use `measureFilter`here instead of 'indicatorFilter'

src/components/profile/ProfileService.js
```diff
@@ -266,7 +345,35 @@ goog.require('ga_urlutils_service');
...
-           return element;
+          //Total Elevation Difference
+          group.append('text')
+              .attr('class', 'ga-profile-label ga-profile-elevation-difference')
+              .attr('x', 0)
+              .attr('y', height + options.margin.bottom - 2)
+              .style('text-anchor', 'left')
+              .attr('font-size', '0.9em')
+              .text('\u03A3 \u0394 ' + $translate.instant(options.yLabel) +
+                  ': ' + indicatorFilter(Math.round(this.diff * 100) / 100));
```
oterral: Use `measureFilter(this.diff)`

src/components/profile/ProfileService.js
```diff
@@ -331,6 +468,20 @@ goog.require('ga_urlutils_service');
                   .tickSize(-width, 0, 0)
                   .tickFormat('')
               );
+          this.group.select('text.ga-profile-elevation-difference')
+            .transition().duration(transitionTime)
```
oterral: Do we really need the transition stuff?

src/components/profile/ProfileService.js
```diff
@@ -331,6 +468,20 @@ goog.require('ga_urlutils_service');
...
+          this.group.select('text.ga-profile-elevation-up')
+            .transition().duration(transitionTime)
+              .text('\u03A3 + ' + $translate.instant(options.yLabel) +
+                  ': ' + indicatorFilter(Math.round(this.twoDiff[0] * 100) /
+                  100));
+          this.group.select('text.ga-profile-elevation-down')
+            .transition().duration(transitionTime)
+              .text('\u03A3 - ' + $translate.instant(options.yLabel) +
+                  ': ' + indicatorFilter(Math.round(this.twoDiff[1] * 100) /
+                  100));
```
oterral: If we don't need the transition stuf I think you can create and `updateElevationLabels(diff, twoDiff)` to avoid duplicate code with [line 386](https://github.com/geoadmin/mf-geoadmin3/pull/2754/files#diff-bfd39f2f09138781ec89cd2819dbd0d7R386)

src/components/profile/ProfileService.js
mariavkourouni: no, we do not need it.

mariavkourouni:
jenkins test

src/components/measure/MeasureService.js
```diff
@@ -42,15 +42,54 @@ goog.provide('ga_measure_service');
         default: units = units || [' km', ' m'];
                  break;
       }
+
+      thousandSeparator = typeof thousandSeparator !== 'undefined' ?
```
oterral: this line is useless

src/components/measure/MeasureService.js
```diff
@@ -42,15 +42,54 @@ goog.provide('ga_measure_service');
...
+        thousandSeparator : false;
       // Having only one unit means we don't want to transform the measure.
       if (units.length == 1) {
-        return cleanAfterComma(measure) + units[0];
+        if (thousandSeparator == true) {
```
oterral: just test ` if (thousandSeparator)`

src/components/measure/MeasureService.js
```diff
@@ -42,15 +42,54 @@ goog.provide('ga_measure_service');
...
+                var str = '-' + km + '\'0' + m + units[0];
+              } else if (m >= 0) {
+                var str = '-' + km + '\'00' + m + units[0];
+              }
+              return str;
+            } else {
+              var str = '-' + measure + units[0];
+              return str;
+            }
+          } else {
```
oterral: I think you can avoid duplicate code, like this:

```
  var minus;
  if (measure < 0) {
      measure = Math.abs(measure);
      minus= '-';  
  }
  if (measure >= factor) {
    var km = parseInt(measure / factor);
    var m = Math.round(measure - km * factor);
    var sep;   // the separator
    if (m >= 100) {
      sep = '\'';
    } else if (m >= 10) {
      sep = '\'0';
    } else if (m >= 0) {
      sep ='\'00';
    }
    str = minus +  km + sep + m +  units[0];
    return str;
  }
} 
```

oterral:
I think there is 2 more thing to do:

- [ ] Fix the reisze of the chart when winow is resizing
- [ ] Avoid duplicate code in measure filter.
- [ ] Am I the only on ewho doesn't like these labels ?  why not `Height diff`,  `up`,  `down`  ?

src/components/measure/MeasureService.js
mariavkourouni: thank you

src/components/profile/ProfileService.js
```diff
@@ -253,8 +288,9 @@ goog.require('ga_urlutils_service');
           group.append('text')
               .attr('class', 'ga-profile-label ga-profile-label-x')
               .attr('x', width / 2)
-              .attr('y', height + options.margin.bottom - 2)
+              .attr('y', height - 4 + 3 * options.margin.bottom / 4)
```
oterral: This formula seems a bit magical , can't we simplify this?

src/components/profile/ProfileService.js
```diff
@@ -262,11 +298,47 @@ goog.require('ga_urlutils_service');
...
               .attr('dy', '1em')
+              .attr('font-size', '0.95em')
               .text($translate.instant(options.yLabel) + ' [m]');
 
-           return element;
+          //Total Elevation Difference
+          group.append('text')
+              .attr('class', 'ga-profile-label ga-profile-elevation-difference')
+              .attr('x', 0)
+              .attr('y', height + options.margin.bottom - 1)
```
oterral: Is the `- 1`necessary ?

src/components/profile/ProfileService.js
mariavkourouni: I did this so there is no overlapping between x-label and elevation labels when resizing the window.
pic1

src/components/profile/ProfileService.js
```diff
@@ -304,6 +383,8 @@ goog.require('ga_urlutils_service');
           }
           var x = d3.scale.linear().range([0, width]);
           var y = d3.scale.linear().range([height, 0]);
+          this.diff = this.elevDiff(data);
```
oterral: On window's resize dats is null, that'y why the profile is not resizing. Move the operation on the previous `else` block

src/components/profile/ProfileService.js
mariavkourouni: I know that it's a detail, but it gives one pixel space below the labels:

oterral: Ok makes sense.
Since this height is used 3 times create a variable with `height + options.margin.bottom - 1` and add a small comment explaining the reason of using `-1`

src/components/profile/ProfileService.js
```diff
@@ -253,8 +288,9 @@ goog.require('ga_urlutils_service');
           group.append('text')
               .attr('class', 'ga-profile-label ga-profile-label-x')
               .attr('x', width / 2)
-              .attr('y', height + options.margin.bottom - 2)
+              .attr('y', height + 2 * options.margin.bottom / 3)
```
oterral: The problem of this formula is if we change `options.margin.bottom` this couldn't work as expected . I prefer to put a magic number  `height + options.margin.bottom  - 30 `  or something like that.

src/components/profile/ProfileService.js
```diff
@@ -262,11 +298,50 @@ goog.require('ga_urlutils_service');
...
               .attr('y', 0 - options.margin.left)
-              .attr('x', 0 - height / 2 - 20)
+              .attr('x', 0 - height / 2 - 30)
               .attr('dy', '1em')
+              .attr('font-size', '0.95em')
               .text($translate.instant(options.yLabel) + ' [m]');
 
-           return element;
+          //For having one pixel space below the elevation labels
+          var elevLabelHeight = height + options.margin.bottom - 1;
```
oterral: Thi valus doesn't represent a height but the y coordinate of a position. Can you replace the name by `elevLabelY`

mariavkourouni:
jenkins test

oterral:
lgtm @mariavkourouni can you sqaash all the commiut in one pls

oterral:
Thank you very much for this @mariavkourouni

mariavkourouni:
Thank you also from my side for the guidance.

gjn:
Agree with @oterral about the labels.

@davidoesch Keep current state or add text (net, up, down)?

davidoesch:
Hi this is not yet ready: 

-> labels are not "human readable"

Follow this visual approach
http://www.outdooractive.com/en/hiking-trail/muensterland/b-h-2015-01-29-rund-um-stromberg/103380368/#dm=1 

icons:
unbenannt

position:
the icons below the profile

- [ ]  if only first line:
    - [x] show azimuth
    - [x] show distance
    - [ ]  highest/lowest point

- [ ] if only a line (more than one point):
    - [ ]  show total m up
    - [ ]  show total m down 
    - [ ] show distance
    - [ ]  highest/lowest point

- [ ] if a polyline
    - [ ] show total m up
    - [ ] show total m down
    - [ ] show distance
    - [ ]  highest/lowest point

As discussed in https://github.com/geoadmin/mf-geoadmin3/issues/1912

oterral:
You can see hightest/lowest point on the graph so  I think it's not needed to write the info to the user

davidoesch:
See but not read it: the user wants the number

oterral:
@davidoesch can we leave as it is but just changing the labels for now. We could modify later the display.

davidoesch:
ok with changing labels and adding icons: step after step

oterral:
@davidoesch  I've just seen something interessant, in most of trails website, they never displayed the total  elevation  can we reomove it ?

davidoesch:
show total m up and down is a must

cedricmoullet:
@davidoesch are the translations ready ?

davidoesch:
- which Translation ids?
- can we pls test before deploy --- there are Open points  see ticket

davidoesch:
I added the translation: no abbrevations -now I assuem the text is to long. I really do vote now for the icons. abbrevations are really bad, every hiking site uses icons wandern.ch , outdooractive. let us do it right from the beginning ...

gjn:
UserValue guys would vouch for text...I agree that icons make more sense here. Therefor, +1 for icons.

oterral:
ok go for icons also.

@mariavkourouni  we use Fontaweome icons to do that. Come to me if you have questions.

gjn:
@mariavkourouni Any updates?

mariavkourouni:
I am working now on this! I think we should update Fontawesome from 3.2.1 to 4.0.0.

oterral:
@mariavkourouni too much changes to update to 4.0.0 : 
   
   -  New prefix
   -  Needs to verify alignment
   -  Update the process to generate the FA file with our cutsom icons (@pfanguin)

If some icons is missing in 3.2.1 ask @pfanguin to add them.

mariavkourouni:
ok! I will ask when I am ready! thanks!

mariavkourouni:
Is it ok what I did with the library?
And do you like the icons?

gjn:
For the total, I would just use the up/down arrow. For down, the icon would need to be mirrored (otherwise, it's the same as the up).

src/components/profile/style/profile.less
```diff
@@ -1,3 +1,5 @@
+@import url("https://netdna.bootstrapcdn.com/font-awesome/4.4.0/css/font-awesome.min.css");
```
oterral: No need of this we already load font-awesome in [app.less](https://github.com/geoadmin/mf-geoadmin3/blob/master/src/style/app.less#L2)

oterral:
@mariavkourouni I've asked @pfanguin to have better icons in our fa build.

mariavkourouni:
ok @oterral

mariavkourouni:
jenkins test

mariavkourouni:
@oterral

src/components/profile/style/profile.less
```diff
@@ -23,6 +23,10 @@
     margin: 0;
   }
 
+  .ga-profile-icon {
+    font-family: FontAwesome;
```
oterral: can you put the `font-size` here too?

mariavkourouni: the icon for elevation differences needs to have different size from the icons of elev up and down.
do I put here font-size and use this class only for icons elev up and down?

oterral:
@mariavkourouni  can you rebase on master and add the `icons-reference.html` and `styles.css` files provided by @pfanguin in the zip file she sent yesterday  ?

mariavkourouni:
yes, I did it @oterral

src/components/profile/style/profile.less
oterral: Ok we can keep like that.

oterral:
Lgtm @davidoesch do you valid?

@mariavkourouni can you squash the 2 font-aweosme commits pls?

davidoesch:
Nicely done @mariavkourouni

@oterral  ok  for mergeing this 3 items

oterral:
Ok @davidoesch  I open an issue with improvements we could do

oterral:
No finally ther is always an issue #1912

mariavkourouni:
@oterral I squashed all the 3 commits instead of the last 2 (by mistake)
thank you @davidoesch

oterral:
@mariavkourouni .  No problem we can live with that. Thank you for the hard work on this.

gjn:
Fantastic!

davidoesch:
step by step: we have some fixed but there is more in
https://github.com/geoadmin/mf-geoadmin3/issues/1912

davidoesch:
@oterral all open issues and improvements are here #1912

PR #2760, opened 2015-10-12
[3D]: Make sure the layer config is loaded before creating the terrain provider

loicgasser:
Since https://github.com/geoadmin/mf-geoadmin3/pull/2729 (I think) 3D permalink was broken because we tried to access the layers before the layer configuration was actually loaded. This fixes this issue.

### DIFF
--- a/src/js/GaCesium.js
+++ b/src/js/GaCesium.js
@@ -110,9 +110,12 @@ var GaCesium = function(map, gaPermalink, gaLayers, gaGlobalOptions, $q) {
     }
   }();
 
-  cesiumLoaded.promise.then(function(activate) {
-    ol3d = initCesiumViewer(map, activate);
-    cesiumClients.resolve(ol3d);
+  // We need the Cesium lib and the Layers config to create
+  // the 3D viewer
+  $q.all([cesiumLoaded.promise, gaLayers.loadConfig()])
+      .then(function(resolutions) {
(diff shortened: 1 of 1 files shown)


src/js/GaCesium.js
```diff
@@ -110,9 +111,16 @@ var GaCesium = function(map, gaPermalink, gaLayers, gaGlobalOptions, $q) {
...
-  cesiumLoaded.promise.then(function(activate) {
-    ol3d = initCesiumViewer(map, activate);
-    cesiumClients.resolve(ol3d);
+  gaLayers.loadConfig().then(function() {
+    layersConfigLoaded.resolve(true);
+  });
+
+  // We need the Cesium lib and the Layers config to create
+  // the 3D viewer
+  $q.all([cesiumLoaded.promise, layersConfigLoaded.promise])
```
oterral: you can do `$q.all([cesiumLoaded.promise, gaLayers.loadConfig()])`

loicgasser: yup, already a promise that's right

loicgasser:
done

oterral:
Thx lgtm

gjn:
Thanks for the fix.

Still I think we can initialize part of cesium before having full layersconfig available...we probably need to separate things in the GaCesium initialisation. For now, we are on the safe side.

PR #2763, opened 2015-10-14
[3D] restrict the camera view

fredj:
Added a new `frustumFar` parameter to control the `far` attribute of the frustum (see https://cesiumjs.org/Cesium/Build/Documentation/PerspectiveFrustum.html)

Everything that is beyond this value is ignored.


gjn:
Is this meters?

fredj:
yes in meters

fredj:
And 500000000 is the default value

fredj:
If the value is low (eg. 200km) the "sky" is not visible anymore

PR #2769, opened 2015-10-16
Pseudo-hide layers behind opaque layers to avoid loading

gjn:
The idea is to not load tiles from layers which are hidden behind a layer which is fully opaque.

Depending on the benchmark, we  20-34% less tiles are loaded (12-31% less MB). See entry in the benchmark document.

**This is for 2D and 3D: increases map rendering performance, decreases number of requests**

We would need to define which layer would hide the others and apply such a parameters to the layersConfig.

Currently, the following layers are hiding fully all layers behind them:
```
'ch.swisstopo.swissimage-product',
'ch.swisstopo.pixelkarte-farbe',
'ch.swisstopo.pixelkarte-grau',
'ch.swisstopo.swisstlm3d-karte-farbe',
'ch.swisstopo.swisstlm3d-karte-grau',
'ch.swisstopo.pixelkarte-farbe-pk25.noscale',
'ch.swisstopo.pixelkarte-farbe-pk50.noscale',
'ch.swisstopo.pixelkarte-farbe-pk100.noscale',
'ch.swisstopo.pixelkarte-farbe-pk200.noscale',
'ch.swisstopo.pixelkarte-farbe-pk500.noscale',
'ch.swisstopo.pixelkarte-farbe-pk500.noscale',
'ch.swisstopo.pixelkarte-farbe-pk1000.noscale'
```

[Testlink](https://mf-geoadmin3.dev.bgdi.ch/gjn_hide/?dev3d=true&lang=de&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bafu.wrz-wildruhezonen_portal,ch.swisstopo.swisstlm3d-wanderwege,ch.swisstopo.swissimage-product&layers_visibility=false,false,false,false,true&layers_timestamp=18641231,,,,)

Toughts?

### DIFF
--- a/src/components/map/MapService.js
+++ b/src/components/map/MapService.js
@@ -82,14 +82,33 @@ goog.require('ga_urlutils_service');
         Object.defineProperties(olLayer, {
           visible: {
             get: function() {
-              return this.getVisible();
+              return this.userVisible;
             },
             set: function(val) {
+              this.userVisible = val;
+              var vis = this.userVisible && !this.hiddenByOther;
               // apply the value only if it has changed
               // otherwise the change:visible event is triggered when it's
-              // unseless
-              if (val != this.getVisible()) {
-                this.setVisible(val);
+              // useless
+              if (vis != this.getVisible()) {
+                this.setVisible(vis);
+              }
+            }
+          },
+          userVisible: {
+            writable: true,
+            value: true
+          },
+          hiddenByOther: {
+            get: function() {
+              return this.get('hiddenByOther');
+            },
+            set: function(val) {
+              this.set('hiddenByOther', val);
+              if (val && this.userVisible) {
+                this.setVisible(false);
+              } else {
+                this.visible = this.userVisible;
               }
             }
           },
@@ -912,6 +931,28 @@ goog.require('ga_urlutils_service');
                   response.data[id].config3d = id + '_3d';
                 }
               });
+
+              // Test layers opaque setting
+              ids = [
+                'ch.swisstopo.swissimage-product',
+                'ch.swisstopo.pixelkarte-farbe',
+                'ch.swisstopo.pixelkarte-grau',
+                'ch.swisstopo.swisstlm3d-karte-farbe',
+                'ch.swisstopo.swisstlm3d-karte-grau',
+                'ch.swisstopo.pixelkarte-farbe-pk25.noscale',
+                'ch.swisstopo.pixelkarte-farbe-pk50.noscale',
+                'ch.swisstopo.pixelkarte-farbe-pk100.noscale',
+                'ch.swisstopo.pixelkarte-farbe-pk200.noscale',
+                'ch.swisstopo.pixelkarte-farbe-pk500.noscale',
+                'ch.swisstopo.pixelkarte-farbe-pk500.noscale',
+                'ch.swisstopo.pixelkarte-farbe-pk1000.noscale'
+              ];
+              angular.forEach(ids, function(id) {
+                if (response.data[id]) {
+                  response.data[id].opaque = true;
+                }
+              });
+
               // Test extent define in config for tiled WMS an WMTS
               response.data['ch.bfe.sachplan-geologie-tiefenlager'].extent =
                   [635432, 182850, 708432, 291850];
@@ -2045,7 +2086,7 @@ goog.require('ga_urlutils_service');
                 }
               }
               if (angular.isDefined(layer)) {
-                layer.setVisible(visible);
+                layer.visible = visible;
                 // if there is no opacity defined in the permalink, we keep
                 // the default opacity of the layers
                 if (opacity) {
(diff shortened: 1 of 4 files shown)


gjn:
@oterral Ready for review.

@davidoesch @cedricmoullet Please give feedback.

@ltclm @AFoletti We need to define which layers have the capability to fully hide all below layers.

davidoesch:
ok from my side

fredj:
very nice improvement !

src/components/map/MapService.js
```diff
@@ -82,14 +82,34 @@ goog.require('ga_urlutils_service');
         Object.defineProperties(olLayer, {
           visible: {
             get: function() {
-              return this.getVisible();
+              return this.appvisible;
```
oterral: What does the `app` prefix means ?   application?

gjn: yes. basically, what the user says visiblity should be. could be renamed though.

src/components/map/MapService.js
```diff
@@ -2473,5 +2515,75 @@ goog.require('ga_urlutils_service');
...
+
+      return function(map) {
+        var scope = $rootScope.$new();
+        scope.layers = map.getLayers().getArray();
+        scope.f = function(l) {
+          if (gaLayerFilters.background(l) ||
+              gaLayerFilters.selected(l)) {
+            return true;
+          }
+          return false;
```
oterral: You can return  directly `return (gaLayerFilters.background(l) || gaLayerFilters.selected(l));`-,

src/components/map/MapService.js
```diff
@@ -2473,5 +2515,75 @@ goog.require('ga_urlutils_service');
...
+        };
+        scope.$watchCollection('layers | filter:f', function(l) {
+          layers = (l) ? l : [];
+          updateHiddenState();
```
oterral: Do we really need `layers`as global variable? can you just add  `l` to `updateHiddenState(l)` as parameter.

src/components/map/MapService.js
gjn: I'll adapt.

src/components/map/MapService.js
gjn: `updateHiddenState` is called above as well. `layers` is used inside of it. To always have the correct layers (those in layermanager + background layer), I'll use the module scope variable.

src/components/map/MapService.js
```diff
@@ -2473,5 +2515,75 @@ goog.require('ga_urlutils_service');
...
+        var hide = false;
+
+        unreg();
+
+        angular.forEach(sortedLayers, function(layer) {
+          layer.hiddenByOther = hide;
+
+          // Register all layers for changes
+          unregKeys.push(layer.on('change:visible', function(evt) {
+            $timeout(function() {
```
oterral: I suggest to use $rootScope.$applyAsync to avoid triggering a digest cycling if not necessary.

src/components/map/MapService.js
oterral: maybe `userVisible` also

src/components/map/MapService.js
gjn: You're probably right.

Thinking about it, it might even better to use gaDebounce to avoid too many calls to updateHiddenState. Thoughts?

oterral:
I've made minor comment otherwise looks good.

src/components/map/MapService.js
gjn: Especially for the opacity change below.

src/components/map/MapService.js
gjn: Better. I'll adapt.

src/components/map/MapService.js
oterral: yes debounce is life :)

src/components/map/MapService.js
```diff
@@ -82,14 +82,34 @@ goog.require('ga_urlutils_service');
...
+            writable: true,
+            value: true
+          },
+          hiddenByOther: {
+            get: function() {
+              return this.get('hiddenByOther');
+            },
+            set: function(val) {
+              this.set('hiddenByOther', val);
+              if (val == true && this.appvisible) {
```
gberaudo: `== true` is unnecessary.

gjn:
Adapted to @oterral and  @gberaudo's comments. Thanks alot for the reviews!

oterral:
There is a bad typo in commit comment.

gjn:
Fixed as well. I'll merge.

ltclm:
the layers in the list of this pr are not fully hiding all other layers:
some sample links with pk25.noscale
* https://map.geo.admin.ch/?topic=ech&lang=de&bgLayer=ch.swisstopo.pixelkarte-farbe&catalogNodes=457,458,477&layers=ch.swisstopo.vec200-miscellaneous-geodpoint,ch.swisstopo.pixelkarte-farbe-pk25.noscale
* https://map.geo.admin.ch/?topic=ech&lang=de&bgLayer=ch.swisstopo.pixelkarte-farbe&catalogNodes=457,458,485,491&layers=ch.swisstopo.geoidmodell-etrs89,ch.swisstopo.pixelkarte-farbe-pk25.noscale&layers_opacity=0.75,1

I think we don't have fully hiding layers at all.

oterral:
@ltclm It's not deployed on prod test with dev : https://mf-geoadmin3.dev.bgdi.ch/?topic=ech&lang=de&bgLayer=ch.swisstopo.pixelkarte-farbe&catalogNodes=457,458,485,491&layers=ch.swisstopo.geoidmodell-etrs89,ch.swisstopo.pixelkarte-farbe-pk25.noscale&layers_opacity=0.75,1

ltclm:
@oterral thanks for the link. 
The listed layers above are hiding  even the base layers without really overlapping them completely.
When adding one of these layers to the top of the layer selection, you might think that no other data exists outside the hiding layer perimeter, but thats not true.

gjn:
@ltclm is correct. Our background layers always have the biggest extent. So with this PR, they disappear when we add a layer configured as opaque. Depending on the opaque layer, this could look quite ugly as well. That's the drawback for this solution.

There are multiple solutions for this:
1) We apply this fix only in 3D. It was introduced for 3D to reduce produced traffic and costs. We could apply it only when we are in 3D. This is technically easy.
2) We could apply this visibility/hide feature only for certain zoom levels. So when whole Switzerland is visibile, we don't hide background layers (and others). This is technically easy.
3) We could try to define the exact extends (on a per zoom level and per tile level resolution) for each opaque layer and load background layers outside of these extends and hide them inside of them. Effort to have this information is quite heavy and technical implementation quite complicated
4) A mixture of the above.

@davidoesch @cedricmoullet @ltclm @oterral Thoughts?

oterral:
+1 for 1) 
Easier and doesn't affect 2d

loicgasser:
1) +1 for now.
2) Not sure about that. I d be tempted to say that seeing extra tiles at the border is equally important.
3) Here we go back to the layer metadata file telling us where we have data. This file is also used by Olivier to reduce traffic in 3d. We should discuss how to extend and generalize their usage in 3D first for raster layers and then in 2D.

davidoesch:
+1 for 1) in a first step. in a second step (if needed) a combination of 1)
+2)

loicgasser:
2) (when zoomed in)

ltclm:
thanks for the list of solutions @gjin 
:+1: for 1)

gjn:
PR with 1) is ready: https://github.com/geoadmin/mf-geoadmin3/pull/2797

PR #2775, opened 2015-10-20
Limit the maximum zoom height

fredj:
500000 is more or less equivalent to the 2d maximum resolution


gjn:
Good first step.

We also need to limit camera for lon/lat, I think.

PR #2779, opened 2015-10-21
[WIP] Add buildings from POC

gjn:
This is an attempt to add buildings from the POC to our current application. I've duplicated code and buildings from the POC. Unfortunately, it doesn't show any buildings/objects yet. There is not error in the console whatsoever. Buildings files are loaded.

@gberaudo Do you have an idea why it doesn't work? Probably the transformation matrix is wrong. I have no idea in which coordinates the gltf files are, so there's probably a mismatch there.

### DIFF
--- a/src/js/GaCesium.js
+++ b/src/js/GaCesium.js
@@ -24,6 +24,109 @@ var GaCesium = function(map, gaPermalink, gaLayers, gaGlobalOptions, $q) {
     return parseInt(params[name] || defaultValue, 10);
   };
 
+  // Building related code taken from original POC
+  // https://raw.githubusercontent.com/geoadmin/3d-testapp/198e5c3f5c5d22841d78183585874b4b5c85ddf4/poc.js?token=ACdAHY4w9qWFM6-g6ZISbNJWkhsLbEG7ks5WL2p0wA%3D%3D
+  //// This is a very simplified loading strategy for building tiles
+  var tileModelProvider = {
+     add: function(swCorner, neCorner, gltf) {
+        var rectangle = Cesium.Rectangle.fromCartographicArray([
+          Cesium.Cartographic.fromDegrees.apply(null, swCorner),
+          Cesium.Cartographic.fromDegrees.apply(null, neCorner)]
+        );
+        var center = Cesium.Rectangle.center(rectangle);
+        center.height = Math.max(swCorner[2], neCorner[2]) + 1400;
+        var top = Cesium.Ellipsoid.WGS84.cartographicToCartesian(center);
+        var tile = {
+          rectangle: rectangle,
+          southwest: Cesium.Cartesian3.fromDegrees.apply(null, swCorner),
+          top: top,
+          loaded: false,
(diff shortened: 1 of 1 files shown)


src/js/GaCesium.js
```diff
@@ -24,6 +24,109 @@ var GaCesium = function(map, gaPermalink, gaLayers, gaGlobalOptions, $q) {
...
+        var pointed = scene.globe.pick(ray, scene);
+        if (!Cesium.defined(pointed)) return;
+
+        var tile = [];
+        var bottom = Cesium.Ellipsoid.WGS84.cartesianToCartographic(pointed);
+        for (var i = 0; i < tiles.length; ++i) {
+          if (!Cesium.Rectangle.contains(tiles[i].rectangle, bottom)) continue;
+          tile.push(tiles[i]);
+        }
+        tile.forEach(function(tile) {
```
loicgasser: probably a little confusing to use tile forEach tile

src/js/GaCesium.js
```diff
@@ -24,6 +24,109 @@ var GaCesium = function(map, gaPermalink, gaLayers, gaGlobalOptions, $q) {
...
+  // https://raw.githubusercontent.com/geoadmin/3d-testapp/198e5c3f5c5d22841d78183585874b4b5c85ddf4/poc.js?token=ACdAHY4w9qWFM6-g6ZISbNJWkhsLbEG7ks5WL2p0wA%3D%3D
+  //// This is a very simplified loading strategy for building tiles
+  var tileModelProvider = {
+     add: function(swCorner, neCorner, gltf) {
+        var rectangle = Cesium.Rectangle.fromCartographicArray([
+          Cesium.Cartographic.fromDegrees.apply(null, swCorner),
+          Cesium.Cartographic.fromDegrees.apply(null, neCorner)]
+        );
+        var center = Cesium.Rectangle.center(rectangle);
+        center.height = Math.max(swCorner[2], neCorner[2]) + 1400;
```
loicgasser: 1400? magic number

gberaudo:
Indeed, there is an issue related to the transform matrix: the buliding tiles are upside down.
We add a similar issue once, which we fixed by converting again the collada to gltfs.
I will compile an updated converter and try out.

src/js/GaCesium.js
loicgasser: by the way this is dead code.

loicgasser:
buildings

Adding 25meters to the height does help, it seems that the initial extent are wrong,.

gberaudo:
eastnorth25
`Corrected` with 25m: the buildings are upside down and the tile is not at the right place; a mirror transformation along the left axis would make the tile `fit`.

northdown0
Using a `Cesium.Transforms.northEastDownToFixedFrame` transform, the buildings are oriented up but the tile is not at the right place; a -90 degrees rotation around the red point would make the tile `fit`.

loicgasser:
Lol it's a mess, is it a gLTF version issue? Why would the same files work in POC but not using the current geoadmin viewer? Sure we can do the transforms for demo purposes.

Issue #2733, opened 2015-10-02
KML import: issues with coordinates well outside extent


procrastinatio:
Look at:

[Good KML](https://s.geo.admin.ch/6752d7872b)

[Wroong KML](https://s.geo.admin.ch/6752d8789d)

```
<coordinates>
	16.7852253163771,0.00176540157163452,0 16.7851783752391,0.00177218988870834,0 
</coordinates> 

```

oterral:
If the placemark (well outside extent) is alone in the KML, the map is well centered on it.
When there is 2 placemarks very far away one from each other the map can't zoom out enough to show the 2 features because we have constrained the resoulton of the view .

So I don't think it's a bug.

oterral:
Or may be we can improve that and at least, intersects the features's extent with the projection extent to avoid the map zooms in a no man's land.

Jenselme:
FYI, the patch from #2741 fixed the problem in dev but not in prod (with build.js). From what I understand, the versions of OpenLayers are not exactly the same, and in dev, ol refuses to fit to an empty extent, whereas in prod it changes the center anyway.

To completely get rid of the problem, I had to add a method to `gaMapUtils`:

```js
        isExtentEmpty: function(extent) {
          return extent[0] > extent[2] || extent[1] > extent[3];
        },
```

and modify `addKmlLayer`:

```js
if (options.zoomToExtent &&
                  !gaMapUtils.isExtentEmpty(layerExtent)) {
```

I can open a PR for that if you are interested (and have the bug with build.js too).

oterral:
I'm not fully understand what do you mean.   Do you mean `olMap.getView().fit([])` doesn't work or  `olMap.getView().fit([200000,600000,200000, 600000])` doesn't work ?

Jenselme:
I mean that on some version of ol, `olMap.getView().fit([200000,600000,190000, 590000])` (which is a zoom on an empty extent) will zoom somewhere and that somewhere can be outside the portal, thus not completely solving the issue.

Is this clearer?

oterral:
Since an extent is defined by `[minx, miny, maxx, maxy]` , your extent is not valid,so I don't think we can consider that as a bug in mf-geoadmin3.  The extent is calculate by the intersection of the `gaGlobalOptions.defaultExtent`and `source.getExtent()`  , what is the value for these extents?

Jenselme:
[...]

PR #2780, opened 2015-10-21
gaMaputils.intersectWithDefaultExtent must not return an invalid extent

Jenselme:
Return undefined if extent is empty. This prevents OpenLayers to try to
zoom on an empty extent.

See
https://github.com/geoadmin/mf-geoadmin3/issues/2733#issuecomment-148775457
and below for details

### DIFF
--- a/src/components/map/MapService.js
+++ b/src/components/map/MapService.js
@@ -1372,6 +1372,10 @@ goog.require('ga_urlutils_service');
   module.provider('gaMapUtils', function() {
     this.$get = function($window, gaGlobalOptions, gaUrlUtils, $q) {
       var resolutions = gaGlobalOptions.resolutions;
+      var isExtentEmpty = function(extent) {
+        return extent[0] >= extent[2] || extent[1] >= extent[3];
+      };
+
       return {
         Z_PREVIEW_LAYER: 1000,
         Z_PREVIEW_FEATURE: 1100,
@@ -1667,12 +1671,17 @@ goog.require('ga_urlutils_service');
           if (!extent || extent.length !== 4) {
             return gaGlobalOptions.defaultExtent;
           }
-          return [
+          var extent = [
             Math.max(extent[0], gaGlobalOptions.defaultExtent[0]),
             Math.max(extent[1], gaGlobalOptions.defaultExtent[1]),
             Math.min(extent[2], gaGlobalOptions.defaultExtent[2]),
             Math.min(extent[3], gaGlobalOptions.defaultExtent[3])
           ];
+          if (!isExtentEmpty(extent)) {
+            return extent;
+          } else {
+            return undefined;
+          }
         },
 
         getFeatureOverlay: function(features, style) {


src/components/map/MapService.js
```diff
@@ -1372,6 +1372,10 @@ goog.require('ga_urlutils_service');
   module.provider('gaMapUtils', function() {
     this.$get = function($window, gaGlobalOptions, gaUrlUtils, $q) {
       var resolutions = gaGlobalOptions.resolutions;
+      var isExtentEmpty = function(extent) {
+        return extent[0] > extent[2] || extent[1] > extent[3];
```
gjn: `<=` and `<=` maybe?

src/components/map/MapService.js
```diff
@@ -1667,14 +1671,19 @@ goog.require('ga_urlutils_service');
...
-          return [
+          var extent = [
             Math.max(extent[0], gaGlobalOptions.defaultExtent[0]),
             Math.max(extent[1], gaGlobalOptions.defaultExtent[1]),
             Math.min(extent[2], gaGlobalOptions.defaultExtent[2]),
             Math.min(extent[3], gaGlobalOptions.defaultExtent[3])
           ];
+          if (!isExtentEmpty(extent)) {
+            return extent;
+          }
```
gjn: We should explicitely return something if extent is empty.

src/components/map/MapService.js
```diff
@@ -1667,14 +1671,19 @@ goog.require('ga_urlutils_service');
...
         },
 
+        isExtentEmpty: isExtentEmpty,
```
gjn: Is this ever used? If not, we should remove it.

src/components/map/MapService.js
Jenselme: Should I explicitly return `undefined`?

Jenselme:
PR updated:
- Use `>=`
- Remove gaMaputils.isExtentEmpty (not used and may be not useful)

src/components/map/MapService.js
gjn: This could result in layers getting extent 'undefined'. If this is ok, then yes. Otherwise, return defaultExtent.

oterral: if there is no intersection with the defaultExtent boundaries I think that makes sense that we return `undefined`.

Jenselme: From the first discussion with @oterral in #2733, it seems like the most logical thing to do. It works since ol won't try to fit with an `undefined` extent.

If it returns the default extent it might be confusing since the user may not understand why the extent changed. Furthermore, it is not the correct extent for the KML. So I think that `undefined` is a good compromise.

What would be really great is to inform the user that the imported KML is outside the geoportal (somewhere below `Parsing OK`). But that is outside the scope of this PR.

PR update to explicitly return undefined.

gjn:
Thanks alot. This looks good now.

PR #2781, opened 2015-10-21
[3D] Fix double map click registration

loicgasser:
I noticed there was a racing condition in the Tooltip directive between cesium initialisation and `isActive` controlling the sent of the identify requests.

This race becomes really obvious when using the non-mimified of version.
See: https://github.com/geoadmin/mf-geoadmin3/pull/2778
But I could observe this issue in prod as well.

To reproduce the error,
1. Open the app with a queryable layer (in dev mode) baseURL + ?dev3d=true&lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.pixelkarte-pk25.metadata&layers_opacity=0.95&X=191502.45&Y=769146.75&zoom=8
2. Switch to 3d
3. Click on an object

@fredj Does it seem like an acceptable fix to you?

### DIFF
--- a/src/components/tooltip/TooltipDirective.js
+++ b/src/components/tooltip/TooltipDirective.js
@@ -118,7 +118,11 @@ goog.require('ga_topic_service');
                 updateCursorStyle, 10, false, false);
 
         // Register click/touch/mousemove events on map
+        var deregMapEvents = angular.noop;
         var registerMapEvents = function(scope, onClick) {
+          if (deregMapEvents !== angular.noop) {
+            return;
+          }
           var map = scope.map;
           var onMapClick = function(evt) {
             var coordinate = (evt.originalEvent) ?
@@ -144,14 +148,19 @@ goog.require('ga_topic_service');
               updateCursorStyleDebounced(map, evt.pixel);
             });
           }
-          return function() {
+          deregMapEvents = function() {
             deregMapClick();
             ol.Observable.unByKey(deregPointerMove);
+            deregMapEvents = angular.noop;
           };
         };
 
         // Register leftclick event on globe
+        var deregGlobeEvents = angular.noop;
         var registerGlobeEvents = function(scope, onClick) {
+          if (deregGlobeEvents != angular.noop) {
+            return;
+          }
           var scene = scope.ol3d.getCesiumScene();
           var handler = new Cesium.ScreenSpaceEventHandler(scene.canvas);
           handler.setInputAction(function(evt) {
@@ -168,10 +177,12 @@ goog.require('ga_topic_service');
               onClick(coordinate, evt.position);
             });
           }, Cesium.ScreenSpaceEventType.LEFT_CLICK);
-          return function() {
+          deregGlobeEvents = function() {
             if (!handler.isDestroyed()) {
-              handler.removeInputAction(Cesium.ScreenSpaceEventType.LEFT_CLICK);
+              handler.removeInputAction(
+                  Cesium.ScreenSpaceEventType.LEFT_CLICK);
               handler.destroy();
+              deregGlobeEvents = angular.noop;
             }
           };
         };
@@ -193,8 +204,6 @@ goog.require('ga_topic_service');
                 canceler,
                 vector,
                 vectorSource,
-                deregMapEvents = angular.noop,
-                deregGlobeEvents = angular.noop,
                 listenerKey,
                 parser = new ol.format.GeoJSON();
 
@@ -263,12 +272,10 @@ goog.require('ga_topic_service');
                   if (scope.isActive) {
                     if (enabled) {
                       deregMapEvents();
-                      deregGlobeEvents = registerGlobeEvents(scope,
-                          findFeatures);
+                      registerGlobeEvents(scope, findFeatures);
                     } else {
                       deregGlobeEvents();
-                      deregMapEvents = registerMapEvents(scope, findFeatures,
-                          gaBrowserSniffer.mobile);
+                      registerMapEvents(scope, findFeatures);
                     }
                   }
                 });
@@ -278,10 +285,9 @@ goog.require('ga_topic_service');
             scope.$watch('isActive', function(active) {
               if (active) {
                 if (is3dActive()) {
-                  deregGlobeEvents = registerGlobeEvents(scope, findFeatures);
+                  registerGlobeEvents(scope, findFeatures);
                 } else {
-                  deregMapEvents = registerMapEvents(scope, findFeatures,
-                      gaBrowserSniffer.mobile);
+                  registerMapEvents(scope, findFeatures);
                 }
               } else {
                 // Remove the highlighted feature when we deactivate the tooltip


gjn:
Why not keep state of registration inside the registerMapEvent function? Simple flag indicating if already registered, flag is reset und dereg call and multiple register calls would simply bail out if flag is already set.

It would be robust against race-conditions and client code (users of registerMapEven and the deref function) don't need to know about and check for noop.

oterral:
Sorry I can't reproduce the bug.

loicgasser:
you should be able to now in debug mode. it happens every time

loicgasser:
@gjn the problem is that deregMapEvents is overwritten.
It goes first through
https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/tooltip/TooltipDirective.js#L283
because 3d is not active but there is a queryable layer and then
https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/tooltip/TooltipDirective.js#L270
when finally cesium is loaded.

oterral:
AH yes my breakpoints made it works.

loicgasser:
exactly ;)

src/components/tooltip/TooltipDirective.js
```diff
@@ -262,13 +263,15 @@ goog.require('ga_topic_service');
...
                     if (enabled) {
-                      deregMapEvents();
+                      deregMapEvents = deregMapEvents();
                       deregGlobeEvents = registerGlobeEvents(scope,
                           findFeatures);
                     } else {
                       deregGlobeEvents();
-                      deregMapEvents = registerMapEvents(scope, findFeatures,
-                          gaBrowserSniffer.mobile);
+                      if (deregMapEvents === angular.noop) {
```
oterral: I think we should be put this test at the begining of  `registerMapEvents`  and doiing the same in `registerGlobeEvents`

gjn:
That was not my point. May point was to keep the registration state inside of the registerMapEvents functions. All checks can then be made inside of registerMapEvents function and the function returned by it. That way, _users_ of those functions can call it as many times as they want it, in any order that they want it and we assure that at any time, only one event is registered.

As it's currently done, responsibility to check state of registrations is passed to the users of those functions.

src/components/tooltip/TooltipDirective.js
loicgasser: you mean pass the callback in registerMapEvents and if !== to angular.noop return the function passed initially. yeah i think it's nice. I can adapt.

gjn: That's basically my point.

oterral: Yes @loicgasser

src/components/tooltip/TooltipDirective.js
```diff
@@ -278,16 +286,17 @@ goog.require('ga_topic_service');
             scope.$watch('isActive', function(active) {
               if (active) {
                 if (is3dActive()) {
-                  deregGlobeEvents = registerGlobeEvents(scope, findFeatures);
+                  deregGlobeEvents = registerGlobeEvents(scope, findFeatures,
+                      deregGlobeEvents);
                 } else {
                   deregMapEvents = registerMapEvents(scope, findFeatures,
-                      gaBrowserSniffer.mobile);
```
loicgasser: third parameter gaBrowserSniffer.mobile was not used

src/components/tooltip/TooltipDirective.js
```diff
@@ -118,7 +118,10 @@ goog.require('ga_topic_service');
                 updateCursorStyle, 10, false, false);
 
         // Register click/touch/mousemove events on map
-        var registerMapEvents = function(scope, onClick) {
+        var registerMapEvents = function(scope, onClick, deregEvent) {
```
gjn: Why not keep state in function scope instead of passing it every time?

loicgasser:
wait a second still not good

src/components/tooltip/TooltipDirective.js
gjn: Something like

```
var registerMapEvents = function() {
    var dereg = angular.noop;
    return function(scope, onClick) {
        if (dereg != angular.noop) {
            return dereg;
        }
        .
        .
        .
        dereg = function() {
            deregMapClick();
            ol.Observable.unByKey(deregPointerMove);
            dereg = angular.noop;
        }
        return dereg;
    };
}();
```

That way, you don't need to track results of dereg calls.
Note: code above not tested.

loicgasser: But it's 2 different calls. I tried anyway and we end up with the same issue. Without an instance or variable outside the function, it won't work.
If you call:
https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/tooltip/TooltipDirective.js#L283
and then
https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/tooltip/TooltipDirective.js#L270
dereg will always be equal to `angular.noop`.
I understant it's better to be self contained though.

loicgasser:
This seems to work. please test again and let me know

src/components/tooltip/TooltipDirective.js
gjn: Hmm, `dereg = angular.noop` is only set if somebody actually calls the `deregMapEvents` function (and not when calling the `registerMapEvents`.

Anyway, we could use a module-wide variably by moving `var deregMapEvents = angular.noop` out of link function and into module function and then use this variable always. We always set `deregMapEvents` either inside the register functions or inside the deregister function.

```
var registerMapEvents = function(scope, onClick) {
        if (deregMapEvents != angular.noop) {
            return;
        }
        .
        .
        .
        deregMapEvents = function() {
            deregMapClick();
            ol.Observable.unByKey(deregPointerMove);
            deregMapEvents = angular.noop;
        }
        // Nothing to return, as deregMapEvents is outside of this function scope.
    };
```

Working?

src/components/tooltip/TooltipDirective.js
```diff
@@ -118,7 +118,11 @@ goog.require('ga_topic_service');
                 updateCursorStyle, 10, false, false);
 
         // Register click/touch/mousemove events on map
+        var deregMapListener = angular.noop;
         var registerMapEvents = function(scope, onClick) {
+          if (deregMapListener !== angular.noop) {
+            return deregMapListener;
```
gjn: No need to return it, as it's upper scope variable already.

src/components/tooltip/TooltipDirective.js
```diff
@@ -144,14 +148,20 @@ goog.require('ga_topic_service');
               updateCursorStyleDebounced(map, evt.pixel);
             });
           }
-          return function() {
+          deregMapListener = function() {
             deregMapClick();
             ol.Observable.unByKey(deregPointerMove);
+            deregMapListener = angular.noop;
           };
+          return deregMapListener;
```
gjn: dito.

src/components/tooltip/TooltipDirective.js
```diff
@@ -144,14 +148,20 @@ goog.require('ga_topic_service');
...
         };
 
         // Register leftclick event on globe
+        var deregGlobeListener = angular.noop;
         var registerGlobeEvents = function(scope, onClick) {
+          if (deregGlobeListener != angular.noop) {
+            return deregGlobeListener;
```
gjn: dito

src/components/tooltip/TooltipDirective.js
```diff
@@ -168,12 +178,15 @@ goog.require('ga_topic_service');
...
+          deregGlobeListener = function() {
             if (!handler.isDestroyed()) {
-              handler.removeInputAction(Cesium.ScreenSpaceEventType.LEFT_CLICK);
+              handler.removeInputAction(
+                  Cesium.ScreenSpaceEventType.LEFT_CLICK);
               handler.destroy();
+              deregGlobeListener = angular.noop;
             }
           };
+          return deregGlobeListener;
```
gjn: dito

src/components/tooltip/TooltipDirective.js
```diff
@@ -267,8 +280,7 @@ goog.require('ga_topic_service');
                           findFeatures);
                     } else {
                       deregGlobeEvents();
-                      deregMapEvents = registerMapEvents(scope, findFeatures,
-                          gaBrowserSniffer.mobile);
+                      deregMapEvents = registerMapEvents(scope, findFeatures);
```
gjn: No need to assign here, as it's done inside the registerMapEvents function

src/components/tooltip/TooltipDirective.js
```diff
@@ -280,8 +292,7 @@ goog.require('ga_topic_service');
                 if (is3dActive()) {
                   deregGlobeEvents = registerGlobeEvents(scope, findFeatures);
                 } else {
-                  deregMapEvents = registerMapEvents(scope, findFeatures,
-                      gaBrowserSniffer.mobile);
+                  deregMapEvents = registerMapEvents(scope, findFeatures);
```
gjn: dito

loicgasser:
no you need 2 variables, try https://github.com/geoadmin/mf-geoadmin3/compare/gal_gjin_popup_fix?expand=1 it doesn't work otherwise.

loicgasser:
sorry forget my comment

gjn:
Yes, one for each event (map and globe). Code @ your link looks good to me.

loicgasser:
ok updated I think we got. ;)

gjn:
LGTM. Thanks alot for your patience.

PR #2789, opened 2015-10-22
Fix linting error

loicgasser:
fix variable redefinition introduced in https://github.com/geoadmin/mf-geoadmin3/pull/2780/files


gjn:
My bad. Sorry for that and thanks for the quick fix.

PR #2797, opened 2015-10-27
Hide Layers behind opaque layers only in 3D

gjn:
This is a reaction to the discussion in https://github.com/geoadmin/mf-geoadmin3/pull/2769. The hiding behind opaqu layers is limited to the 3D view only.

[Testlink](https://mf-geoadmin3.dev.bgdi.ch/gjn_3dhide/?dev3d=true)


oterral:
Lgtm thx @gjn nice improvment

Issue #2805, opened 2015-10-29
print does not work with WMS layers


AFoletti:
the print does not work anymore with WMS layers. baseURL is missing in the request.
@gjn already identified a possible fix

oterral:
Fixed

PR #2767, opened 2015-10-15
Use multi-urls for tiled WMS and defines default template and subdoma…

oterral:
…ins in one place

Fix #2717

This PR affects WMTS and WMS. For these layers the url and subdomains used are those defined in the `layersConfig `  if those properties are not defined in the layersConfig that uses the default values defined in the `index.mako.html` 

[Test](https://mf-geoadmin3.dev.bgdi.ch/teo_tpls/?lang=fr&topic=luftbilder&bgLayer=ch.swisstopo.pixelkarte-grau&dev3d=true&layers=ch.swisstopo.swissimage-product,ch.swisstopo.swisstlm3d-wanderwege,ch.bazl.laermbelastungskataster-zivilflugplaetze_zweite-nachtstunde,ch.swisstopo.lubis-luftbilder_schraegaufnahmen&layers_opacity=1,1,0.75,1&catalogNodes=1179&layers_timestamp=,,,&lon=8.58313&lat=46.40191&elevation=106925&heading=0.357&pitch=-39.552)

PR #2807, opened 2015-10-30
Apply wms template on layer creation only

gjn:
This is a fix for https://github.com/geoadmin/mf-geoadmin3/issues/2805

Since we introduced multi-domain support for geoadmin WMS backends [1], all print requests which contained geoadmin wms layers did not work. This PR adresses this by still keeping multi-domain support for wms but using single-domain for printing.

[1] https://github.com/geoadmin/mf-geoadmin3/pull/2767

### DIFF
--- a/src/components/map/MapService.js
+++ b/src/components/map/MapService.js
@@ -821,7 +821,11 @@ goog.require('ga_urlutils_service');
         var layers;
 
         // Returns a unique WMS template url (e.g. //wms{s}.geo.admin.ch)
-        var getWmsTpl = function(tpl, wmsParams) {
+        var getWmsTpl = function(wmsUrl, wmsParams) {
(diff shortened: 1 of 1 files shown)


oterral:
Do we want to remove  the `wmsUrl` property from the LayersConfig? I thought yes.
If `yes` we should fix this differently otherwise we will fix the same bug in the future.

@gjn what do you think ?

gjn:
we need the wmsUrl property for testing purposes (to test dev wms, etc). I think it's used for data integration process as well.

oterral:
Yes sure  but the default one could be remove.

gjn:
Agree. But alot of stuff in layersConfig could be removed and apply default values.

oterral:
OK we will see when we refactor the layer config.

PR #2814, opened 2015-11-02
Use automatic render loop to save resources when idle

gberaudo:
The browser will stop rendering the globe after a few seconds of inactivity like no mouse action and no downloading of tiles.
Test link: https://mf-geoadmin3.dev.bgdi.ch/gberaudo_autorenderloop/?dev3d=true

PR #2803, opened 2015-10-28
Fix #2785 : Display layers depending on the lod

oterral:
Fix #2785
[Test](https://mf-geoadmin3.dev.bgdi.ch/teo_zoomtolod/?dev3d=true&lang=de&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bafu.wrz-wildruhezonen_portal,ch.swisstopo.swisstlm3d-wanderwege&layers_visibility=false,true,false,false&layers_timestamp=18641231,,,&lon=8.21201&lat=46.85379&elevation=500015&heading=0.105&pitch=-89.981)
Based on  #2787

PR #2816, opened 2015-11-02
Screenspace error correction

gberaudo:
Optionnal, need to be enabled with parameter `sseEnabled=true`.

Linearly decrease computed tile error of distant tiles according to 3 zones.
-  error * 1 when distance < mindist;
- linear(error) when mindist < distance < maxdist;
- error * maxerrorfactor 

Tweakable parameters are:
- `sseMindist`, the minimum distance before applying correction
  default:  5000
  units: meters
- `sseMaxdist`, the maximum distance for applying linear correction
  default: 10000
  units: meters
- `sseMaxerrorfactor`, the factor to apply after maxdist
  default: 0.25
  units: none
- `sseEnabled`, to enable this optimization
- `sseNoheight`, to disable using camera height to adjust mindist and maxdist
- `sseNopickglobe`, to disable using the distance to a point on the screen
- `ssePickposition`, the canvas height ratio for defining the position where to pick the globe
  default: 0.6666 (2 third of the screen starting from the top)
  units: none
- `sseMaxheight`: camera height starting from which to disable the optimization
  default: 0 (optimization always enabled)
  units: meters
- `sseMincamfactor`, a factor to apply to mindist when using picking
  default: 0.9
- `sseMaxcamfactor`, a factor to apply to maxdist when using picking
  default: 1.2

These distances are automatically `corrected` according to the camera height:
even when the  camera is above maxdistance, correct terrain will be shown.

Test link: https://mf-geoadmin3.dev.bgdi.ch/gberaudo_performance/?dev3d=true&sseEnabled=true&sseMindist=5000&sseMaxdist=10000

Camera factor is used like this:
```
if (!noheight) {
  min = Math.min(mindist, mincamfactor * cameraHeight);
  max = Math.max(maxdist, maxcamfactor * cameraHeight);
}
```


gberaudo:
This PR is on top of #2814.

gjn:
Do the cesium/ol3-cesium versions correspond with https://github.com/geoadmin/mf-geoadmin3/pull/2803? Does one break the other?

gjn:
I tested some time, comparing results to current master, and I have to tell I can hardly see any difference between the two. If the metrics are correct, this is a huge improvement.

gberaudo:
This PR uses the latest versions of Cesium and ol3-cesium.

gberaudo:
@gjn, there was an error in the initial URL (was missing the mindist and maxdist values).
Could you please try with the full URL? You will certainly see a difference.

gberaudo:
Cesium has improved their fog branch and are planning to merge it this week: https://github.com/AnalyticalGraphicsInc/cesium/pull/3154
They added some code to correct the computed screenspace error.
Would be interesting to compare.

gberaudo:
Updated the test link.

gjn:
I'll retry a fog demo today.

gjn:
Could you rebase/squash this?

gberaudo:
Link updated, ready for more testings.

oterral:
@gberaudo are you sure the units of `mindist`and `maxdist` are kilometers and not meters?

src/js/GaCesium.js
```diff
@@ -178,3 +181,93 @@ var GaCesium = function(map, gaPermalink, gaLayers, gaGlobalOptions, $q) {
...
+ */
+SSECorrector.prototype.newFrameState = function(frameState) {
+    this.cameraHeight = frameState.camera.positionCartographic.height;
+
+    if (this.pickglobe) {
+      var scene = frameState.camera._scene;
+      var canvas = scene.canvas;
+      var pixelHeight = this.pickposition * canvas.clientHeight;
+      var pixel = new Cesium.Cartesian2(canvas.clientWidth / 2, pixelHeight);
+      this.cameraHeight = undefined;
```
oterral: Is it wanted ? I think that's why the `maxheight` param doesn't work, because the `Math.max` operation further  always returns `NaN`

src/js/GaCesium.js
```diff
@@ -178,3 +181,93 @@ var GaCesium = function(map, gaPermalink, gaLayers, gaGlobalOptions, $q) {
...
+var SSECorrector = function() {
+  var params = {};
+  location.search.substr(1).split('&').reduce(function(previous, current) {
+    var splitted = current.split('=');
+    if (splitted.length === 2) {
+      params[splitted[0]] = splitted[1];
+    } else {
+      params[splitted[0]] = true;
+    }
+  });
```
oterral: Why not using the gaPermalink service instead of using the `location.search` object? 

`var params = gaPermalink.getParams()`

src/js/GaCesium.js
gberaudo: Indeed, I updated the PR a few minutes ago.

src/js/GaCesium.js
gberaudo: It was copied from the app viewer. All this code is temporary so I did not spent time using `gaPermalink`. Do you think I should anyway update the PR to use it?

gberaudo:
All parameters has been prefixed by `sse` and the test URL was updated.

src/js/GaCesium.js
gberaudo: Without this code, there is a big difference between the camera height (vertical) and the distance to the center of the screen (oblique). With an oblique view, all tiles could be blured.

src/js/GaCesium.js
oterral: Temporary for a long time :) . Yes I'll prefer to use the gaPermalink, You can pass the `params` object in the parameter of the constructor.

oterral: I've seen that with the `reduce` function the 1rst param is never set in the `current` value.

gberaudo: Done, thanks for the kind reminder.

src/js/GaCesium.js
oterral: Yes exactly what I saw, it wasn't testing the last version.

gberaudo:
Was rebased onto #2860.
If no other comment, can be merged once #2860 is merged.

oterral:
Thx @gberaudo it will merged

gberaudo:
Thanks for the review @oterral.

PR #2829, opened 2015-11-05
Change the minimumZoomDistance according to the camera height

### LINKED ISSUE #2668: collision distance: default mode
collision distance: default mode 

*default* views like this https://mf-chsdi3.dev.bgdi.ch/ltmoc/shorten/6702a2f1bf

do not make any sense

proposal:
an altitude dependent collision setting:

for camera positions of altitudes above 1800m : use the same settings for distance camera - terrain as in schweizmobil
for camera positions of altitudes below 1800m : use half the distance schweizmobil has currently for urban area as for eg city of Bern

fredj:
500m if the camera is above 1800m, 250m otherwise.

Fixes #2668 
Link: https://mf-geoadmin3.dev.bgdi.ch/minimumZoomDistance/?dev3d=true

### DIFF
--- a/src/js/GaCesium.js
+++ b/src/js/GaCesium.js
@@ -69,7 +69,6 @@ var GaCesium = function(map, gaPermalink, gaLayers, gaGlobalOptions, $q) {
     globe.maximumScreenSpaceError = maximumScreenSpaceError;
     var scene = cesiumViewer.getCesiumScene();
     scene.globe.depthTestAgainstTerrain = true;
-    scene.screenSpaceCameraController.minimumZoomDistance = 50;
     scene.screenSpaceCameraController.maximumZoomDistance = 500000;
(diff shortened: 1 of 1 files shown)


fredj:
please review

oterral:
I see one drawback with this. In the moutain we can't see level 18 of swissimage. May be decrease a bit the 500m fix this.

fredj:
what about 200m / 400m instead of 250m / 500m ?

oterral:
Good for me thx @fredj

fredj:
@oterral, @davidoesch The minimum distance is now 400m if the height is above 1800m and 200m otherwise. Please test: https://mf-geoadmin3.dev.bgdi.ch/minimumZoomDistance/?dev3d=true

davidoesch:
ok... lets go with this one

PR #2841, opened 2015-11-06
Remove 3d for production env

gjn:
We need to remove 3d for production environment on next deploy.

@oterral Could you quickly review/merge this?


oterral:
LGTM

PR #2845, opened 2015-11-06
Display some city names in 3d

oterral:
[Test](https://mf-geoadmin3.dev.bgdi.ch/teo_names/?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&dev3d=true&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bafu.wrz-wildruhezonen_portal,ch.swisstopo.swisstlm3d-wanderwege,ch.swisstopo.swissimage-product&layers_visibility=false,false,false,false,false&layers_timestamp=18641231,,,,&lon=7.87210&lat=47.46146&elevation=1747&heading=275.119&pitch=-15.349)


gjn:
Good stuff! I think principle is good. A few remarks:

- texts are not all visible. Height's are probably not correct.
- text size is ok for big cities. should probably depend on LOD for smaller villages
- why 2 datasets for each LOD? For testing?

oterral:
That's where I have seen a problem. Lod are not managed by billboards we can only play with distance. 
The others problems is we can't modified height and scale according to the distance. May be we could go through all the primitives eache time the distance change  but I'm pretty sure is not very efficient. I need to have a look to the QuadTreePrimitive as suggested by @gberaudo.

Currently we have 2 layers used to manage lod:

 - 1rst billboard includes 11 main cities and are displayed  when the distance is higher than 30km and with a height of 2000 m
 - 2nd billboard includes 100 random city names (incliuding the 11 main cities and are displayed when the distance is lower than 30 km and with a height of 800m

gberaudo:
@oterral, it is possible to modify scale by distance by passing the
`scaleByDistance` option on the Billboard.
There are also the `pixelOffset` and `pixelOffsetScaleByDistance`  options,
however it is broken (voluntarily)
in the C2C branch (so we can fix it if needed).

To have billboards appear only when close to the camera, you can use the
follewing C2C specific feature:
    var visibilityDistances = new Cesium.NearFarScalar(
          0, // disable position decimation
          0, // unused
          minDistance, // hide when distance is greater than this
          0); // unused
    bbOptions.pixelOffsetScaleByDistance = visibilityDistances;

davidoesch:
if a billboard is within a valey and then you zoom out : the text is hitting the valley walls and dissapear: eg
https://mf-geoadmin3.dev.bgdi.ch/teo_names/?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&dev3d=true&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bafu.wrz-wildruhezonen_portal,ch.swisstopo.swisstlm3d-wanderwege,ch.swisstopo.swissimage-product&layers_visibility=false,false,false,false,false&layers_timestamp=18641231,,,,&lon=7.81879&lat=47.36492&elevation=7720&heading=79.283&pitch=-57.662

expected results: the billboard height increases automatically when there is no space on the side
unbenannt

oterral:
@gberaudo I've seen the `scaleByDistance` property but doesn't exist for `[Label](https://cesiumjs.org/Cesium/Build/Documentation/Label.html)` . So I've tried to implemented it and see that a  `Label` is composed by glyphs (one for each letter) and then each glyph has a billboard. So I succeed to scale letter separately but not the entire word, There is probably somewhere somehting to measure the size of the text, but I didn't go further in my investigations.

For `pixelOffsetScaleByDistance` , the fact is broken can explain why I didn't see any effects using it :)

For visibility I use the `translucencyByDistance`. It does the work.

PR #2852, opened 2015-11-09
[WIP] Add drag & drop feature on layer selector

Jenselme:
Hi,

This my proposal to implement the reordering of layers with a drag and drop features. According to my tests, it works well but I didn't test it on mobile yet. All comments and suggestions are welcomed.

cc @tigerfoot

### DIFF
--- a/src/components/layermanager/LayermanagerDirective.js
+++ b/src/components/layermanager/LayermanagerDirective.js
@@ -127,8 +127,9 @@ goog.require('ga_urlutils_service');
       link: function(scope, element, attrs) {
         var map = scope.map;
         var drag = {
-          above: 'drag-enter-above',
-          under: 'drag-enter-under',
+          info: $('#drag-info'),
+          above: 'above',
+          under: 'under',
           scrollSpeed: 10,
           layerManager: $('[ga-layermanager]'),
           element: null,
@@ -242,6 +243,13 @@ goog.require('ga_urlutils_service');
             drag.element.css('top', drag.element.position().top + 'px');
             drag.element.css('position', 'absolute');
 
+            // The size of #drag-info can be incorrect and not match the actual
+            // size of the ul element. We correct it before to start the drag
+            // and drop.
+            var list = $('[ga-layermanager] ul').get(0);
+            var listWidth = getComputedStyle(list).width;
+            drag.info.css('width', listWidth);
+
             drag.layerManager.mousemove(function(evt) {
               updateDragInfo(evt);
               scrollLayerManager();
@@ -254,36 +262,39 @@ goog.require('ga_urlutils_service');
             drag.ty = evt.clientY - drag.previousClientY;
             drag.previousClientY = evt.clientY;
 
-            cleanDragClass();
-            applyDragClass();
+            drag.info.hide();
+            showDragInfo();
           } else {
             drag.previousClientY = evt.clientY;
           }
         };
 
-        var cleanDragClass = function(ty) {
-          if (ty === undefined || ty !== 0) {
-            var dragInfo = $('[ga-layermanager] li .drag-info');
-            dragInfo.removeClass(drag.under);
-            dragInfo.removeClass(drag.above);
-          }
-        };
-
-        var applyDragClass = function() {
-          var dragClass;
+        var showDragInfo = function() {
+          var direction;
           if (drag.ty > 0) {
-            dragClass = drag.under;
+            direction = drag.under;
           } else if (drag.ty < 0) {
-            dragClass = drag.above;
+            direction = drag.above;
           }
           $('li.drag-enter').each(function(index, elt) {
             if ($(elt).attr('class').indexOf('dragging') === -1 &&
                 drag.ty !== 0) {
-              $(elt).children('.drag-info').addClass(dragClass);
               drag.position.index = $('[ga-layermanager] li').index(elt);
-              drag.position.direction = dragClass;
+              drag.position.direction = direction;
+              updateDragInfoElement(elt);
             }
           });
+        };
+
+        var updateDragInfoElement = function(elt) {
+          var elementPosition = $(elt).position().top;
+          var elementHeight = $(elt).height();
+          if (drag.position.direction === drag.under) {
+            drag.info.css('top', elementPosition + elementHeight - 5);
+          } else if (drag.position.direction === drag.above) {
+            drag.info.css('top', elementPosition - 5);
+          }
+          drag.info.show();
         };
 
         var scrollLayerManager = function() {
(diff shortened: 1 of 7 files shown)


oterral:
Doesn't work on dev and prod mode :  

Dev error : 

```
Uncaught Error: [$injector:modulerr] Failed to instantiate module ga due to:
Error: [$injector:modulerr] Failed to instantiate module ga_layermanager due to:
Error: [$injector:modulerr] Failed to instantiate module ga_layermanager_directive due to:
Error: [$injector:modulerr] Failed to instantiate module ngDraggable due to:
Error: [$injector:nomod] Module 'ngDraggable' is not available! You either misspelled the module name or forgot to load it. If registering a module ensure that you specify the dependencies as the second argument.
```

Jenselme:
@oterral it looks like I forgot to update `index.mako.html`. PR updated. It should work now.

oterral:
thx I test

oterral:
Ok it works, tested on chrome. It seems we loose some css : 

   -   checkboxes have been replaced by native one
   -   the alignement of elements (this one is very important) 
   -   the bottom border of cells
   -   when we click on the tools button there is an overflow weird effect at the bottom of the list.

Jenselme:
@oterral thanks for you remarks. I just pushed a commit that should correct them except the alignment which is still not completely correct. I don't seem to be able to get it right. I will try again tomorrow. If you have any idea, please share!

PS: I changed how the "arrow cursor" works so the bottom border of the cells can be correctly displayed. On the top and bottom of the layer manager, the "arrow cursor" is always completely visible. Let me know if this is a problem or not.

oterral:
@Jenselme you loose the alignement because you remove replace `<label>` with a  `<span>` for the element containing the name of the layer

src/components/layermanager/partials/layermanager.html
```diff
@@ -1,17 +1,27 @@
...
-      <label class="ga-checkbox">
-        <input type="checkbox" ng-model="layer.visible" />
-        <span></span>
-        {{layer.label}}
-      </label>
+        <span class="ga-checkbox">
+          <input type="checkbox" ng-model="layer.visible" />
+          <span></span>
+          {{layer.label}}
+        </span>
```
oterral: keep the `label` element instead of span.

Jenselme:
@oterral I thought it was something like that but if I replace the `span` by a `label`, I loose the text:

screen shot 2015-11-12 at 17 12 35

oterral:
Yes , I guess some small css can fix this.

Jenselme:
I tried some, nothing conclusive so far. I will keep digging.

oterral:
I can have look if you want. Css in the layermanage are a bit sensible.

Jenselme:
@oterral please do!

oterral:
Here is my work https://github.com/geoadmin/mf-geoadmin3/tree/ioda-net_dndrop

I've rebased on master and added one commit https://github.com/ioda-net/mf-geoadmin3/commit/3d40715d1a7a5e3b7070fe8ab3b82a2065e106a4

[Test](https://mf-geoadmin3.dev.bgdi.ch/ioda-net_dndrop/?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&X=190000.00&Y=660000.00&zoom=1)

It's still not perfect

oterral:
I would like to have something smoother, I found that a bit brutal when we drop an element, we are not sure what's happening.

oterral:
LIke this https://pornel.net/slip/

Jenselme:
@oterral thanks for the improvements. I like how you managed the tools to work. I'll try to do more testing/improvemnts next week.

There is one problem with your work thought: each time I drag or drop a layer, it is selected/deselected. This is why I changed `<label class="ga-checkbox">` to `<span class="ga-checkbox">`. It may not be a big problem to correct it. I also find the empty emplacement for the layer we drag confusing. I fixed it with CSS but it may have generated some problem with the other layers.

I agree that something smoother would be better. I tried to do it at one point, but this feature took me far more time that I originally imagined.

tigerfoot:
@oterral thanks for the work done. And yes the pornel slip example is really nice how it work.
Tested on mobile, really smooth.
@Jenselme don't worry too much about the time spend. I'm really in favor of this.
We will find a way.

oterral:
I'm glad to hear you will have the time to work on this, it could be really a cool feature even on mobile if it's smooth.

The drawback of using a angular directive which manage events is the digest cycles which are triggered. On desktop, it's fine but on mobile that breaks the smoothness of the dnd (in perticular on old mobile). That's why may be using a pure js library which manage the dnd and only trigger a digest cycle when drop is done could be better.

Jenselme:
> The drawback of using a angular directive which manage events is the digest cycles which are triggered. On desktop, it's fine but on mobile that breaks the smoothness of the dnd (in particular on old mobile). That's why may be using a pure js library which manage the dnd and only trigger a digest cycle when drop is done could be better.

If I ended with ng-draggable is because I hoped that with a library that supports Angular the integration would be easier mostly with the digest cycle. Your suggestion worth a try I think. I'll see what I can do later this week.

Jenselme:
@oterral I managed to implement a pure js solution based on slip.js. You can view it here: https://github.com/ioda-net/mf-geoadmin3/tree/2747-drag-drop-slipjs What do you think? I still have some problems: sometimes, the layer is not inserted at the right position (I already had that issue with ngDraggable).

Jenselme:
@oterral did you have the time to check my new solution (https://github.com/ioda-net/mf-geoadmin3/tree/2747-drag-drop-slipjs)?

oterral:
sure @Jenselme I'll try to test this tomorrow or on monday.

oterral:
Just tested a bit and it's really GREAT!!!!! That's so smooth. There is bugs but it's a really good start.

src/components/layermanager/LayermanagerDirective.js
```diff
@@ -184,9 +204,132 @@ goog.require('ga_urlutils_service');
...
+            drag.position.index = scope.layers.length - 1;
+          }
+          evt.event.stopPropagation();
+          evt.event.preventDefault();
+          if (currentLayer) {
+            var layersCollection = map.getLayers();
+            var index = scope.layers.indexOf(currentLayer);
+            layersCollection.removeAt(index);
+            var correctedIndex = scope.layers.length - drag.position.index;
+            layersCollection.insertAt(correctedIndex, currentLayer);
```
oterral: Have you tried to use the `scope.moveLayer` function to remove/insert the layer at the good place. Since we use z-index to order layers, It's possible there is vector layers between 2 layers managed by the layermanager directive. May be that's why the layers are sometime not well inserted when we drop it.

Jenselme: Thanks for the tip. It was part of the problem (with the fact that slipjs fired the reorder event more than once). I am doing more testing, but it looks like the slipjs works really good.

Jenselme:
Closing this PR since #3043 proposes a better implementation of drag and drop.

PR #2855, opened 2015-11-10
Set the pitch to -50 deg when flying to a location

### LINKED ISSUE #2713: location search: zoom to : tilt in 45°
When choosing a result in swisssearch, eg creux du van

the camera position is looking into nadir
https://mf-geoadmin3.dev.bgdi.ch/?dev3d=true&topic=geol&lang=fr&bgLayer=ch.swisstopo.pixelkarte-farbe&lon=6.73053&lat=46.93696&elevation=2509&heading=360.000&pitch=-90.000&catalogNodes=1786,1787

expecting result when in 3D: pitch 45° to north
https://mf-geoadmin3.dev.bgdi.ch/?dev3d=true&topic=geol&lang=fr&bgLayer=ch.swisstopo.pixelkarte-farbe&lon=6.73067&lat=46.92857&elevation=2151&heading=359.347&pitch=-45.000&catalogNodes=1786,1787

fredj:
Fixes #2713

### DIFF
--- a/src/components/controls3d/Controls3dDirective.js
+++ b/src/components/controls3d/Controls3dDirective.js
@@ -73,14 +73,7 @@ goog.require('ga_map_service');
           if (finalAngle > 0 || finalAngle < -Cesium.Math.PI_OVER_TWO) {
             return;
           }
-          var bottom = olcs.core.pickBottomPoint(scene);
-          if (bottom) {
-            var transform = Cesium.Matrix4.fromTranslation(bottom);
-            olcs.core.rotateAroundAxis(
-                camera, -angle, camera.right, transform, {
-                  duration: 100
-                });
-          }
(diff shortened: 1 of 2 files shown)


fredj:
demo: https://mf-geoadmin3.dev.bgdi.ch/flyTo_pitch/?dev3d=true

cedricmoullet:
It makes a kind of jump when it arrives at destination. Any way to make that smoother ?

cedricmoullet:
I tested with Bern and Zürich

fredj:
We have this jump between the animations because cesium is not able to `flyTo` a location with a specific pitch.

To do so we need to know the altitude at the destination and compensate (move the camera south)

fredj:
See #2908 for a better alternative

Issue #2835, opened 2015-11-06
3D BG data with lighting


cedricmoullet:
Process map bg with relief information

ltclm:
Tile generation is ongoing and will take around 3 days for swisstlm3d-karte-farbe and 3 days for swisstlm3d-karte-grau.

```
https://wmts.geo.admin.ch/1.0.0/ch.swisstopo.swisstlm3d-karte-farbe.3d/default/20150501_50/4326/{TileMatrix}/{TileRow}/{TileCol}.jpeg
https://wmts.geo.admin.ch/1.0.0/ch.swisstopo.swisstlm3d-karte-grau.3d/default/20150501_50/4326/{TileMatrix}/{TileRow}/{TileCol}.jpeg
```

gjn:
I don't understand the OP. Relief will be inside swisstlm3d?

ltclm:
yes

ltclm:
**ch.swisstopo.swisstlm3d-karte-farbe.3d** tiles with relief are ready:
```
https://wmts.geo.admin.ch/1.0.0/ch.swisstopo.swisstlm3d-karte-farbe.3d/default/20150501_50/4326/{TileMatrix}/{TileRow}/{TileCol}.jpeg
```

gjn:
Demo upcoming.

gjn:
[Visible difference](https://mf-geoadmin3.dev.bgdi.ch/gjn_relief/?dev3d=true&topic=ech&lang=de&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bafu.wrz-wildruhezonen_portal,ch.swisstopo.swisstlm3d-wanderwege&layers_visibility=false,false,false,false&layers_timestamp=18641231,,,&lon=7.56266&lat=46.25252&elevation=4897&heading=35.005&pitch=-18.462):

![image](https://cloud.githubusercontent.com/assets/2572317/11058116/b5f49dcc-8790-11e5-87ca-db4c67a57c2a.png)

![image](https://cloud.githubusercontent.com/assets/2572317/11058120/c1aab7d2-8790-11e5-8408-2ed03cd3fad7.png)

@davidoesch @cedricmoullet + or -?

cedricmoullet:
+100

davidoesch:
+100

ltclm:
[...]

PR #2856, opened 2015-11-10
Use relief tlm for 3D background

gjn:
All in title. This is the PR for https://github.com/geoadmin/mf-geoadmin3/issues/2835

We need to discuss official names/paths...for alpha, this is ok.


oterral:
Nice!!

gberaudo:
Looks really good!

PR #2861, opened 2015-11-11
Display measure tooltips well-positioned in 3d

oterral:
[Test](https://mf-geoadmin3.dev.bgdi.ch/teo_overlays3d/?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&dev3d=true&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bafu.wrz-wildruhezonen_portal,ch.swisstopo.swisstlm3d-wanderwege,KML%7C%7Chttps:%2F%2Fpublic.dev.bgdi.ch%2FcfG9RDtiR7W54Kg7RWwotA&layers_visibility=false,false,false,false,true&layers_timestamp=18641231,,,,&lon=7.25159&lat=46.62316&elevation=12260&heading=328.830&pitch=-43.380)

Fix #2783

Works only for polygon because lines are under the terrain.

PR #2871, opened 2015-11-12
Hide map's overlays in 3d

### LINKED ISSUE #2783: Overlays are not well positionned
Measure overlays .

oterral:
Alternative to #2861
Fix #2783
[Test](https://mf-geoadmin3.dev.bgdi.ch/teo_overlays3d_hide/?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&dev3d=true&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bafu.wrz-wildruhezonen_portal,ch.swisstopo.swisstlm3d-wanderwege,KML%7C%7Chttps:%2F%2Fpublic.dev.bgdi.ch%2FcfG9RDtiR7W54Kg7RWwotA&layers_visibility=false,false,false,false,true&layers_timestamp=18641231,,,,&lon=7.25159&lat=46.62316&elevation=12260&heading=328.830&pitch=-43.38)

### DIFF
--- a/src/components/map/MapDirective.js
+++ b/src/components/map/MapDirective.js
@@ -156,6 +156,43 @@ goog.require('ga_styles_service');
                 pitch: Cesium.Math.toDegrees(camera.pitch).toFixed(3)
               });
             }, 1000, false));
+
+            var dereg = [];
+            var setRealPosition = function(itemOrEvt) {
+              var item = (itemOrEvt instanceof ol.Overlay) ? itemOrEvt :
+                  itemOrEvt.element;
+              item.set('realPosition', item.getPosition());
+              item.setPosition();
+              dereg.push(item.on('change:position', function(evt) {
+                var val = evt.target.getPosition();
+                if (val) {
+                  item.set('realPosition', val);
+                  item.setPosition();
+                }
+              }));
+            };
+
+            // Watch when 3d is enabled to show/hide overlays
+            scope.$watch(function() {
+              return ol3d.getEnabled();
+            }, function(active) {
+              if (active) {
+                // Hide the overlays
+                map.getOverlays().forEach(setRealPosition);
+                dereg.push(map.getOverlays().on('add', setRealPosition));
+              } else {
+                // Show the overlays
+                dereg.forEach(function(key) {
+                   ol.Observable.unByKey(key);
+                });
+                dereg = [];
+                map.getOverlays().forEach(function(item) {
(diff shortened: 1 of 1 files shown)


gjn:
@oterral testlink does not work for me (I see nothing in 3D, not even background layer). Same behavioiur when checking out branch and compile myself.

oterral:
I've currently no tiles on another branch too so it's not this PR in particular. I rebase in case.

oterral:
Rebase redeployed @gjn

gjn:
It also hides the context-popup link (crosshair=true parameter) in 3D. This should not be the case. Maybe there are others.

oterral:
It's normal if it's an overlay we hide it on 3d because they are not well-positioned.
The question is should we display the crosshair as a vector layer ?

gjn:
hmm. The icon we show when selecting a search result is not an overlay? It's the same icon as crosshair icon. So both should be shown, IMO. Measurements icons are a different animal.

oterral:
We have transformed the search marker in a vector layer (old FeatureOverlay). I guess the crosshair is still an Overlay (= html elements).

oterral:
So i think we could/should use a vector layer for the crosshair.

gjn:
So it's probably best to transform crosshair to vector layer.

oterral:
jenkins test

oterral:
Pr update and rebased , can we merge @gjn?

src/components/map/MapDirective.js
```diff
@@ -156,6 +156,43 @@ goog.require('ga_styles_service');
...
+                var val = evt.target.getPosition();
+                if (val) {
+                  item.set('realPosition', val);
+                  item.setPosition();
+                }
+              }));
+            };
+
+            // Watch when 3d is enabled to show/hide overlays
+            scope.$watch(function() {
```
gjn: Is this triggered on every digest cycle?

oterral: yes

gjn: It is, but it might be considered cheap.

oterral: Yes. we have this watch on multiple places maybe we should broadcast an event when 3d is active/deactive.

gjn: Yes, this might be worthwile.

Issue #2818, opened 2015-11-03
IE 11 Perfomance 3d: setting the viewer.resolutionScale property in IE only, for example, try it with 0.5 


davidoesch:
Proposal to fix IE 11

"
Are we talking unusably slow, or just noticeably slower than all other" browsers?  For example, what does this link report for FPS.

If it's the former, you may be experiencing software rendering caused by IE's aggressive and large hardware black list.  If you look in the IE dev console when running Cesium, do you see any messages about software mode or rendering?  The best solution here is to make sure your video drivers are up to date and that Windows is fully patched, even then IE 11 might refuse to use hardware accelerated graphics for your card (what graphics card is it, by the way?)

If it's the latter, than the specific performance problems you are seeing are most likely a result of IE 11 not supporting transferable typed arrays, which causes additional copy and slower performance when marshaling data to and from web workers.  Also, IE 11's JavaScript engine is just flat out slow compared to Chrome, Firefox, and Edge.

While there's no magic "make faster" setting you can enable (if there was, we would already be doing it for you) you can try and make a trade off between quality and performance by rendering at a lower resolution in IE compared to other browsers.  You do this by setting the viewer.resolutionScale property in IE only, for example, try it with 0.5 and see if it affects my first link."

see https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/cesium-dev/OTwnQkSpInw

-> can we test the impact of " setting the viewer.resolutionScale property in IE only, for example, try it with 0.5 "

gjn:
I played around a bit with fed IE11. It seems to me that performance is generally worse by factor 2-3 for cesium demo viewer; but still slower for our application. I think it's because we load more data than the cesium demo viewer.

Each request individaully does not appear to be slower to me, but they are issued in much greater intervals. This might be triggered by slow/slower software rendering as CPU usage increases considerably if I open the app in IE11 (as compared to chrome).

@gberaudo The `resolutionScale`  property mentioned in the thread above is attached to the viewer widget of cesium and we don't have it in ol3-cesium. Is there an easy way to add it?

gjn:
I've created an issue with our IT regarding GPU and IE11.

gberaudo:
@gjn, it seems easy to implement it. I will have a try to it.

gberaudo:
@gjn, I added the ability in ol3-cesium: https://github.com/openlayers/ol3-cesium/pull/281.

[...]

PR #2877, opened 2015-11-13
Show alert for 3D on IE < Edge

gjn:
With this PR, the application displays an alert when a) swithcing to 3D or b) starting application in 3D for IE9, IE10 and IE11.

To adress https://github.com/geoadmin/mf-geoadmin3/issues/2818


gjn:
I'll merge once tests are done because it's needed for tests on Monday.

Issue #2872, opened 2015-11-12
South west of switzerland is missing in 3d 


oterral:
https://mf-geoadmin3.dev.bgdi.ch/?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&dev3d=true&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bafu.wrz-wildruhezonen_portal,ch.swisstopo.swisstlm3d-wanderwege&layers_visibility=false,false,false,false&layers_timestamp=18641231,,,&lon=6.08960&lat=46.20285&elevation=18158&heading=4.306&pitch=-89.994

oterral:
Caused by ol3, ol3-cesium and cesium update https://github.com/geoadmin/mf-geoadmin3/pull/2787

gberaudo:
This happens at all border tiles with missing terrain.
The tiles are not refined, and stay at zoom level 8:
![non_refining_border_tiles_zoom8](https://cloud.githubusercontent.com/assets/7294662/11142671/7e6192aa-89ef-11e5-87bf-ec6fa87a908a.png)

gjn:
There are other regions missing as well:

https://mf-geoadmin3.dev.bgdi.ch/?&lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&dev3d=true&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bafu.wrz-wildruhezonen_portal,ch.swisstopo.swisstlm3d-wanderwege&layers_visibility=false,false,false,false&layers_timestamp=18641231,,,&lon=9.96651&lat=46.34553&elevation=47089&heading=359.569&pitch=-89.515

I think it's because of the terrain layer.json definition, which simply excludes this regions for bigger LOD's

gjn:
@gberaudo @oterral Oli referred to code change. Would you know where to look?

(I'm currently analysing layer.json file, but it's maybe not related to layer.json).

gjn:
I think the current layer.json is correctly defined (at least for LOD the tiles are defined, but not loaded for given region). So problem seems to be in code.

gjn:
@gberaudo @oterral Could it be this: https://github.com/camptocamp/cesium/commit/6d42f47726df2343f4e210206381e1c0b5487d77

gberaudo:
[...]

PR #2883, opened 2015-11-14
New timestamp for terrain tiles

loicgasser:
Zoom level 8 - 11 have been generated with the new data including Switzerland surroundings.

We have no more artifacts (for these zoom levels), tiles are slightly smaller and it seems to fix:
https://github.com/geoadmin/mf-geoadmin3/issues/2872

We'll start the production of the mesh for the other LODs next week.

[Test Link](https://mf-chsdi3.dev.bgdi.ch/shorten/683233e408)

(We're still copying level 17, it should be over in a couple of hours)


gjn:
Great stuff. Strange that it fixes #2872, but I'll take it.

I'll merge it and we'll use this timestamp for the alpha.

PR #2886, opened 2015-11-15
Enable fog by default

gjn:
Fog is enabled by default.


cedricmoullet:
@davidoesch really sure that we need to have the frog ?
image

davidoesch:
since we save quite a lot of traffic with this --- let's start with this...

PR #2908, opened 2015-11-17
Set the pitch to -50 deg when flying to a location

### LINKED ISSUE #2713: location search: zoom to : tilt in 45°

fredj:
Fixes #2713

### DIFF
--- a/src/components/map/MapService.js
+++ b/src/components/map/MapService.js
@@ -1437,7 +1437,7 @@ goog.require('ga_urlutils_service');
    */
   module.provider('gaMapUtils', function() {
     this.$get = function($window, gaGlobalOptions, gaUrlUtils, $q,
-        gaDefinePropertiesForLayer) {
+        gaDefinePropertiesForLayer, $http) {
       var resolutions = gaGlobalOptions.resolutions;
       var lodsForRes = gaGlobalOptions.lods;
       var isExtentEmpty = function(extent) {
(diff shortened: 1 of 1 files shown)


fredj:
demo: https://mf-geoadmin3.dev.bgdi.ch/fredj/?dev3d=true

gjn:
Animation is good, but it's quite far away, IMO.

gberaudo:
There is an option in Cesium we could use to limit the maxium height of the fly.

fredj:
I'll try to add this option, thanks

src/components/map/MapService.js
```diff
@@ -1530,20 +1530,39 @@ goog.require('ga_urlutils_service');
           return layer;
         },
 
+        flyToAnimation: function(ol3d, center, destination, defer) {
+          var scene = ol3d.getCesiumScene();
+          var camera = scene.camera;
+          var globe = scene.globe;
+          var carto = globe.ellipsoid.cartesianToCartographic(destination);
+          $http.get(gaGlobalOptions.apiUrl + '/rest/services/height', {
```
oterral: @fredj on this [PR](https://github.com/geoadmin/mf-geoadmin3/pull/2861/files#diff-66ac86f99915855fde57c0a20d8f6d6bR1797) I use a cesium function to get the height of the Terrain without using the height service. Can you have a look a tell us which method we should use height service or Cesium function?

fredj: In this case, the cesium function wont give the right value because the terrain tiles are not loaded yet (height at the flyTo destination)

oterral: Are you sure because I specify the zoom level 17 as parameter of the function. May be Cesium load the tiles it needs to make the calculation. I haven't verfiied but I had presumed.

gberaudo: @oterral, you can use the sample terrain call. It will work but the drawback is that it will download a full tile (at level 17 since it is what you asked for).

oterral: Yes but this tile will be downloaded as well when you zoom in . So I don't think it's problem.

gberaudo: Cesium does not keep the terrain tile in cache but indeed the browser will probably do.
To me either way (sampling or calling an external service) are good.

gjn:
We could use the height service and add 50m (different ellipsoid).

Usage of height service: https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/contextpopup/ContextPopupDirective.js#L131

cedricmoullet:
Why not use te geodesy service ?

2015-11-27 17:31 GMT+01:00 Gilbert Jeiziner <notifications@github.com>:

> We could use the height service and add 50m (different ellipsoid).
>
> Usage of height service:
> https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/contextpopup/ContextPopupDirective.js#L131
>
> —
> <https://github.com/geoadmin/mf-geoadmin3/pull/2908#issuecomment-160171101>
> .
>



-- 

gjn:
That's good enough for me.

PR #2912, opened 2015-11-17
Manage render errors

### LINKED ISSUE #2890: 3D mode:  does not load in IE11 on mobile device (surface)
OS: Windows 8
BROWSER: IE11
LINK: https://mf-geoadmin3.int.bgdi.ch/3dalpha/
STEPS TO REPRODUCE: click on 3D button
RESULT: background becomes white and then application freezes.

oterral:
Fix  #2890

### DIFF
--- a/src/js/MainController.js
+++ b/src/js/MainController.js
@@ -95,6 +95,15 @@ goog.require('ga_topic_service');
         startWith3D = true;
       }
 
+      var onRenderError = function(scene, error) {
+        $scope.globals.is3dActive = undefined;
+        alert($translate.instant('3d_render_error'));
+        $window.console.error(error.stack);
+        // Avoid the alert comes twice
+        $scope.ol3d.getCesiumScene().renderError.removeEventListener(
+           onRenderError);
+      };
+
       var cesium = new GaCesium($scope.map, gaPermalink, gaLayers,
                                 gaGlobalOptions, gaBrowserSniffer, $q,
                                 $translate);
@@ -102,6 +111,9 @@ goog.require('ga_topic_service');
         $scope.ol3d = ol3d;
         if (!$scope.ol3d) {
           $scope.globals.is3dActive = undefined;
+        } else {
+          $scope.ol3d.getCesiumScene().renderError.addEventListener(
+              onRenderError);
         }
       });
 
--- a/src/locales/empty.json
+++ b/src/locales/empty.json
@@ -355,5 +355,6 @@
   "help_search_data": "",
   "query_operator": "",
   "no_more_information": "",
-  "3d_ie11_alert": ""
+  "3d_ie11_alert": "",
+  "3d_render_error": ""
 }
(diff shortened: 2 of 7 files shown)


src/js/MainController.js
```diff
@@ -95,6 +95,15 @@ goog.require('ga_topic_service');
         startWith3D = true;
       }
 
+      var onRenderError = function(scene, error) {
+        $scope.globals.is3dActive = undefined;
+        alert($translate.instant('3d_render_error'));
+        $window.console.error(error.stack);
```
gjn: We want to keep that?

src/lib/Cesium/Cesium.js
```diff
@@ -139438,17 +139438,17 @@ define('Scene/QuadtreeTile',[
...
-                result[index++] = new QuadtreeTile({
+        for (var y = topRight.y; y <= bottomLeft.y; y++) {
+            for (var x = bottomLeft.x; x <= topRight.x; x++) {
+                result.push(new QuadtreeTile({
                     tilingScheme : tilingScheme,
                     x : x,
                     y : y,
                     level : level
-                });
+                }));
```
gjn: These changes here seem to be unrelated...

src/locales/empty.json
```diff
@@ -355,5 +355,6 @@
   "help_search_data": "",
   "query_operator": "",
   "no_more_information": "",
-  "3d_ie11_alert": ""
+  "3d_ie11_alert": "",
+  "3d_render_error": ""
```
gjn: Translations needed before merge

src/lib/Cesium/Cesium.js
oterral: Totally I miss my checkout of this file

src/js/MainController.js
oterral: I'm in favor to keep this because it's a fatal error of cesium (we already use it when the inialisation failed). It's easier to see why the 3d doesn't work, and also because the native webgl msg is not so clear.

gjn: ok

oterral:
@davidoesch what message do you want to display?

davidoesch:
the one we already have in place for IE11

gjn:
This is different. The one before is saying to switch browser because of performance.

This is saying 3D does not work and might also happen on other platforms/browsers.

davidoesch:
did you find anything at CESIUM? How do they handle this? I did not find a
text

On Wed, Nov 18, 2015 at 11:44 AM, Gilbert Jeiziner <notifications@github.com

oterral:
Cesium displays only the native error message and the stacktrace

davidoesch:
did update all msgid "3d_render_error"

@pauloamado the https://support.google.com/maps/answer/3096703?hl=en we will copy paste for the 3d help

oterral:
@gjn translation updated

gjn:
Thanks alot.

PR #2914, opened 2015-11-18
Minor update of cesium

oterral:


oterral:
Triial self merge

gjn:
Why was it needed?

oterral:
That fix the calculation of minimum tiles to download at level 8. Without this we can't defined a smaller rectangle for the terrain.

PR #2918, opened 2015-11-19
Remove lazy loading after idle time

gjn:
As discussed in yesterday's meeting, this PR removes the lazy loading of the cesium library after a given idle time. It saves data transfered (to all clients not using 3D) in exchange for longer 3D loading times.

On a 2G connection, initial 3D view is 15s slow (50s insead of 35s). On my home connection, I loose < 1s. I think that's reasonable.

Furthermore, lazy loading was disabled anyway for mobile devices :)


gberaudo:
LGTM

PR #2936, opened 2015-11-24
Fix #2934: Remove preview layers from time management

### LINKED ISSUE #2934: [catalog & time button] button active with non time-enabled layers
UC: open app, go into ECH geocatalog, navigate in the catalog to "Base Maps", load one of the National Map "noscale" layers
Result: the time button on the right side below the zoom-in / zoom-out buttons is visible and functional and stays functional even when I drop the layer from the selection.
Expected: button is not visible at all.

oterral:
Fix #2934 
[Test](https://mf-geoadmin3.dev.bgdi.ch/teo_2934/?topic=luftbilder&lang=fr&bgLayer=ch.swisstopo.pixelkarte-grau&layers=ch.bfs.gebaeude_wohnungs_register,ch.bafu.wrz-wildruhezonen_portal,ch.swisstopo.swisstlm3d-wanderwege,ch.swisstopo.pixelkarte-farbe-pk1000.noscale,ch.swisstopo.zeitreihen,ch.swisstopo.lubis-luftbilder-dritte-firmen,ch.swisstopo.lubis-luftbilder_infrarot&layers_visibility=false,false,false,false,false,true,true&X=204700.00&Y=745050.00&zoom=0&layers_timestamp=,,,,18641231,20011231,20081231&catalogNodes=1179,1180,1186)

### DIFF
--- a/src/components/catalogtree/CatalogitemDirective.js
+++ b/src/components/catalogtree/CatalogitemDirective.js
@@ -80,13 +80,9 @@ goog.require('ga_layer_metadata_popup_service');
               // Leaf
               } else if (!gaBrowserSniffer.mobile) {
                 iEl.on('mouseenter', function(evt) {
-                  scope.$applyAsync(function() {
-                    addPreviewLayer(scope.map, scope.item);
-                  });
+                  addPreviewLayer(scope.map, scope.item);
                 }).on('mouseleave', function(evt) {
-                  scope.$applyAsync(function() {
-                    removePreviewLayer(scope.map);
-                  });
+                  removePreviewLayer(scope.map);
                 });
               }
               compiledContent(scope, function(clone, scope) {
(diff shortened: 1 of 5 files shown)


oterral:
jenkins test

gjn:
build error because of js eror, I think.

oterral:
jenkins test

gjn:
Nice refactoring, and works well.

There are pro and cons to go more jquery and remove angular specific constructs (mouseenter and mouseleave) in this case. Where do we want to go with this in general?

oterral:
For me on mobile, ng-mousenter and ng-mouseleave must be banished to avoid digest cycle. 
On desktop, we don't care.
Ideally it would be great to make this distinction in   the partials like we do in `index.mako` :+1: 

```
<div id="elt"
if desktop
   ng-mouseenter=""
   ng-mouseleave=""
endif
></div>
```

oterral:
Or maybe rewrite a custom `ga-mouseenter`, `ga-mouseleave` which avoid digest cycle if we are on mobile.

gjn:
Does ng-mouseenter/ng-mouseleave is even triggered on mobile, without any mouse?

If yes, the approach in this PR is reasonable.

oterral:
yes these events are triggered.

gjn:
I'm fine then.

PR #2943, opened 2015-11-27
Update coordinates in tests

### LINKED ISSUE #2941: Swisssearch test fails
It's a swisssearch parameter test [1] that fails. And indeed, if I re-do it by hand, it fails. Looking at the code, it looks for an element with `Flugplatz`, but it doesn't exist. It seems that changes in the sphinxsearch index created this failure. Test needs to be adapted to search for an existing element and click it.

@alcapat Can you try to adapt?

[1] https://github.com/geoadmin/mf-geoadmin3/blob/master/test/selenium/swisssearch_test.js#L20

oterral:
Fix #2941


oterral:
Tests pass I merge

gjn:
ugh. was working on this :)

Thanks alot.

gjn:
@alcapat Most probably the sauclab tests need to be adapted as well.

PR #2940, opened 2015-11-25
3d keyboard controls

fredj:
fixes #2938 

preview: https://mf-geoadmin3.dev.bgdi.ch/3d_keyboard_controls/?dev3d=true

PR #2952, opened 2015-12-01
[WIP] add help 3d config

pfanguin:


pfanguin:
@cedricmoullet, @gjn what is the status of this ? Is it always of actuality?

pfanguin:
test link : https://mf-geoadmin3.dev.bgdi.ch/fap_help3d/

cedricmoullet:
@davidoesch @oterral @gjn @pfanguin :
My proposal would be to complete this:
1. Add a permanent text on the left if swissimage is added on the map
2. Complete the navigation elements with the arrows implemented here: https://github.com/geoadmin/mf-geoadmin3/pull/2940
3. Make it multlingual
@oterral could you work with @pfanguin to complete this PR ?

davidoesch:
not at all convinced

1) help icon is way to big/prominent. now the control has 3 elements.it is not a map function as tilt and rotate but a help-- so a small "'?" in grey would help. I would put it in the middle bottom (where pegman will be). 
2) use a windwo as the infobox . to print and close it. the current solution is surely nice, but does not follow CI CD we have. sorry
3) Actually: the help is wrong: it does not explain the controls tilt / rotate : so those two icons are now actually  obsolete
4) for time function: we insisted NOT to make/provide a help (remember :flyout) - same applies for 3D IMHO..
5) help belongs into help.geo.admin.ch

Therefore I propose the following (pareto)
A) don't add the mouse/keyboard  help to ALPHA.
B) put this help into help.geo.admin.ch
C) add a mouseover help to tilt and rotate button (as we have for zoom + -)

cedricmoullet:
we need to display:
1. The way we navigate with keys
2. The swissimage message
Howto ?

gjn:
B) adding that the help button in-between the navigation icons will create a popup containing the help of help.geo.admin.ch (as we have for all other help buttons inside the app.

Swissimage info: currently brainstorming.

oterral:
@cedricmoullet @gjn is this PR still needed ?

PR #2956, opened 2015-12-03
Revert "Update coordinates in tests"

oterral:
Reverts geoadmin/mf-geoadmin3#2943


gjn:
We have 2 result sets, it seems. Please see https://github.com/geoadmin/service-sphinxsearch/pull/245

gjn:
I think what's currently on master is now correct with the new index as specified in sphinxsearch PR above. So this should fail now. Let's see.

jenkins test

oterral:
Ok thx for the check

gjn:
Note, it's wrong yet again. It will be unstable until tonight. Then I hope we'll have it under control.

PR #2969, opened 2015-12-09
Prevent users to augment the size of the texteara for feedback

Jenselme:
Users shouldn't be able to make a textarea so big, the 'Send' button is
not accessible any more.

### DIFF
--- a/src/components/feedback/style/feedback.less
+++ b/src/components/feedback/style/feedback.less
@@ -15,6 +15,10 @@
   .progress {
     margin-top: 8px;
   }
+  textarea {
+    max-width: 320px;
+    max-height: 75px;
+  }
   [ga-draw] {
     width: 100%;
 


src/components/feedback/style/feedback.less
```diff
@@ -15,6 +15,10 @@
   .progress {
     margin-top: 8px;
   }
+  textarea {
+    max-width: 320px;
+    max-height: 75px;
+  }
```
oterral: That makes sense to block the width but use `max-width:100%;`.

I'm not sure about the height, when you have a long text it can beuseful.

Jenselme: @oterral I added the height because without we can have situation like this:

screenshot from 2015-12-09 14-28-50

It can also be fixed with an overlay but I found the `max-height` solution to be more pleasant.

OK to use `max-width:100%;`

PR #2977, opened 2015-12-10
Correct class for warn color

gjn:


loicgasser:
I thought I pushed it sorry.

Issue #2980, opened 2015-12-10
Error message after usage for measurement tool in layermanager


Jenselme:
Linked to #2975. The cause of the bugs are the same, I have the same ideas for solution. I think we should expect to be bitten because of this quite often.

How to reproduce:

- Enable draw & measure on map
- Use the measurement tool
- Disable draw & measure on map

Problems:
- Try to hide the layer. I get in the Chrome console (caused by `for (var i in features)` in https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/measure/MeasureService.js, 4 ocurences than may cause the bug) :
```
TypeError: olFeature.get is not a function
    at Object.isMeasureFeature (http://dev.geojb/components/map/MapService.js:1748:54)
    at null.<anonymous> (http://dev.geojb/components/measure/MeasureService.js:268:32)
    at goog.events.EventTarget.fireListeners (http://dev.geojb/lib/ol3cesium-debug.js:11841:23)
    at Function.goog.events.EventTarget.dispatchEventInternal_ (http://dev.geojb/lib/ol3cesium-debug.js:11938:26)
    at goog.events.EventTarget.dispatchEvent (http://dev.geojb/lib/ol3cesium-debug.js:11753:34)
    at ol.Object.notify (http://dev.geojb/lib/ol3cesium-debug.js:12277:8)
    at ol.Object.set (http://dev.geojb/lib/ol3cesium-debug.js:12297:12)
    at ol.layer.Base.setVisible (http://dev.geojb/lib/ol3cesium-debug.js:19939:8)
    at Object.defineProperties.visible.set (http://dev.geojb/components/map/MapService.js:94:22)
    at fn.assign (eval at <anonymous> (http://dev.geojb/lib/angular.js:13145:15), <anonymous>:4:410)
```
- Try to modify the transparency of the layer. I get in the Chrome console (caused by `for (var i in overlays)` in MeasureServices again):
```
TypeError: overlays[i].getElement is not a function
    at null.<anonymous> (http://dev.geojb/components/measure/MeasureService.js:284:33)
    at goog.events.EventTarget.fireListeners (http://dev.geojb/lib/ol3cesium-debug.js:11841:23)
    at Function.goog.events.EventTarget.dispatchEventInternal_ (http://dev.geojb/lib/ol3cesium-debug.js:11938:26)
    at goog.events.EventTarget.dispatchEvent (http://dev.geojb/lib/ol3cesium-debug.js:11753:34)
[...]

PR #2981, opened 2015-12-10
Remove Array overwrite and more robust code

gjn:
This fixes https://github.com/geoadmin/mf-geoadmin3/issues/2980


oterral:
There is one more in https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/StylesFromLiteralsService.js#L121

gjn:
Isn't `this.styles[geomType]` an object, and not an array?

oterral:
yes it's an object.

loicgasser:
LGTM

PR #2984, opened 2015-12-11
[WIP] Activate 3D

gjn:
This is the final merge before we deploy 3D.

Snapshot to disable 3d will be created just before we merge this.

### DIFF
--- a/src/index.mako.html
+++ b/src/index.mako.html
@@ -718,11 +718,7 @@ itemscope itemtype="http://schema.org/WebApplication"
 
         module.constant('gaGlobalOptions', {
           //dev3d to be removed once 3d goes live
-% if staging=='prod' :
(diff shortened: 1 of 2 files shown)


src/index.mako.html
```diff
@@ -718,11 +718,7 @@
 
         module.constant('gaGlobalOptions', {
           //dev3d to be removed once 3d goes live
-% if staging=='prod' :
-          dev3d: false,
-% else:
-          dev3d: !window.location.search.match(/(dev3d=false)/),
-% endif
+          dev3d: true,
```
cedricmoullet: Could we get rid of this variable ?

gjn: not yet. It's the easiest 3D switch for now.

PR #2986, opened 2015-12-11
Fix KML loading

gberaudo:
Test link: https://mf-geoadmin3.dev.bgdi.ch/gberaudo_fix_examples/

AssertionError: Assertion failed: invalid extent or geometry
    at new goog.asserts.AssertionError (https://map.geo.admin.ch/src/lib/ol3cesium-debug.js:6582:20)
    at Object.goog.asserts.doAssertFailure_ (https://map.geo.admin.ch/src/lib/ol3cesium-debug.js:6639:11)
    at Object.goog.asserts.assert [as assert] (https://map.geo.admin.ch/src/lib/ol3cesium-debug.js:6669:18)
    at ol.View.fit (https://map.geo.admin.ch/src/lib/ol3cesium-debug.js:24859:18)
    at https://map.geo.admin.ch/src/components/map/MapService.js:755:33
    at processQueue (https://map.geo.admin.ch/src/lib/angular.js:14569:28)
    at https://map.geo.admin.ch/src/lib/angular.js:14585:27
    at Scope.$eval (https://map.geo.admin.ch/src/lib/angular.js:15848:28)
    at Scope.$digest (https://map.geo.admin.ch/src/lib/angular.js:15659:31)
    at Scope.$apply (https://map.geo.admin.ch/src/lib/angular.js:15953:24)

### DIFF
--- a/src/components/map/MapService.js
+++ b/src/components/map/MapService.js
@@ -752,7 +752,10 @@ goog.require('ga_urlutils_service');
               }
 
               if (options.zoomToExtent) {
-                olMap.getView().fit(olLayer.getExtent(), olMap.getSize());
+                var extent = olLayer.getExtent();
+                if (extent) {
+                  olMap.getView().fit(extent, olMap.getSize());
+                }
               }
             }
           });


src/components/map/MapService.js
```diff
@@ -752,7 +752,8 @@ goog.require('ga_urlutils_service');
               }
 
               if (options.zoomToExtent) {
-                olMap.getView().fit(olLayer.getExtent(), olMap.getSize());
+                var extent = olLayer.getSource().getExtent();
```
oterral: The source can be an `ol.source.ImageVector` which has not a `getExtent` function.

What's wrong with `layer.getExtent()`?

oterral:
@gberaudo the problem is probably here : https://github.com/geoadmin/mf-geoadmin3/blob/gberaudo_fix_kml_loading/src/components/map/MapService.js#L704

How do you get this error?

src/components/map/MapService.js
gberaudo: The result of `layer.getExtent()` is `undefined` while the result of `source.getExtent()` is a correct array.

gberaudo:
@oterral I got the error by opening a KML file from my computer.
I see, my KML was outside Switzerland so `intersectWithDefaultExtent()` returned `undefined`.

oterral:
So it's normal that returns undefined, We don't want the map zoom outside switzerland. So I suggest you test if `getLayer().getExtent()` is undefined the if not zoom otherwise do nothing.

oterral:
lgtm thx @gberaudo

gberaudo:
Thanks for the review @oterral.

PR #2995, opened 2015-12-15
Add icons for retina display

oterral:
Improve a lot the display on retina screen.(and generally when `devicePixelRatio` is more than 1.3)
[Test](https://mf-geoadmin3.dev.bgdi.ch/teo_retina/?lang=fr)

Drawback: the size of app.less is a lot bigger.

Thx @pfanguin and @pauloamado for the icons


oterral:
The problem of the bigger size is caused by the big bund logo. May be we could avoid using `data-uri` for these logos.

oterral:
100kb more for app.css file and one http request more for the big bund logo

@gjn what do you think ?

For me it's worth using it, the display is a lot nicer on retina.

gjn:
Ideally, we should have both. And on detection of device, load what is most appropriate.

Partial loading of content is something we should invest a little time for map.geo. Initial load is one of the key parameters we should always try to optimize.

oterral:
We can load a different css file for pixel ratio > 1.3 . I didin't know that exists but it seems we could https://css-tricks.com/resolution-specific-stylesheets/

gjn:
Adds complexity in handling of css and generating them, but I would be ok with that. What do you think? Easily done?

oterral:
I don't know I'm looking into it this afternoon.

oterral:
We could load a different css file but it will break the appcache configurations and no way to fixed it without creating useless complexity in the build.

According to this doc : http://mydevice.io/devices/  more and more devices have a pixelRatio > 1, I think we can use the `x2` images by default and remove use of old images.  With modifying the css of the  Background selector we can reduce the size of the current css files and use `x2` images everywhere.

What do you think @gjn ?

oterral:
The app.css file is 50 ko  lighter with this solution.

oterral:
Ready to review merge.

oterral:
jenkins test

gjn:
@oterral rebase.

oterral:
@gjn Rebased

oterral:
I've merged it's just new images.

Issue #3003, opened 2015-12-16
hiking time: overlapping in footer


davidoesch:
draw profile

result 
![unbenannt](https://cloud.githubusercontent.com/assets/4577727/11841472/c7efb582-a3fd-11e5-99b5-6768509186cd.PNG)

expected result
there is a bigger space between the result and and the axis annotation so there is now overlap

mariavkourouni:
#3007

oterral:
Fixed

PR #3007, opened 2015-12-16
Hiking time: fix overlapping in footer

mariavkourouni:
Fix for https://github.com/geoadmin/mf-geoadmin3/issues/3003

[Demo](http://mf-geoadmin3.dev.bgdi.ch/kom_hikingTime/)

### DIFF
--- a/src/components/profile/ProfileService.js
+++ b/src/components/profile/ProfileService.js
@@ -517,7 +517,7 @@ goog.require('ga_urlutils_service');
           group.append('text')
               .attr('class', 'ga-profile-icon')
               .attr('x', 480)
-              .attr('y', elevLabelY)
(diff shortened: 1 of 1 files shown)


gjn:
oops, needs rebase now.

gjn:
@mariavkourouni I still think the distance between the label and the numbers is too big (in vertical direction).
image

Also, in the PR post (first), please add the issue when you are adressing the issue. And use the 'fix for' prefix, which will autoclose the issue once it's merged.

mariavkourouni:
do you mean too small? Yes, I agree. The svg low boundary is one pixel below the numbers. I can try adjust the diagram height or the svg height and see which option is better.

oterral:
lgtm

Issue #2942, opened 2015-11-26
Mouseclick sensitivity tooltip: very small


davidoesch:
This is a reoccurring  issue for all tooltips
"Ich finde den sensiblen Bereich für den Klick auf ein Symbol (oranger Kreis) ein bisschen klein, vor allem wenn man weit hineingezoomt hat:"
![unbenannt](https://cloud.githubusercontent.com/assets/4577727/11425944/daad664c-9458-11e5-94e7-cb0b19caed4b.PNG)

Proposal

A) we really do increase once again the click perimeter

or

B) We extend the tooltip highlight / sensitivity from a circle to the icon size , or at least we define a paramtere for the sensitivity related to the mako template

or

C) we minimize the size of the icons smaller on large scale to fit the current extent of the tooltip circle /sensitivity

Decision:

"increase  the click perimeter"

- [ ] define ranges depending on zoom levels, 
- [ ] test them with a couple of typical layers
- ch.bazl.luftfahrthindernis
- ch.bafu.hydroweb-messstationen_gefahren
- ch.bfs.gebaeude_wohnungs_register
- ch.bav.haltestellen-oev
- [ ] decision on how to proceed (add an additional col in BOD? "wait" for vec tiles? etc)

davidoesch:
C) +1

gjn:
C) would require re-creating tiles for all affected layers...And the icons are IMO not too big.

There are 2 issues at play here.

1) We have a fixed tolerance in pixels for the underlying identify request. It's 5 pixels on desktop and 15 on mobile devices. This is clearly too small with high zoom levels. For points, the obvious solution would be to adapt the tolerance to the icon size. But we don't know icon sizes on pixel tiles. It's known during data creation. Maybe it's worth keeping track of those. This would be needed for B)

2) The actual request is not done on the icon/graphical representation, but on the underlying geometry.

I think a first approach would be A), generally increase to click tolerance. We can even define ranges depending on zoom levels, test them with a couple of typical layers and then decide. (Luftfahrthindernisse is a good example as well of this problematic).

Of course, vector layers (with vector tiles) would solve all of this...

davidoesch:
Decision:

"increase the click perimeter"

 define ranges depending on zoom levels,
 test them with a couple of typical layers
ch.bazl.luftfahrthindernis
ch.bafu.hydroweb-messstationen_gefahren
ch.bfs.gebaeude_wohnungs_register
[...]

PR #3010, opened 2015-12-17
Increase left-click pixel tolerance (tooltip request)

gjn:
Based on the discussions and tests in https://github.com/geoadmin/mf-geoadmin3/issues/2942, we do this quick adaption.

We increase the click tolerance from 5px to 10px (15 to 20 on mobile) to increase likelyhood of positive result. This is not the final solution.

@davidoesch Ok?

Vector tiles to the rescue!


loicgasser:
+1

davidoesch:
10px for all zoom levels?

might be to much but give it a try

-> +1

PR #3029, opened 2016-01-07
Don't return the geometry if not highlightable

loicgasser:
We avoid querying a geometry we won't use by setting `returnGeometry` to false if the property `highlightable` (a Boolean) is false. Defaults to true.


gjn:
Yes, I'd like to discuss this, as well as the changes in references PR.

loicgasser:
yep tomorrow.

loicgasser:
ok I adapted it.

gjn:
Nice. Thanks alot.

PR #3078, opened 2016-01-26
Statifying services and layersConfig

procrastinatio:
to avoid using apache proxy (for offline uses, files have to be on the same server)
Files are written to `prd/cache`

The `dev`mode still uses the remote service

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,7 @@ CESIUM_VERSION ?= b4da72d63ca6df06dd4af513fdef15f5ff94ce0c # camptocamp/c2c_patc
 DEFAULT_TOPIC_ID ?= ech
 TRANSLATION_FALLBACK_CODE ?= de
 LANGUAGES ?= '[\"de\", \"en\", \"fr\", \"it\", \"rm\"]'
+LANGS ?= de fr it rm en
 TRANSLATE_GSPREAD_KEYS ?= 1F3R46w4PODfsbJq7jd79sapy3B7TXhQcYM7SEaccOA0
 TRANSLATE_CSV_FILES ?= "https://docs.google.com/spreadsheets/d/1F3R46w4PODfsbJq7jd79sapy3B7TXhQcYM7SEaccOA0/export?format=csv&gid=0"
 TRANSLATE_EMPTY_JSON ?= src/locales/empty.json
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -307,6 +309,11 @@ prd/geoadmin.appcache: src/geoadmin.mako.appcache \
 	    --var "api_url=$(API_URL)" \
 	    --var "public_url=$(PUBLIC_URL)" $< > $@
 
+prd/cache/:
+	mkdir -p $@
+	curl -q -o prd/cache/services http:$(API_URL)/rest/services
+	$(foreach lang, $(LANGS), curl -q -o prd/cache/layersConfig.$(lang) http:$(API_URL)/rest/services/all/MapServer/layersConfig?lang=$(lang);)
+
```
gjn: I think we need a clean cache command as well, to remove old files. Or by default, remove the created files when make clean is called.

gjn:
Don't we risk to hit the same geoip/spamhouse problem when API_URL is api3.geo.admin.ch? (As the request is done from within amazon network)?

Makefile
procrastinatio: `make clean` removes directory `prd`

procrastinatio:
Don't think so. Curling from dev or prod to api3.geo.admin.ch is OK, while proxying is not.

apache/app.mako-dot-conf
```diff
@@ -37,8 +37,13 @@ RedirectMatch ^${apache_base_path}$ ${apache_base_path}/
 
 Alias ${apache_base_path}/src ${apache_base_directory}/src
 Alias ${apache_base_path}/ ${apache_base_directory}/prd/
-RewriteRule ${apache_base_path}/(src|[0-9]+)/layersConfig http:${api_url}/rest/services/all/MapServer/layersConfig [P]
-RewriteRule ${apache_base_path}/(src|[0-9]+)/services http:${api_url}/rest/services [P]
+RewriteRule ${apache_base_path}/src/layersConfig http:${api_url}/rest/services/all/MapServer/layersConfig [P]
```
oterral: `https`?

oterral:
PR rebased

procrastinatio:
Parsing the downloaded files with python and displaying the topics and layers number as a check. Should be enough for now.

gjn:
Deployed to int. Works well.

PR #3080, opened 2016-01-27
Makefile changes

oterral:
Some minor changes:

  - Adding global variable for bin files
  - Symplify lessc process


Makefile
```diff
@@ -380,64 +380,58 @@ prd/checker: src/checker
...
 
-src/style/app.css: src/style/app.less \
-	    src/style/print.less \
-	    src/style/offline.less \
-	    src/style/ga_bootstrap.less \
-	    src/style/ga_variables.less \
-	    $(SRC_COMPONENTS_LESS_FILES) \
-	    node_modules
-	node_modules/.bin/lessc $(LESS_PARAMETERS) $< $@
+src/style/app.css: $(SRC_LESS_FILES)
```
loicgasser: node_modules dependency disappeared. I am not sure this is wanted.

loicgasser:
+1 for this, one comment otherwise nice improvement.

Makefile
oterral: since there is none less file dependent of a node module. I think we don't need it.

oterral: and lessc node module is added when you do a `make all` the first time.

PR #3081, opened 2016-01-27
Remove version parameter to get fa in appcache

oterral:
Fix the loading of font from appcache on Safari.


oterral:
jenkins test

Issue #3084, opened 2016-01-27
icons do not load


davidoesch:
Android 601 nexus
Icons like logo and checkboxes do not load
Delete cache 
Restart app everything is OK
Reload by clicking on swiss icon 
Result: icons do not load

Is it just me?

davidoesch:
![screenshot_20160127-201230](https://cloud.githubusercontent.com/assets/4577727/12625231/2a802384-c533-11e5-971c-3a5e31786652.png)

cedricmoullet:
This affects also Safari on Mac. Font awesome is not found

cedricmoullet:
@gjn emergency deploy ?

gjn:
I'll have a look later.

gjn:
Can't reproduce as I don't have the correct environment here.

I assume it's a appcache issue (appcache is such crap...)

@oterral Please have a look first thing in the morning. If you find it, please do an emergency deploy.

cedricmoullet:
I have the same issue on iPhone and Firefox.

mariokeusen:
Can confirm: have no icons neither in Chrome nor in IE11 this morning. And in addition the layer-selection is not visible/available at all.

ltclm:
marcel from our it told me that he has a similar problem with ie11.  
since yesterday the icon on the right side of the swisssearch form is not visible anymore.

![image](https://cloud.githubusercontent.com/assets/5286659/12637797/3bfad16c-c598-11e5-93e6-53343da3a4a9.png)

gjn:
all: doeva disabling and cleaning if browser cache help?

oterral:
Just test on frefox a simple click on th refresh button fix the issue.

oterral:
I redeploy the snapshot

oterral:
It seems a machine hasn't the good version

gjn:
ok.

I could get rid of the problem always with Cltr-F5.

gjn:
For everybody: if you detect similar problems in the future, please safe the page source and send it to me by email.

gjn:
@davidoesch @cedricmoullet Is this ok now?

gjn:
ping @davidoesch

cedricmoullet:
yes, it's fine

2016-01-29 12:24 GMT+01:00 Gilbert Jeiziner <notifications@github.com>:

> ping @davidoesch <https://github.com/davidoesch>
>
> —
[...]

PR #3085, opened 2016-01-28
Rename appcache file to get rid of old trailing appcache files

gjn:
I think alot of our problems that are similar to https://github.com/geoadmin/mf-geoadmin3/issues/3084 is related to old trailing appcache files, especially on desktop version, where they were active initially but then were removed.

According [this article](http://stackoverflow.com/questions/8815280/removing-html5-appcache-manifest-permanently), simply pointing the appcache to something non-existant should be good. So that's what I'll do here for desktop/embed with this PR.

@oterral Maybe we can try to mergency deploy this.

### DIFF
--- a/src/index.mako.html
+++ b/src/index.mako.html
@@ -4,6 +4,8 @@
 itemscope itemtype="http://schema.org/WebApplication"
 % if device == 'mobile' and mode=='prod' :
  manifest="geoadmin.appcache"
+% else:
+ manifest="we.do.not.cache.this"
 % endif
 >
   <head>


gjn:
This would need some tests when it comes to offline.

gjn:
[New info](http://stackoverflow.com/questions/8815280/removing-html5-appcache-manifest-permanently), I'll extend this.

gjn:
Ok, this is updated. See OP.

gjn:
@davidoesch @cedricmoullet Getting rid of a once installed appcaceh is really, really not easy. See the thread here: http://stackoverflow.com/questions/8815280/removing-html5-appcache-manifest-permanently/31083156

We will now try to apply the last comment.

PR #3088, opened 2016-01-28
Use an empty appcache file in desktop and prod mode to clear the old …

oterral:
…appcache config if it exists.

@gjn ready

### DIFF
--- a/src/index.mako.html
+++ b/src/index.mako.html
@@ -4,6 +4,8 @@
 itemscope itemtype="http://schema.org/WebApplication"
 % if device == 'mobile' and mode=='prod' :
  manifest="geoadmin.appcache"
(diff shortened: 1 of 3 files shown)


src/index.mako.html
```diff
@@ -4,6 +4,8 @@
 itemscope itemtype="http://schema.org/WebApplication"
 % if device == 'mobile' and mode=='prod' :
  manifest="geoadmin.appcache"
+% elif device == 'desktop' and mode=='prod' :
+ manifest="empty.appcache"
 % endif
```
gjn: In general: why don't we use it in int/dev?

gjn:
Thanks alot. Please emergency deploy.

PR #3089, opened 2016-01-28
Fix tooltip for layerGroups

loicgasser:
See related issue for explanations: https://github.com/geoadmin/mf-chsdi3/issues/1866

I don't want to use the serverLayerName just yet as I have checked them all for layer groups.
This keeps the old behaviour.

@gjn It may require an emergency deploy.


gjn:
I'll review. When good we'll include it in the deploy.

@oterral Please wait.

gjn:
LGTM. Thanks for acting quickly.

PR #3098, opened 2016-01-29
Remove pixel detection for tooltip on iOS

oterral:
Fix tooltip on iOS both mobile and desktop version

[Test](https://mf-geoadmin3.dev.bgdi.ch/teo_tool/)


oterral:
trivial self merge and need it for emergency deploy

Issue #3096, opened 2016-01-29
schneesport: topic selector


davidoesch:
open per default as well the selection window as in topic ech
(tech lead , project lead)

davidoesch:
Hi there is an issue now on dev:
in the slection window ALL layer should be loaded but onl the six should be active

gjn:
The Heli Layers is now removed as well. Use testlink https://github.com/geoadmin/mf-geoadmin3/pull/3099 for tests that fixes all open issues.

davidoesch:
Just me on Android that in 3d the routes are not showing up? Mapproxy
configuration missing?

https://mf-geoadmin3.dev.bgdi.ch/gjn_plconfigFix/mobile.html?topic=schneesport&lang=de&bgLayer=voidLayer&layers=ch.swisstopo.pixelkarte-farbe-pk50.noscale,ch.swisstopo.hangneigung-ueber_30,ch.swisstopo-karto.hangneigung,ch.bafu.wrz-jagdbanngebiete_select,ch.bafu.wrz-wildruhezonen_portal,ch.bazl.heliports-gebirgslandeplaetze,ch.swisstopo-karto.schneeschuhrouten,ch.swisstopo-karto.skitouren,ch.swisstopo.skitourenkarte-50.metadata,ch.bav.haltestellen-oev&layers_opacity=0.85,0.5,0.2,0.6,0.6,1,0.8,0.8,1,1&layers_visibility=true,false,true,true,true,false,true,true,false,false&catalogNodes=2352&lon=8.36540&lat=46.61928&elevation=7131&heading=360.000&pitch=-40.016

Gilbert Jeiziner <notifications@github.com> schrieb am Fr., 29. Jan. 2016,
14:36:

gjn:
Yes. Most probably layers are missing in mapproxy.
Am 30.01.2016 10:12 nachm. schrieb David Oesch <notifications@github.com>:Just me on Android that in 3d the routes are not showing up? Mapproxy
configuration missing?

[...]

Issue #3093, opened 2016-01-28
Background selection not possible when snowsports topic activated in dev


rfrancel:
When snowsports topic is activated, and we click on background layers selector, only "no background" is proposed:
https://mf-chsdi3.dev.bgdi.ch/shorten/69b3f2f3f1
Other topics propose all the choices.

loicgasser:
oups good catch

loicgasser:
topic title could be "Sports de neige" instead of "Sport de neige"

gjn:
@loicgasser Most certainly introduced by https://github.com/geoadmin/mf-geoadmin3/pull/3068

I think for the background layer list, we can't use the plConfig parameter.

gjn:
@rfrancel Well done to find this!

loicgasser:
Probably not, should I take it or do you have time?

gjn:
I'll have time tomorrow.

loicgasser:
okay

gjn:
@alcapat The backgrounds information is missing in the bod. This is needed for all topics, as it's a list that is _not_ part of the permalink. So please add this array for the schneesport topic.

PR #3068, opened 2016-01-21
Use new PL topic configuration parameter

gjn:
This is to fix https://github.com/geoadmin/mf-geoadmin3/issues/3031

Topic initical configuration can now be specified with a permalink parameter that contains 'bgLayer', 'layers', 'layers_opacity', 'layers_visibility' and 'layers_timestamps' from the bod.

Note that all other plConfig values that might exist are simply ignored. If we want to extent it, we have to adapt front-end code.

[Testlink for Schneesport](https://mf-geoadmin3.dev.bgdi.ch/gjn_topicpl/?topic=schneesport)

PR #3099, opened 2016-01-29
Fix PLConfig to have correct topic initialisatoin

gjn:
Fix for https://github.com/geoadmin/mf-geoadmin3/issues/3096
Fix for https://github.com/geoadmin/mf-geoadmin3/issues/3093

Second part of 3096 will come with a data fix.

[Test](https://mf-geoadmin3.dev.bgdi.ch/gjn_plconfigFix/?topic=schneesport)

### DIFF
--- a/src/components/topic/TopicService.js
+++ b/src/components/topic/TopicService.js
@@ -11,7 +11,8 @@ goog.require('ga_permalink');
    * Topics manager
    */
   module.provider('gaTopic', function() {
-    this.$get = function($rootScope, $http, gaPermalink, gaGlobalOptions) {
+    this.$get = function($rootScope, $http, gaPermalink, gaGlobalOptions,
+                         gaUrlUtils) {
       var topic; // The current topic
       var topics = []; // The list of topics available
 
@@ -28,6 +29,30 @@ goog.require('ga_permalink');
             }
             if (!value.plConfig || !value.plConfig.length) {
               value.plConfig = false;
+            } else {
+              //plConfig overwrites some default settings. So we
+              //apply them here
+              var p = gaUrlUtils.parseKeyValue(value.plConfig);
+              //Overwrite background layer if available
+              if (p.bgLayer) {
+                value.defaultBackground = p.bgLayer;
+              }
+              //Overwrite activated and selected layers
+              if (p.layers && p.layers.length) {
+                value.activatedLayers = [];
+                value.selectedLayers = [];
+                var ls = p.layers.split(',');
+                var lv = p.layers_visibility ?
+                         p.layers_visibility.split(',') : [];
+                for (var i = 0; i < ls.length; i++) {
+                  if (i < lv.length && lv[i] != 'false') {
+                    value.selectedLayers.push(ls[i]);
+                  } else {
+                    value.activatedLayers.push(ls[i]);
+                  }
+                }
+              }
+
             }
           });
           topic = getTopicById(gaPermalink.getParams().topic, true);


loicgasser:
I cannot reproduce the bug with uncompiled code. (Before this PR of course) 

I see that it works but I don't really understand why.

oterral:
LGTM but shouldn't we remove what has been added in https://github.com/geoadmin/mf-geoadmin3/pull/3068/files ?

loicgasser:
what about opacity and timestamps?

loicgasser:
PS I achieve the same result if I remove the usage of the currently useless parameter backgroundLayers from the topic service.

See:
[TestLink](https://mf-geoadmin3.dev.bgdi.ch/gal_bglayers/?topic=schneesport)

Diff:
https://github.com/geoadmin/mf-geoadmin3/compare/gal_bglayers?expand=1


Why not simply get rid of this parameter, remove code instead of adding some?

gjn:
if we remove this parameters, all topics will have the same backgroundlayers as selection. I'm not sure that's currently the case. Should be checked in the db.
Am 29.01.2016 4:46 nachm. schrieb Gasser Loïc <notifications@github.com>:PS I achieve the same result if I remove the usage of the currently useless parameter backgroundLayers from the topic service.

See:
TestLink

Diff:
https://github.com/geoadmin/mf-geoadmin3/compare/gal_bglayers?expand=1

Why not simply get rid of this parameter, remove code instead of adding some?

gjn:
What about them? They are already applied.
Am 29.01.2016 4:26 nachm. schrieb Gasser Loïc <notifications@github.com>:what about opacity and timestamps?

gjn:
No, here we apply active bsckground and layers visibility/opacity/timestamps.
For sure it could be optimized.
Once bod is cleaned up, we can do that.
Am 29.01.2016 4:17 nachm. schrieb oterral <notifications@github.com>:LGTM but shouldn't we remove what has been added in https://github.com/geoadmin/mf-geoadmin3/pull/3068/files ?

loicgasser:
backgroundLayers always has the same 3 layers for all topics.

gjn:
Yes, you are right. I'll remove this parameter next week.

oterral:
@gjn @loicgasser I'm not against removing this config but I think @Jenselme needs it,  see [old PR](https://github.com/geoadmin/mf-geoadmin3/pull/2440#issuecomment-114081017)

loicgasser:
We could maybe define the default backgrounds in the Makefile then instead and inject it in the `index.html` like we do it for a bunch of other things.

oterral:
We could go far and totally remove the request to [https://mf-geoadmin3.dev.bgdi.ch/ioda-net_dndrop/1454319773/services] and statify this in the build process.

gjn:
topics request is part of the ERSI rest standard and we need to support it in api3. But nothing against statifying it here in kf-geoadmin3.
Am 01.02.2016 11:01 vorm. schrieb oterral <notifications@github.com>:We could go far and totally remove the request to [https://mf-geoadmin3.dev.bgdi.ch/ioda-net_dndrop/1454319773/services] and statify this in the build process.

oterral:
@gjn are you sure the definition of topics is an ESRI REST Service ?  

I don't find anything in [doc](http://resources.arcgis.com/en/help/arcgis-rest-api/#/What_s_new_in_the_ArcGIS_REST_API/02r3000000q6000000/) about this.

oterral:
Whatever I merge this PR because it fixes the bug but I think we should use one way (plConfig) or the other (json object) but both  at the same time.

AFoletti:
I just tested with the link provided and I found a bug:
if I open the bgselector in FF, the options are there and they work as defined, but they all display a swissimage thumbnail
This is confirmed in PROD too, by me and by a webgis email.

oterral:
omg thx fo finding this. I fix.

Jenselme:
@gjn @loicgasser @oterral I indeed need the `backgroundLayers` parameter since as explained in the old PR, we don't have the same background layers on all of our topics. I would appreciate that this parameters stays.

But, if you do remove it, please mention me in the PR so that I know it before our next merge and avoid searching why bg selection is broken ;-)

gjn:
@Jenselme We will keep it. Just for you. :)

gjn:
@oterral @loicgasser Next steps after tomorrows deploy is to get rid of all the unneeded columns in the bod regarding topic configuration and put it into the plConfig column.

Regarding ESRI Rest Services API. I think the root of the service point (/rest/services/) is used to give an overview of the available 'maps'. This is from memory. @loicgasser Can you confirm?

loicgasser:
Yes it is defined as the Catalogues Resource (our so called topics). Of course we just kept the principle and gave it our own geoadmon flavour. :D Here is the official doc: http://resources.arcgis.com/en/help/arcgis-rest-api/#/Catalog/02r3000000tn000000/

oterral:
Thanks for the link. But apart the name there is nothing common in the response.

loicgasser:
Indeed, I wouldn't know what to do with all that,

Jenselme:
@gjn Thanks!

PR #3113, opened 2016-02-02
Remove pixel detection when we get the layers to query

oterral:
[Test](https://mf-geoadmin3.dev.bgdi.ch/teo_tooltip/?lang=fr&topic=schneesport&bgLayer=voidLayer&X=146659.97&Y=693981.58&zoom=3&dev3d=true&debug&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bafu.wrz-wildruhezonen_portal,ch.swisstopo.swisstlm3d-wanderwege,ch.swisstopo.pixelkarte-farbe-pk50.noscale,ch.swisstopo.hangneigung-ueber_30,ch.swisstopo-karto.hangneigung,ch.bafu.wrz-jagdbanngebiete_select,ch.bazl.heliports-gebirgslandeplaetze,ch.swisstopo-karto.schneeschuhrouten,ch.swisstopo-karto.skitouren,ch.swisstopo.skitourenkarte-50.metadata,ch.bav.haltestellen-oev&layers_visibility=false,false,false,false,true,false,true,true,false,true,true,false,false&layers_timestamp=18641231,,,,,,,,,,,,&layers_opacity=1,1,1,1,0.85,0.5,0.2,0.6,1,0.8,0.8,1,1&catalogNodes=2352)

### DIFF
--- a/src/components/tooltip/TooltipDirective.js
+++ b/src/components/tooltip/TooltipDirective.js
@@ -62,33 +62,14 @@ goog.require('ga_topic_service');
           return l;
         };
 
-        // Get all the queryable layers at a pixel
-        var getLayersToQueryAtPixel = function(map, pixel, is3dActive) {
-          var bodIds = [];
+        // Get all the queryable layers
+        var getLayersToQuery = function(map) {
           var layersToQuery = [];
-          if (!is3dActive && !gaBrowserSniffer.ios &&
-              (!gaBrowserSniffer.msie || gaBrowserSniffer.msie > 10)) {
-            // Works only on sublayers
-            map.forEachLayerAtPixel(pixel,
-              function(l) {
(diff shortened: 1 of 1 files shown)


gjn:
Hmm. This is not exactly as it was before.

In this PR, the no-info dialog is always shown when the identify request does not have any results. I think this is disturbing to the user as most clicks will trigger a no-info dialog.

Before, the no-info dialog is only shown when a) identify does not have results and b) the mouse-cursor was a hand at the moment of the click.

I'll merge it and let @davidoesch decide if we need further adaption.

davidoesch:
Thanks works on Int as it should. Happy to discuss with you your proposal to improve the tooltip detection

PR #3125, opened 2016-02-05
Catalog checkbox represents layer visibility

gjn:
@davidoesch As discussed this morning on the phone, the catalog handling has been adapted. If a layer is currently not added to the layer selection, the behaviour is as before: a click on the catalog entry will add the layer to the map.

Once the layer is in the layer selection, the behaviour in the catalog changes. The checkbox represents the layer's visibility. Clicking on the catalog item will toggle the layer visibility.
To remove a layer from the layer selection, it has to be removed in the layer manager.

[Testlink](https://mf-geoadmin3.dev.bgdi.ch/gjn_cat/)

### DIFF
--- a/src/components/catalogtree/CatalogitemDirective.js
+++ b/src/components/catalogtree/CatalogitemDirective.js
@@ -28,6 +28,13 @@ goog.require('ga_layer_metadata_popup_service');
           gaPreviewLayers.removeAll(map);
         };
 
+        var getOlLayer = function(map, item) {
+          if (!item) {
+            return undefined;
+          }
+          return gaMapUtils.getMapOverlayForBodId(map, item.layerBodId);
+        };
+
         return {
           restrict: 'A',
           replace: true,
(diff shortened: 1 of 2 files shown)


davidoesch:
a beauty! this is *exactly* how it should work
-> ok for prod

oterral:
I think the red color is incoherent , we should leave it black.

src/components/catalogtree/CatalogitemDirective.js
```diff
@@ -28,6 +28,13 @@ goog.require('ga_layer_metadata_popup_service');
           gaPreviewLayers.removeAll(map);
         };
 
+        var getOlLayer = function(map, item) {
+          if (!item) {
```
oterral: item is never undefined , you shoudl test item.layerBodId instead

gjn: on initialisation, it is undefined. (at least, i had such an error once.)

gjn:
red color: well, it indicates if a map is added or not in the catalog. @davidoesch 's call.

loicgasser:
+1

oterral:
I agree but it's strange that the color becomes red and when you click again it's still red. it's not coherent with the layer manager. And the red in the  app it's when something wrong happens :)

davidoesch:
"it's strange that the color becomes red and when you click again it's still red."
-> but check box is active/inactive on click. and the layer is active: in the selection window. when you remove the layer in selction window it becomes black

" red as a color"
-> historical : all portal used to have red active layers in CMS. we need to have a relaunch in the near future of the UI to fit the new CMS ( good luck with that...)

gjn:
Tend to agree with @oterral on this. But history wins this one.

And we'll have fun with the new CMS design...

Issue #3129, opened 2016-02-09
Statify layer.json for raster and remove it for vector layers


loicgasser:
Layer.json will be removed for the following layers.

```
                'ch.swisstopo.swisstlm3d-wanderwege',
                'ch.bafu.wrz-wildruhezonen_portal',
                'ch.swisstopo.fixpunkte-agnes',
```

I will be kept and statified in geomadmin for the background layers in WGS84.

```

                'ch.swisstopo.swissimage-product_3d',
                'ch.swisstopo.swisstlm3d-karte-farbe_3d',
                'ch.swisstopo.swisstlm3d-karte-grau_3d'
```

PR #3133, opened 2016-02-10
Restrict all raster layers to terrain availability

loicgasser:
As we can't render anything in 3d if we don't have terrain, I propose to fix this isssue https://github.com/geoadmin/mf-geoadmin3/issues/3129 by using a single layer.json for all raster tiles (e.g. the terrain config).

The ranges in 
'ch.swisstopo.swissimage-product_3d',
'ch.swisstopo.swisstlm3d-karte-farbe_3d',
'ch.swisstopo.swisstlm3d-karte-grau_3d'

are already the same than the terrain one. With this approach we'll have less requests on S3 and no more maintenance of these layer.json file for each raster layer.

What do you guys think? cause I like!!

### DIFF
--- a/src/components/map/MapService.js
+++ b/src/components/map/MapService.js
@@ -940,23 +940,8 @@ goog.require('ga_urlutils_service');
           return $http.get(url).then(function(response) {
             // Live modifications for 3d test
             if (response.data) {
-              // Imagery layers with metadata
-              var ids = [
-                'ch.swisstopo.swisstlm3d-wanderwege',
-                'ch.bafu.wrz-wildruhezonen_portal',
-                'ch.swisstopo.fixpunkte-agnes',
-                'ch.swisstopo.swissimage-product_3d',
-                'ch.swisstopo.swisstlm3d-karte-farbe_3d',
-                'ch.swisstopo.swisstlm3d-karte-grau_3d'
-              ];
-              angular.forEach(ids, function(id) {
-                if (response.data[id]) {
-                  response.data[id].has3dMetadata = true;
-                }
-              });
-
               // Test layers opaque setting
-              ids = [
+              var ids = [
                 'ch.swisstopo.swissimage-product',
                 'ch.swisstopo.pixelkarte-farbe',
                 'ch.swisstopo.pixelkarte-grau',
@@ -1115,6 +1100,9 @@ goog.require('ga_urlutils_service');
               maxLod = gaMapUtils.getLodFromRes(
                   config3d.resolutions[config3d.resolutions.length - 1]);
             }
+
+            var terrainTimestamp = this.getLayerTimestampFromYear(
+                gaGlobalOptions.defaultTerrain, gaTime.get());
             provider = new Cesium.UrlTemplateImageryProvider({
               url: params.url,
               subdomains: params.subdomains,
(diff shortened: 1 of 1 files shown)


loicgasser:
[Test Link](https://mf-geoadmin3.dev.bgdi.ch/gal_layerjson/)

src/components/map/MapService.js
```diff
@@ -1128,10 +1116,9 @@ goog.require('ga_urlutils_service');
               tileHeight: params.tileSize,
               hasAlphaChannel: (format == 'png'),
               availableLevels: window.imageryAvailableLevels,
-              // Experimental
-              metadataUrl: config3d.has3dMetadata ?
-                  getTerrainTileUrl(requestedLayer, timestamp) + '/' :
-                  undefined
+              // Experimental: restrict all rasters to terrain availability
+              metadataUrl: getTerrainTileUrl(
+                  gaGlobalOptions.defaultTerrain, terrainTimestamp) + '/'
```
oterral: What's happening if we remove it totally the `metadataUrl`?

loicgasser: Hundreds of 204 responses.

gjn:
When I open the testlink, add luftfahrthindernisse to the map and switch then to 3D, I have 8 requests to the layers.json configuration. I could understand 5, but not 8. Why?

Also,are there ways  to load it once and apply it to all layers?

oterral:
caused by layer group.

gjn:
Thinking about it: does cesium really do requests to image tiles where we don't have any terrain?

I'll check that.

gjn:
I can't detect any difference in 204 requests with this patch or when we remove the metadataUrl completely. In both cases, we have 204 requests on pixel tiles.

But maybe I'm mistaken.

loicgasser:
You might be right after all. Now, if I recall correctly I ended using the ranges of the background layers to restrict the terrain ranges and not the other way around. (remember the terrain had a wider extent than the background layers which used to cause the black background we wanted to avoid). Maybe removing it entirely might be the right solution. To be confirmed.

loicgasser:
I just checked it I get a ton of 204 without it.


An error occurred in "UrlTemplateImageryProvider": Failed to obtain image tile X: 271 Y: 63 Level: 8.
Cesium.js:30378 An error occurred in "UrlTemplateImageryProvider": Failed to obtain image tile X: 272 Y: 63 Level: 8.
Cesium.js:30378 An error occurred in "UrlTemplateImageryProvider": Failed to obtain image tile X: 269 Y: 63 Level: 8.
Cesium.js:30378 An error occurred in "UrlTemplateImageryProvider": Failed to obtain image tile X: 265 Y: 63 Level: 8.
Cesium.js:30378 An error occurred in "UrlTemplateImageryProvider": Failed to obtain image tile X: 270 Y: 63 Level: 8.
Cesium.js:30378 An error occurred in "UrlTemplateImageryProvider": Failed to obtain image tile X: 264 Y: 63 Level: 8.
Cesium.js:30378 An error occurred in "UrlTemplateImageryProvider": Failed to obtain image tile X: 267 Y: 63 Level: 8.
Cesium.js:30378 An error occurred in "UrlTemplateImageryProvider": Failed to obtain image tile X: 268 Y: 63 Level: 8.
Cesium.js:30378 An error occurred in "UrlTemplateImageryProvider": Failed to obtain image tile X: 263 Y: 63 Level: 8.
Cesium.js:30378 An error occurred in "UrlTemplateImageryProvider": Failed to obtain image tile X: 266 Y: 63 Level: 8.

I stop here...

So I would say keep it as it as Cesium does not restrict the rasters according to the terrain.

loicgasser:
For the request on the layer.json files, I know it's not ideal but this file is in the cache after the first request so it's not so bad even if it's not ideal.

oterral:
+1

Issue #3115, opened 2016-02-03
3D search - marker is not visible after moving


cedricmoullet:
Chrome  48.0.2564.82 on windows 10
Start https://mf-chsdi3.int.bgdi.ch/shorten/69d11c5039
Search for "La Berra"
Select "La Berra (FR) - La Roche"
Marke is not visible in view
![image](https://cloud.githubusercontent.com/assets/319676/12774464/6e47ee32-ca45-11e5-905d-f8548f5c3f31.png)

oterral:
Something strange the marker is displayed but not where the map is zooming.

cedricmoullet:
yes, exactly

cedricmoullet:
related to the new terrain ?

oterral:
Yes probably it seems all zooming after a search is not centered on the marker.

loicgasser:
What about new Cesium?

loicgasser:
I think it's related to the Cesium update:
[See new terrain without Cesium update](https://mf-geoadmin3.dev.bgdi.ch/gal_new_terrain/)

oterral:
Same errors

oterral:
When zooming we use the height service to get the altitude instead of using terrain. I guess the problem  comes from there.

loicgasser:
my browser is a mutant cause I see the point everytime not exactly centered but pretty close.

gjn:
`my browser is a mutant`

Truer words have never been spoken. It's true for any browser.

PR #3135, opened 2016-02-11
Fix flyToAnimation

loicgasser:
https://github.com/geoadmin/mf-geoadmin3/issues/3115

I took a slightly different approach and it seems to work well.

Please check with [TestLink](https://mf-geoadmin3.dev.bgdi.ch/gal_fly_3d/)

### DIFF
--- a/src/components/map/MapService.js
+++ b/src/components/map/MapService.js
@@ -1492,10 +1492,16 @@ goog.require('ga_urlutils_service');
         },
 
         flyToAnimation: function(ol3d, center, destination, defer) {
+          // In degrees
+          var pitch = 50;
+          // Default camera field of view
+          // https://cesiumjs.org/Cesium/Build/Documentation/Camera.html
+          var cameraFieldOfView = 60;
           var scene = ol3d.getCesiumScene();
-          var camera = scene.camera;
           var globe = scene.globe;
           var carto = globe.ellipsoid.cartesianToCartographic(destination);
+          var camera = scene.camera;
+
           $http.get(gaGlobalOptions.apiUrl + '/rest/services/height', {
             params: {
               easting: center[0],
@@ -1503,12 +1509,15 @@ goog.require('ga_urlutils_service');
             }
           }).then(function(response) {
             var height = carto.height - response.data.height;
-            var offset = Math.tan(Cesium.Math.toRadians(50)) * (height + 50);
-            destination.x += offset * 2;
+            var magnitude = Math.tan(
+                Cesium.Math.toRadians(pitch + cameraFieldOfView / 2)) * height;
+            // Approx. direction on x and y (only valid for Swiss extent)
+            destination.x += (7 / 8) * magnitude;
+            destination.y += (1 / 8) * magnitude;
             camera.flyTo({
               destination: destination,
               orientation: {
-                pitch: Cesium.Math.toRadians(-50)
+                pitch: Cesium.Math.toRadians(-pitch)
               },
               complete: defer.resolve,
               cancel: defer.resolve
--- a/src/components/search/SearchTypesDirectives.js
+++ b/src/components/search/SearchTypesDirectives.js
@@ -335,8 +335,8 @@ goog.require('ga_urlutils_service');
               unregisterMove();
               //Gazetteer results that are not points zoom to full bbox extent
               if (res.attrs.origin == 'gazetteer') {
-                isGazetteerPoly = (Math.abs(e[0] - e[2]) > 0.1 &&
-                                   Math.abs(e[1] - e[3]) > 0.1);
+                isGazetteerPoly = (Math.abs(e[0] - e[2]) > 100 &&
+                                   Math.abs(e[1] - e[3]) > 100);
 
               }
               var ol3d = $scope.ol3d;


loicgasser:
The reason why we can't only add the offset on x coordinate using `Cartesian3` or `ECEF` coordinate system.
ecef

src/components/map/MapService.js
```diff
@@ -1492,10 +1492,15 @@ goog.require('ga_urlutils_service');
         },
 
         flyToAnimation: function(ol3d, center, destination, defer) {
+          // In degrees
+          var pitch = 50;
+          // Default camera field of view
+          var cameraFieldOfView = 60;
```
oterral: Why `60` ?

src/components/map/MapService.js
```diff
@@ -1503,12 +1508,15 @@ goog.require('ga_urlutils_service');
             }
           }).then(function(response) {
             var height = carto.height - response.data.height;
-            var offset = Math.tan(Cesium.Math.toRadians(50)) * (height + 50);
-            destination.x += offset * 2;
+            var offsetView = Math.tan(
+                Cesium.Math.toRadians(pitch + cameraFieldOfView / 2)) * height;
+            // Approx. direction on x and y (only valid for Swiss extent)
+            destination.x += offsetView;
+            destination.y += offsetView / 8;
```
oterral: Why `8` ?

src/components/map/MapService.js
```diff
@@ -1503,12 +1508,15 @@ goog.require('ga_urlutils_service');
             }
           }).then(function(response) {
             var height = carto.height - response.data.height;
-            var offset = Math.tan(Cesium.Math.toRadians(50)) * (height + 50);
-            destination.x += offset * 2;
+            var offsetView = Math.tan(
+                Cesium.Math.toRadians(pitch + cameraFieldOfView / 2)) * height;
```
oterral: Why divided by 2 ?

src/components/map/MapService.js
```diff
@@ -1503,12 +1508,15 @@ goog.require('ga_urlutils_service');
...
+            // Approx. direction on x and y (only valid for Swiss extent)
+            destination.x += offsetView;
+            destination.y += offsetView / 8;
             camera.flyTo({
               destination: destination,
               orientation: {
-                pitch: Cesium.Math.toRadians(-50)
+                pitch: Cesium.Math.toRadians(-1 * pitch)
```
oterral: `-pitch` ?

src/components/map/MapService.js
loicgasser: because you want to be in the middle of the camera field view. Need drawing tomorrow.

src/components/map/MapService.js
loicgasser: This is an approximation (like before), you need to move to the cartographic south, but here we are in Cartesian3 earth centered coordinate. In fact the offset view only gives you a magnitude.
Ideally we should have a unit vector in the direction we want and multiply it with our `offsetView` which more of a magnitude to get the translation vector of the positioned at the Nadir (no pitch).

src/components/map/MapService.js
loicgasser: I found it in the doc see comment above

src/components/map/MapService.js
loicgasser: yes because facing north

davidoesch:
Works well on win 7 and Android 6 with chrome

src/components/map/MapService.js
loicgasser: I'll rename to magnitue makes more sense

loicgasser:
I tuned it a tiny bit more and changed the variable name as discussed.

src/components/map/MapService.js
oterral: I juste said you can write `-pitch`instead of ` -1 * pitch`

oterral:
It's still the same for the test case described in #3115 otherwise it's a lot better for others places.

src/components/map/MapService.js
oterral: Can you put the link to the doc please.

loicgasser: https://cesiumjs.org/Cesium/Build/Documentation/Camera.html
In the first example, I ll add it to the code.

src/components/map/MapService.js
loicgasser: ok

loicgasser:
Test case is fixed for me... But I trust that you have other results just hard for to test it.

loicgasser:
It seems to me that we are now always centered. The issue is now more related to the marker that goes under the terrain for a moment and emerges after a second or 2. At least that s what I can observe at La Berra.

loicgasser:
Damn I can reproduce, but La Berra Val-de-Charmey zooms on the correct location... I am a little puzzeled.

loicgasser:
carto.height is evaluated at 105m ..

loicgasser:
Ok I got it gazetter polygons can be super small (10 cm). As a result the zoom to extent function with such small extents in not super happy in 3d. I simply increased this super small threshold in the search and the problem is now solved.

oterral:
lgtm

PR #3137, opened 2016-02-12
Remove git on deploysnapshot

loicgasser:
As we don't need it on int and prod we might as well avoid deploying it.

### DIFF
--- a/scripts/deploydev.sh
+++ b/scripts/deploydev.sh
@@ -47,6 +47,10 @@ done
 if [ $CREATE_SNAPSHOT == 'true' ]; then
   sudo -u deploy deploy -c deploy/deploy.cfg $SNAPSHOTDIR
   echo "Snapshot of branch $GITBRANCH created at $SNAPSHOTDIR"
+  cd $SNAPSHOTDIR
+  git describe --tags --abbrev=0 > .last-release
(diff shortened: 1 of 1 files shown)


gjn:
I like the idea.

Could you try to write the git version of the snapshot in a file? And this could all be done as part of the snapshotting step (in deploydev.sh).

loicgasser:
Good point

we could write the last release with
`git describe --tags --abbrev=0 > .release`

or the last commit reference
`git log -1 --pretty=format:"%h - %an, %ar : %s" > .last-commit-ref`

oterral:
Nice

gjn:
Why not do it in the deploydev after we created the snapshot?

loicgasser:
Because we if remove git we would need a fresh clone every time. How can you git pull/reset/clean if git is not there anymore? also I remember you saying you wanted to keep git in the main dev project.

gjn:
yes, keep it in main always.

But in the snapshot, we can remove it just after we created the snapshot (just after this line: https://github.com/geoadmin/mf-geoadmin3/blob/master/scripts/deploydev.sh#L49) - which could be done in the devploydev directory.

loicgasser:
Ah ok, yes of course we can do it then. I'll adapt.

loicgasser:
updated

PR #3138, opened 2016-02-12
New appcache, never cache it

gjn:
This PR does 2 things to address various potential problems with appcache.

1) It created a new appcache file name. The idea is that old existing appcaches which still tangle in various browsers (desktop and mobile pages) is invalidated by triggering a 404 response code. Becaus the old `geoadmin.appcache` does not exist anymore, this should happen.

2) Assures that appcache file is never cached. Despite having https://github.com/geoadmin/mf-geoadmin3/blob/master/apache/app.mako-dot-conf#L10, I've detected that browsers still cache this file. By putting it in the same block as index.html, we assure that it's never cached and always requested.

@oterral Some issues we have are still potentially caused by bad working appcache updates. I hope this will get rid of them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,6 @@ help:
 	@echo "- testprod         Run the JavaScript tests in prod mode"
 	@echo "- teste2e          Run browserstack tests"
 	@echo "- apache           Configure Apache (restart required)"
-	@echo "- appcache         Update appcache file"
 	@echo "- fixrights        Fix rights in common folder"
 	@echo "- all              All of the above (target to run prior to creating a PR)"
 	@echo "- clean            Remove generated files"
@@ -131,9 +130,6 @@ teste2e: guard-BROWSERSTACK_TARGETURL guard-BROWSERSTACK_USER guard-BROWSERSTACK
 
 .PHONY: apache
 apache: apache/app.conf
-
-.PHONY: appcache
-appcache: cleanappcache prd/geoadmin.appcache prd/index.html prd/mobile.html prd/embed.html
 
 .PHONY: deploydev
 deploydev:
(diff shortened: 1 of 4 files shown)


davidoesch:
Let me know when you plan to deploy this fix: will check with Helpdesk to see if it will reduce the reports of "white page "

Makefile
```diff
@@ -615,6 +615,7 @@ cleanall: clean
 .PHONY: cleanappcache
 cleanappcache:
 	rm -f prd/geoadmin.appcache
```
loicgasser: why leave this rule?

loicgasser:
Good idea, we could maybe use the `VERSION` variable to generate file appcache file names when needed? I guess this is just a test for now.

Makefile
gjn: I want to make sure we don't have this file anywhere. appcache made me paranoid :)

gjn:
@oterral What do you think about the idea to create a new filename for each version? Might be worth a try.

@loicgasser Good input!

Makefile
loicgasser: git clean to the rescue!

oterral:
I'm testing the PR.

oterral:
Sorry guys. This doesn't work. Even if the the `geoadmin.appcache` returns 404, the page will not load the 'geoadmin2.appcache'. The page just do nothing and use what is already in the appcache cache.

oterral:
Forget the previous comment, I forgot to replace a variable. 
Yes it works as expected, if you feel more confident using a different appcache file on each build, no problem with me, it seems to work great.

gjn:
@oterral Can we have this for tomorrow?

oterral:
Yes I can finsih the PR.

oterral:
One more positive  argument with this solution, there shouldn't the auto treload of the page on mobile when there is a new version.

oterral:
Updated, needs review.

I've tested on all major browser and it seems to work. The auto reload is still present when a new file is available.

gjn:
LGTM in general.

Maybe we need to treat more events? http://www.html5rocks.com/en/tutorials/appcache/beginner/ (like error, for example)

oterral:
I don't think it's necessary. we could add it afterwards if needed.

PR #3139, opened 2016-02-12
Cleanup all untracked files and folders

loicgasser:
If a folder name changes in between 2 versions and is not cleaned anymore in the cleanall target, we might end up with unwanted files and folders. (as we are not creating a fresh clone when deploying to dev)

Use: `git clean -n -fxd .` for a dry run

### DIFF
--- a/scripts/deploydev.sh
+++ b/scripts/deploydev.sh
@@ -24,7 +24,7 @@ fi
 cd $BASEDIR/geoadmin
 
 # remove all local changes and get latest GITBRANCH from remote
-git fetch --all && git reset --hard && git checkout $GITBRANCH && git reset --hard origin/$GITBRANCH
+git fetch --all && git reset --hard && git checkout $GITBRANCH && git reset --hard origin/$GITBRANCH && git clean -fxd .
 
 # build the project
 source rc_dev 


gjn:
Why the `-x`? And don't we need `-ff` when we have `-d`?

https://git-scm.com/docs/git-clean

loicgasser:
`-x` is to also remove the files in the `.gitignore` file

I agree that the doc is a little confusing.


(.venv)ltgal@ip-10-220-4-46 (vpc/mf0 - test):~/mf-geoadmin3$ git clean -n -fxd
Would remove .build-artefacts/
Would remove apache/app.conf
Would remove deploy/deploy-branch.cfg
Would remove node_modules/
Would remove prd/
Would remove src/TemplateCacheModule.js
Would remove src/deps.js
Would remove src/embed.html
Would remove src/index.html
Would remove src/lib/angular-translate-loader-static-files.js
Would remove src/lib/angular-translate-loader-static-files.min.js
Would remove src/lib/angular-translate.js
Would remove src/lib/angular-translate.min.js
Would remove src/lib/angular.js
Would remove src/lib/angular.min.js
Would remove src/mobile.html
Would remove src/style/app.css
Would remove test/karma-conf-dev.js
Would remove test/karma-conf-prod.js

Which is what we want.

loicgasser:
`-ff` would be to remove submodules if I understand the doc correctly

oterral:
Anyone remember why the angular files are not versioned ?  It should be, I guess.

loicgasser:
I think it's because they are part of package.json now. You PR. ;)

loicgasser:
Status?

oterral:
Iknow :) All other files are versionned in src/lib/ I think we should also do this for angular files.

gjn:
I think we could/should use this git command in the `make cleanall` target - as it's pratically the same and we mean the same.

gjn:
@loicgasser Could you adapte the make cleanall command?

loicgasser:
Yes

loicgasser:
After giving it some more thoughts I am not sure we want git to clean our projects.
Think about the post-restore hooks and the fact that we remove git when deploying -> not so sure.
I am much happier having it in the deploydev script before we remove the `.git` folder.

gjn:
So only inside snapshot directory?

loicgasser:
Yes, If you really think it s not an issue I ll follow your lead, but I think you also remember the issue we had recently with git/makefile/deploy.

gjn:
i'm ok with sandbox only.

gjn:
I meant snapshot.

gjn:
Can you adapt to only do it in snapshot directory?

gjn:
ping @loicgasser

gjn:
ping @loicgasser

loicgasser:
Makes no sense to me to do it anywhere else than where it is right now.

If you do it here:
https://github.com/geoadmin/mf-geoadmin3/blob/master/scripts/deploydev.sh#L48
It will break the build  because it comes after the `make all`.

If you do it in cleanall it is bad parctice. So I'll close as you seem to be against it.
Feel free to reopen if you think it's

gjn:
I was wrong.

If you stil care, please re-open and merge.

Issue #3092, opened 2016-01-28
GeoAdmin in S3


procrastinatio:
Goal is to run enterly mf-geoadmin3 in AWS S3.

The biggest gain is to have a basic (app + tiles) rock solid application. Further advantages:
- No more varnish flush issue.
- Testing the prod build before deploying
- Activating (or reverting) a new version will be a matter of changing the index.html file
- Deploying only whats really needed (about 36MB) and not everything (about 500MB)

PR https://github.com/geoadmin/mf-geoadmin3/pull/3078 has statifyed both services which needs to be runned under map.geo.admin.ch, but still need an apache rewrite rule, what is not possible in S3.

Service http://api3.geo.admin.ch/rest/services can without problem served as a file

Service http://api3.geo.admin.ch/rest/services/all/MapServer/layersConfig has langage dependend attribute. So either remove then and put them into <lang>.json or adapt the service to use the `lang`parameter int the path or filename.

POC (obvioulsy everythin that need an `admin.ch` referer is not working)

http://dummy-geoadmin.s3-website-eu-west-1.amazonaws.com/

loicgasser:
:+1:

gjn:
YES.

cedricmoullet:
awesome. But please don't put in danger skitouren go live.

2016-01-28 14:22 GMT+01:00 Gilbert Jeiziner <notifications@github.com>:

> YES.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/geoadmin/mf-geoadmin3/issues/3092#issuecomment-176183774>
> .
>



-- 
Twitter: http://twitter.com/cedricmoullet
Linked In: http://www.linkedin.com/in/cedricmoullet

<http://twitter.com/cedricmoullet>

davidoesch:
two enthusiastic thumbs up!!  I woudl even go as far and put them later on into the cdn

oterral:
Done congrats :tada:

gjn:
Yes. We'll put this on cloudfront as well after we gain experience with the tiles.

PR #3142, opened 2016-02-15
[WIP] Deploy and manage geoadmin on AWS S3

procrastinatio:
This PR adds a script to deploy static versions of geoadmin to an AWS bucket. Many versions may live together in the bucket, but only one is the active one (i.e. running at the _root_).

:exclamation:  Important update:  See my comment https://github.com/geoadmin/mf-geoadmin3/pull/3142#issuecomment-232587798

**Currently**, we use a code **snapshot**, which is only a isolated copy of code as basis for deploy. Deploy will **build** geoadmin on every hosts it is deployed to.

With **this PR**, you have to **build** geoadmin with the correct settings, before deploying. Deplyoing does nothing but copying files to S3. Check the _prd/info.json_ file to see the used settings.

See the [README.md](https://github.com/geoadmin/mf-geoadmin3/blob/mom_layersconfig_lang/README.md)

```
{
  "branch":"mom_layersconfig_lang",
  "version": "1456401810",
  "api_url": "//mf-chsdi3.dev.bgdi.ch",
  "last_commit_hash":"3beb581ccb1cd4d1367691843f581665b4c549cb",
  "last_commit_date":"2016-02-25 12:54:56 +0100",
  "user": "ltmom",
  "build_date": "2016-02-25 12:03:30 +0000"
}
```

Addition to `Makefile`
```
- s3upload       Uploads current build to S3, from current dir if SNAPSHOT=xxx is not mentioned
- s3activate     Activates version on S3 with VERSION=xxxxxxxxx
- s3list         Lists uploaded version to S3
- s3delete       Deletes version VERSION=xxxxxx on S3
```

Results
Via **S3**
[active version](https://s3-eu-central-1.amazonaws.com/mf-geoadmin3-prod-frankfurt/)
[active version /src](https://s3-eu-central-1.amazonaws.com/mf-geoadmin3-prod-frankfurt/src)
[random version, index.1458035776.html](https://s3-eu-central-1.amazonaws.com/mf-geoadmin3-prod-frankfurt/index.1458035776.html)

Idem via **cloudfront**

[active version](https://map2.geo.admin.ch/?lang=rm)
[active version (/src) ](https://map2.geo.admin.ch/src?lang=it)

Au final, les vhosts suivant:
- mf-geoadmin3.dev-bgdi.ch  --> inchangé
- mf-geoadmin3.int-bgdi.ch  -->  AWS S3 + CF
- mf-geoadmin3.prod-bgdi.ch  --> AWS S3 + CF

See https://github.com/geoadmin/mf-geoadmin3/issues/3092


procrastinatio:
jenkins test

procrastinatio:
Ready to review.

Some questions:

1. One bucket, is is enough ? Several build may live together, but only one is active.

2. Build labeling. For now it is a timestamp, but do you preferer something for explicit, like '<branch>@<git hash>:<build version>' ?

gjn:
Not sure I fully understand the general approach. Looks like renames before uploading, etc. Why not have fully ready static structure in prd before upload? Then there's not need for a) renames before s3 upload and b) no apache rewrites for dev. IMO, what we have on dev should be equal to what we have on s3 (where possible).

But let's this discuss tomorrow.in detail.

If possible, we should set the pragma: no-chache, remove the etag, etc [see here](http://stackoverflow.com/questions/49547/making-sure-a-web-page-is-not-cached-across-all-browsers)
Also headers. map.geo index.html headers: 
```
Accept-Ranges:bytes
Access-Control-Allow-Headers:x-requested-with, Content-Type, origin, authorization, accept, client-security-token
Access-Control-Allow-Methods:POST, GET, OPTIONS
Access-Control-Allow-Origin:*
Age:0
Cache-Control:no-cache, must-revalidate
Connection:keep-alive
Content-Encoding:gzip
Content-Length:6874
Content-Type:text/html
Date:Thu, 25 Feb 2016 13:06:51 GMT
Expires:Thu, 25 Feb 2016 13:06:51 GMT
Pragma:no-cache
Server:Apache/2.2.22 (Debian)
Via:1.1 varnish-v4
X-Cache:MISS
X-UA-Compatible:IE=Edge
X-Varnish:146250001
```

map2.geo.admin.ch headers:
```
Accept-Ranges:bytes
Cache-Control:no-cache, no-store, max-age=0, must-revalidate
Connection:keep-alive
Content-Encoding:gzip
Content-Length:6937
Content-Type:text/html
Date:Thu, 25 Feb 2016 13:07:13 GMT
ETag:"086e5dc2f7c8de8a683bdcc0b8146ecc"
Expires:Fri, 01 Jan 1990 00:00:00 GMT
Last-Modified:Sat, 20 Feb 2016 12:22:21 GMT
Server:AmazonS3
Via:1.1 62b561fc8f2846ecc9295346a2fcbde3.cloudfront.net (CloudFront)
X-Amz-Cf-Id:3i1i71UyQ4_1dlmDuAhA7DT6--_JYRUHWg_b6zaqVX_-QldwO0B9GQ==
x-amz-meta-vary:Accept-Encoding
X-Cache:Miss from cloudfront
```

Finally, what would be the process on deploy? We build for prod in the snapshot and then we upload to S3? Or do we create snapshot for int and prod and deploy those to int and prod?

procrastinatio:
@gjn AWS S3 is doing a bit of black magic with headers concerning caching. It turns `Prama: no-cache`into `x-amz-meta-pragma: no-cache` and `Etag` is set by S3, and there is no way to remove it.

gjn:
ok. Thanks alot for investigating.

Pragma: no-cache is mainly for older browsers - so I hope it won't lead to many problems.

loicgasser:
@procrastinatio small tip

add your script here:
https://github.com/geoadmin/mf-geoadmin3/blob/mom_layersconfig_lang/Makefile#L5

then
`make autolint`

and to see if there are remaining errors the auto-corrector cannot handle:
`make lintpy`

scripts/s3manage.py
```diff
@@ -0,0 +1,415 @@
+#!/usr/bin/env python
```
loicgasser: We want to use the virtualenv python for that so this is not needed as long as s3manage is executable

scripts/s3manage.py
```diff
@@ -0,0 +1,415 @@
...
+    return object_url
+
+
+def main():
+    if BUCKET_NAME is None:
+        print "Please define the BUCKET_NAME you want to deploy."
+        sys.exit(2)
+    if s3client is None:
+        init_connection()
+    get_url()
```
loicgasser: ?

procrastinatio:
@gjn For **very** old browser indeed, as RFC2616 (HTTP 1.1),   
```Note that HTTP/1.0 caches might not implement Cache-Control and might only implement Pragma: no-cache (see section 14.32).```

File: scripts/s3manage.py (hunk 1 above)
Comment (procrastinatio) [inline scripts/s3manage.py]: Yes, it's why it's `/usr/bin/env python`and not `/usr/bin/python`for instance. No ?

**Comment by procrastinatio** [comment]
jenkins test

**Comment by procrastinatio** [comment]
jenkins test

**Comment by procrastinatio** [comment]
jenkins keeps failing for totaly unrelated reason  :angry:

**Comment by gjn** [comment]
browserstack troubles.

I'll prepare a PR to remove most of browserstack.

File: src/info.mako.json [hunk 3]
```diff
@@ -0,0 +1,10 @@
+{
+  "branch":"${git_branch}",
+  "version": "${version}",
+  "api_url": "${api_url}",
+  "mapproxy_urls": "${mapproxy_url}"
```
loicgasser: missing comma

scripts/s3manage.py
loicgasser: ok so after a `source .../bin/activate`
You can do `scripts/s3manage.py {command}`

In makefile you'r always using ${PYTHON_CMD} I'm not sure which one is used in that case...

gjn:
Changed to WIP because we have to wait for the logging infrastructure.

Also, this needs a rebase.

procrastinatio:
Rebased

gjn:
Let's move forward with this, deploy to S3 and put varnish in front of it.

Then when logging/cloudfront/S3 is solved, we'll remove varnish.

scripts/s3manage.py
```diff
@@ -0,0 +1,568 @@
...
+            'prd/cache/services'),
+        '{}/src/services'.format(VERSION),
+        cached=True,
+        mimetype='application/js')
+
+    # for lang in ('de', 'fr', 'it', 'rm', 'en'):
+    #    save_to_s3(os.path.join(base_dir, 'prd/cache/layersConfig.{}.json'.format(lang)),
+    #               '{}/src/layersConfig.{}.json'.format(VERSION, lang),
+    #               cached=True,
+    #               mimetype='application/js')
```
loicgasser: Why is this part commented?

README.md
```diff
@@ -161,6 +161,67 @@ If you want to use the deploy configuration of directory from which you are exec
...
+- Uploading to S3
+- Activating a version, as only one version may be active at time (`index.html`)
+
+You may deploy a build from your local environnement or a previously build snapshot.
+
+    export BUCKET_NAME=<int/prod aws s3 bucket>
+
+If you want to upload a locally build project:
+
+    make s3upload
```
loicgasser: Please update doc, as s3upload is not a valid target anymore

scripts/s3manage.py
loicgasser: Ok it's above but remove dead code it's confusing.

README.md
```diff
@@ -161,6 +161,67 @@ If you want to use the deploy configuration of directory from which you are exec
...
 Note: we should NOT manually adapt code in /var/www/vhosts/mf-geoadmin3 directory
 
+
+## Building the project to int and prod (S3)
+
+Deploying a project to int or prod consists of three steps:
+
+- Building
+- Uploading to S3
+- Activating a version, as only one version may be active at time (`index.html`)
```
loicgasser: Ok, so only one version at a time is active for dev, int and prod. Now how do we handle branches? If Varnish doesn't point to EC2 instances anymore but directly to S3 bucket, it seems to me that all our branches will be obsolete. I think we need a mechanism to upload branches as well. Or I missed something in the code, in which case can you show me how this can be done in the code.

scripts/s3manage.py
```diff
@@ -0,0 +1,568 @@
...
+                sys.exit(2)
+            # VERSION = SNAPSHOT
+        else:
+            if len(sys.argv) == 4:
+                base_dir = os.path.abspath(sys.argv[3])
+                if not os.path.isdir(base_dir):
+                    print "No code found in directory {}".format(base_dir)
+                    sys.exit(2)
+            else:
+                base_dir = os.getcwd()
```
loicgasser: This is in contradiction with line 484 and is dead code as you will exit if the `len(sys.argv) < 4`

scripts/s3manage.py
```diff
@@ -0,0 +1,568 @@
...
+
+        VERSION = get_index_version(ctx)
+        print "Using version={} from directory={}".format(VERSION, base_dir)
+
+        active_version = get_active_version()
+
+        if VERSION == active_version:
+            msg = "WARNING!!!\nVersion {} is the active one!!!\n" + \
+                "Do you really want to upload it from '{}'?: [y/N]"
+            response = raw_input(msg.format(VERSION, base_dir))
```
loicgasser: Do we really want to let people override an active version? I would tell the user to forget about it and exit.

scripts/s3manage.py
```diff
@@ -0,0 +1,568 @@
...
+    print "      print this message"
+
+
+def upload(version, base_dir):
+
+    epoch_time = int(time.time())
+
+    FILES = ['prd/lib/build.js',
+             'prd/style/app.css',
+             'prd/index.html',
```
loicgasser: we are saving index.html line 226 I think you can remove it from here.

scripts/s3manage.py
```diff
@@ -0,0 +1,568 @@
...
+             'prd/lib/build.js',
+             'prd/img',
+             'prd/style',
+             'prd/lib',
+             'prd/locales',
+             'prd/info.json'
+             ]
+    if UPLOAD_SRC_DIR:
+        FILES.insert(0, 'src')
```
loicgasser: With that statement, we will upload all that is in `src` later on except for `EXCLUDES`? Then you can probably also add `mako` in the EXCLUDES.

README.md
loicgasser: I guess that the versions will now be the snapshot timestamp.
Ideally this should be documented and at the end of the command, it should produce an url we can test.

scripts/s3manage.py
```diff
@@ -0,0 +1,568 @@
...
+    print "      print build info on 'version' (branch, build date, git hash...)"
+    print
+    print "  delete <version>"
+    print "      delete the given 'version' (both directory and indexes files"
+    print
+    print "  help"
+    print "      print this message"
+
+
+def upload(version, base_dir):
```
loicgasser: The whole function is a bit too sensitive I think and we will need to remember to adapt this script if we change the project structure and so on.
Can't we:

1. The mapping between the extension and the mimetype
2. What we want to always exclude
3. What should not be cached.

And then simply `os.walk` through the `prd` directory.

According to the staging `dev` and `int` for instance always set UPLOAD_SRC_DIR to True and if staging is `prod` set it to False.

README.md
gjn: yes, branch deploys are snapshoted as all releases. Graphic in my whiteboard should make it clear.

Varnish will do the magic.

scripts/s3manage.py
```diff
@@ -0,0 +1,568 @@
...
+        print("{} --> {}.html".format(src_key_name, n))
+
+        s3client.copy_object(
+            Bucket=BUCKET_NAME,
+            CopySource=BUCKET_NAME +
+            '/' +
+            src_key_name,
+            Key=n +
+            '.html',
+            ACL='public-read')
```
loicgasser: ok nice, I understand what is being done here. Only move 3 files and the app is deployed. in prod. I would use only one function to do both operations though.

loicgasser: `try` and `except` would be nice as well.

scripts/s3manage.py
```diff
@@ -0,0 +1,568 @@
...
+        try:
+            s3client.copy_object(
+                Bucket=BUCKET_NAME,
+                CopySource=BUCKET_NAME + '/' + src_key_name,
+                Key=os.path.basename(src_key_name),
+                CopySourceIfModifiedSince=datetime(2015, 1, 1),
+                ACL='public-read')
+        except botocore.exceptions.ClientError as e:
+            print "Cannot copy {}: {}".format(j, e)
+
```
loicgasser: Can you explain more in details what are those `Special files`.  The only difference here as far as I can tell is `CopySourceIfModifiedSince=datetime(2015, 1, 1)` which is an hardcoded datetime.

README.md
loicgasser: A graphic in the doc would make it clear forever :)

procrastinatio:
According to the discussion with @gjn, @danduk82 the S3 build will do the following

* We do not use the current snapshot system
* Instead we isolate the code pulled from GitHub and build it in a temp directory, possbily on mf0, according to the follwoing structure:

```
  <branch name>/<git hash>/<build time>/<build target>/<build type>

  <branch name>    branch beining used, e.g. _master_, _my_super_feature_
  <git hash>           git commit short hash (6 characters ?)
  <build time>        or timestamp. Epoch time (note. we sometime rebuild to get
                               a new timestamp (appcache stuff)
  <build target>     _prod_, _int_, __dev_. These are systematically build
  <build type>        Again _release_ and _development_ also build
```
e.g.
```
teo_fix_2848/51ad10/1467875719/prod/src
```
* Once build, the code is frozen (no more changes)
* These code is pushed **as is** on a **single** AWS S3 bucket
* A complexe rewrite with either varnish or *AWS Route63*  will match to above directory (_master_ and _release_ being omited.

e.g. mf-geoadmin.51ad10-1467875719-prod.bgdi.ch/teo_fix_2848/src

* A deploy will also consist in testing the code with the technical adresses, and then switch the public one.

README.md
gjn: http://www.gravizo.com/

Would enable us to integrat small graphs in *.md files, which would directly be shown as graphs on github.

gjn:
I think we can even go a little further: build once prd/src for all targets **once**

e.g. `master/0730ec523/20150725153122/` for prd and `master/0730ec523/20150725153122/src` for src

This would mean to adapt the application to derive links (to API, to itself, etc) from the domain and path where it is served by varnish. If the above is called with `mf-geoadmin3.dev.bgdi.ch`, then api would be corresponding dev api. If the above is called with `map.geo.admin.ch`, then api3.geo.admin.ch will be used. Basically, **we push application configuration from build-time to run-time**. That's also proposed in the book on my table, 'Continouse Delivery` (build once and run it on different environments)

I've started work on this to verify that we cover all grounds (appcache, twitter cards, back-end etc) - but I think it's possible. It's basically some adaptions in index.mako. Especially if we allow re-configuring back-ends via url parameters. 

The big advantage would be that deploying from int to prod only involves re-configuration on varnish level (no copying needed at all)

procrastinatio:
@gjn Yes, it it clear that all targets will be build or rebuild and deployed. We build the tree environments ans deplyo to these tree buckets.

In the end, a smart rewrite maps:

`mf-geodmin3.0730ec523-20150725153122.bgdi.ch` to `master/0730ec523/20150725153122/` in the prod bucket  or 

`mf-geodmin3.0730ec523-20150725153122.bgdi.ch/my_branch/src`  for  `my_branch/0730ec523/20150725153122/src`

My first idea was to use only a timestamp (we are not building that many versions), and using a file to store the build metdata:

```

  "branch":"mom_layersconfig_lang",
  "version": "1467875719",
  "api_url": "//mf-chsdi3.dev.bgdi.ch/ltmom",
  "mapproxy_urls": "//wmts{s}.dev.bgdi.ch"`,
  "last_commit_hash": "adc3e3cc92f7d1308539cf8049c6caf9bc0c47e8",
  "last_commit_date": "2016-07-07 09:06:55 +0200",
  "user": "ltmom",
  "build_date": "2016-07-07 07:18:17 +0000"
}
```

gjn:
My idea is not to do 3 builds for the 3 environments. It's to do 1 build for all environments, in one bucket, in one directory.

Also, why is there a need for something like `mf-geodmin3.0730ec523-20150725153122.bgdi.ch` in varnish? I think it's better to keep it simple:
Root (e.g. mf-geoadmin3.bgdi.ch) would point to active directory master/0730ec523/20150725153122/
Other paths (e.g. mf-geoadmin3.dev.bgdi.ch/blabla/blablablub) would point to corresponding folder in the bucket. Special rules for development is needed.

But varnish could be quite simple, with pseudo code for varnish:
```
if (path == '/' or '')
   point to specific folder in bucket
else if (path == 'lt***') //regexp used
   point to mf0-dev
else
   point to bucket + path
```

A deployment would only change the first if part. The rest does not need adaptions for new builds.

Might be better explained on whiteboard..

gjn:
Please see https://github.com/geoadmin/mf-geoadmin3/pull/3142 for a first version of what I mean regarding builds.

loicgasser:
You linked the PR to itself

procrastinatio:
OK, I see one build, but accessed through different _vhosts_  to point on the corresponding backends (dev, int or prod).

procrastinatio:
Third and hopefully last attempt...

* Single bucket for deployment, on which mf-geoadmin.(dev|int|prod).bgdi.ch are pointing
* A single JS build for all targets (see #3402)
* A single methode `make s3builddeploy` to build and deploy, if test are passing
* Each build are deployed into a single directory base on branch name, git hash and build timestamp
* Switching from one version to another is done on Varnish, e.g
  `map.geo.admin.ch/index.html --> master/0a9e883/1468840423/index.html`

https://mf-geoadmin3.infra.bgdi.ch/master/0a9e883/1468840423/index.html or

https://mf-geoadmin3.infra.bgdi.ch/gjn_build_once/d74ba8d/1468763819/index.html?

loicgasser:
New pr has been open due to heavy rebase conflicts. I close this one.

Issue #2467, opened 2015-07-03
Draw v2


oterral:
Improvement
- [x] Share procedure (How to inform about the share links) ?

Code:

- [x]  ~~Use Mapfish V3 for SVG icons and text's stroke ?~~ v3 migration abandoned
- [x]  Use color service (2nd possibility in https://github.com/geoadmin/mf-geoadmin3/issues/2618)
- [x]  Use at least ol [3.7](https://github.com/openlayers/ol3/releases) for : 
            - https://github.com/openlayers/ol3/pull/3461
            - https://github.com/openlayers/ol3/pull/3635
            - https://github.com/openlayers/ol3/pull/3766
- [x]  Refactoring of the draw events callbacks
- [x]  Refactoring/rename the togglePopup stuff
- [x]  Clean the dropdown toggle hack
- [x]  Verify if `ga-draw-info-save`  and `.ga-draw-disclaimer ` use a CD bund color
- [ ]  Warn the user or us when a file save has failed?
- [ ]  ~~Use a map property in profile ?~~ 
- [ ]  ~~Hide menu when drawing ?~~

New tools:

- [ ]  ~~Rectangle~~
- [ ]  ~~Circle~~

New styles:

- [ ]  ~~Stroke width of polygon and line~~
- [x]  Size of text including a carriage return **(prio1)**
- [x]  Color of icon  **(prio1)**
- [ ]  ~~Color of text background  **(prio1)**~~
- [x]  Symbol & description text shown at the same (checkbox)**(prio1)**

New functionalties:

- [x]  Delete last point
- [ ]  Continue a line / measure after saving
- [ ]  Undo/redo  ?
- [x]  Add a button to delete the current feature selected 
- [ ]  Enter multiple symbols in a row without having to click each time a button (use case: the guy who wants to mark a route with points )
- [ ]  ~~On small screen dropdown menu of  `more` button is hidden~~Menu removed
- [ ]  Set the "Name" of the layer, and use it as kml export name as well
- [ ]  ~~Embed in more button "embed into my site"~~Share stuff moved in modal
- [ ]  Keep the last style used for the next one
- [x] Enhance profile as described in https://github.com/geoadmin/mf-geoadmin3/issues/1912 
- [ ] show length of vector while drawing as in measure (user suggestion)
- [ ] GPX Support as in https://github.com/geoadmin/mf-geoadmin3/issues/2301
- [ ] magnetic drawing based on ve25 : to be confirmed by TOPO
[...]

PR #3149, opened 2016-02-17
Draw v2

### LINKED ISSUE #3417: Text on map in feedback dialog: can't edit text.
When adding a text to the feedback drawing, the text can't be editet. It makes little sense like that. Always  `new text` is displayed.

Should be considered in https://github.com/geoadmin/mf-geoadmin3/pull/3149

### LINKED ISSUE #3421: KML Mapstyle issue
Hi

open fusion table map
https://fusiontables.google.com/DataSource?docid=1E86rvB--SKZYJh87YX2F2zcHunGuGAfSDdrSQhRc#map:id=3 

go to map of geometry

result
unbenannt

then go to 
file 
download
download as kml

import the kml into map.geo.admin.ch
result

unbenannt

Bern does not have the same color class as in the fusion view ...

Question
- is google fusion messing up the export?
- is the <styleUrl> interpreted wrong?

oterral:
enhancements  prio 1 of drawing. 

Fix #3417 
Fix #3421

[Spec](https://docs.google.com/presentation/d/18sa5HyAFUA8DXjBLNJRhXH-zi0g0J6yEVfbns0bsBKg)

[Test](https://mf-geoadmin3.int.bgdi.ch/teo_draw/a6fe903/1472482385/index.html)

- [x] main UI
- [x] Share pop up
- [x] All objects
  - [x] Add  link URL
  - [x] Add image/video/ URL with icon  
  - [x] URL: per default set the text to DE "Mehr dazu..." FR "Plus d'information..." IT "Più informazioni..."  EN "More info ...", RM "Ulteriuras infurmaziuns ..." , no underline in result for link
  - [x] make sure that the image width and iframe width is set to 100% 
- [x] Edit objects (Line) 
  - [x] position with 2'000'000 1'000'000 format
tooltip/edit
- [x] Edit objects (Marker)
  - [x]  Color of icons (prio1)  icon http://fontawesome.io/icon/paint-brush/  
  - [x]  Symbol & text shown at the same (checkbox)(prio1)
  - [x]  order/text as in https://docs.google.com/presentation/d/18sa5HyAFUA8DXjBLNJRhXH-zi0g0J6yEVfbns0bsBKg/edit#slide=id.g10d6700982_0_29 ; label, show label checkbox,text
  - [x] Edit objects (Text)
  - [x]  text including a carriage return (prio1)
  - [x] Size of text 
- [x] Intermediate measure (prio1)
- [x] measure in KML tooltip
- [x] open profile from KML tooltip 

- [x]  change coordsdisplay  to lv95
- [x]  icons as in spec

Profile: in tooltip result for line
- [x] icon [http://fontawesome.io/icon/line-chart/](line-chart)

When done, inform paulo to update the help
- [x] help updated with new screenshots

This PR uses https://github.com/geoadmin/mf-chsdi3/pull/1918


cedricmoullet:
Thanks, nice new functions. 
For the symbols, we have now really a lot of options and I fear that it makes the tool very complex to use.
- The dialog to edit the characteristics of the elements is placed bottom left. It causes trouble with the symbol list since it's not visible when opens. I'm wondering if we shouldn't come back to a very CAD approach like in https://www.scribblemaps.com.
- Regarding colors: couldn't we have more freedom and use something like in Google Docs:
image
- Regarding fonts: my subjective impression is that the text doesn't look very nice on the map. Don't we have an antialiasing problem here ? It's also not very nice in the edit dialog. And could we have the same approach like google docs for the colors ?

davidoesch:
Brainstormed one hour with oterral. Not an easy one. We need here a .... Meeting

mariokeusen:
@davidoesch : did you have a meeting about this? What is the status? Customer requests for that are getting more frequent (also lots of SMO customers)

davidoesch:
on hold until Oli is back from holidays. spec is here https://docs.google.com/presentation/d/18sa5HyAFUA8DXjBLNJRhXH-zi0g0J6yEVfbns0bsBKg/edit#slide=id.p

oterral:
You can start review I've redesigned the draw menu.

davidoesch:
main UI
Share pop up
-> both are ok  IE and Chrome tested

oterral:
I've added the popup close to the feature to modify. Please test.

davidoesch:
- [ ] when in modify mode and move a feature /icon/text: close pop up while moving the feature and open it as in .google.com/mymaps (otherwise you can not move icon to the position under the pop up)
- [ ] icon for color change http://fontawesome.io/icon/paint-brush/ or a bucket

oterral:
@davidoesch for the tooltip we can't change like it's defined in the doc (header modification), because the tooltip contains more than one object.

davidoesch:
If you mean kml tooltip: then we have to brainstorm - how does gmap solve this? Multiple windows will break a lot of things... any other ideas @cedricmoullet ?

oterral:
@davidoesch ready for more testing

Icons are coming.

I see 2 regressions:  

1) a profile can be displayed only for a measure feature (not line anymore)
2) a measure can't be remove with a button, only pressing the Delete key or Delete all button

Are we ok with that or should I fix?

davidoesch:
1) Profile remove from draw : ok with me ( will need explainig but makes sense )-> @cedricmoullet your opinion here?
2) a measure can't be remove with a button, only pressing the Delete key or Delete all button: ok with me. What is not yet in this ticket but often required is: "delete last point" https://github.com/geoadmin/mf-geoadmin3/issues/2467
->  as pulldown in delete? any ideas?

gjn:
1) there's a big wish already to show profile of imported kml lines - something we want to do in the future. So it would make sense to show profiles for all line data. Might need re-discussion of overall design.
2) I think button is required - but I could live not having one for now.

oterral:
I think the delete last point button is not neceessary if we add a correct help tooltip close to the mouse explaing you can remove the point if you click on it.

1) I thought adding a profile button on toolitp of KML feature to show the profile. (like the shop stuff)
2) I can add the same delete button on bottom right of the profile tooltip.

gjn:
For me, those proposals are fine.

1) This would allow to show profile without going into drawing mode, correct?

oterral:
1) yes

gjn:
Go from my side.

davidoesch:
1) profile button on toolitp of KML feature to show the profile.: nice one so we can have it for all lignes even imported ones- > nice
2) delete last point: lets stick to: explain this in help

In general: Size of text change is missing

cedricmoullet:
1) Ok for now. In future, we add an icon in the style dialog for the line profile
2) Can live with that -> we'll see the feedback
One more point: the "style selector" (the dialog that appears when you click on the icon or on the text in the style dialog") is closed when you click outside -> it's for me not intuitive. Couldn't you put a cross on the top right ?

davidoesch:
icons and image and www url is missing according to spec

oterral:
FA icons added.
Close button added.
Text size modification added.

davidoesch:
bug
- [ ] switch to 3d while drawing the back to 2d: information object is not at feature
unbenannt

-[ ] scroll bar covers 2X" in icon selector
unbenannt

oterral:
I've added more help tooltips when modifying (EN and FR). Translations must de filled in the doc

oterral:
@davidoesch you can test the add of link, image or iframe.

davidoesch:
add image/iframe
- [ ] use icon http://fontawesome.io/icon/camera/ not a pulldonw as in spec - use a modal window similar to link /admin link to add link, with ok button, two lines: add image url (img tag) and video (iframe tag)
- [ ] after adding an image/video via above:  show image/video (which is an iframe) directly  in the text window (preview imagenot code..see summary below)

add URL:
- [ ] lets make it simpler: how about : if user paste an url (text string beginning with "http") in the edit window: it converts it to a link ( similar as in maymaps google). alternative: if not: then a modal window two lines with the url text and the per default "More info..." as second line, can be changed... 


to sum it up: how to overcome the issue that we either have to show raw html or interpreted HTML in the edit window - having a raw mode and a preview mode is to complex. I tend to have a raw mode  as oli implemented it

@cedricmoullet your thoughts?

gjn:
Isn't it the idea that in draw mode, you have the raw html - and in non-draw mode, you show the rendered html in the standard tooltip?

davidoesch:
From a coder perspective yes, but from a standard user no (see google my maps or map.search.ch ).

cedricmoullet:
1. Yes, we show the raw html in Edit mode
2. Symbol: current dialog is not intuitive for me. For text, add an empty
textbox. If user adds a text, it appears in the map, otherwise nothing
appears. Checkbox is useless then.
[image: Images intégrées 1]
3. To add iFrame/Hyperlink/Image, I would use the same logic like for
symbology: we add an icon in the bottom right and this pops a dialog with
the following elements:
- HTML element to be added in the description (with a checkbox:
image/link/iFrame)
- xxx URL:
- "Add" button
Something like that:
[image: Images intégrées 2]
We need to find an icon... Rational behind that: it's an expert function
and we don't need to make the main dialog complex.


2016-07-12 20:43 GMT+02:00 Gilbert Jeiziner <notifications@github.com>:

> Isn't it the idea that in draw mode, you have the raw html - and in
> non-draw mode, you show the rendered html in the standard tooltip?
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/geoadmin/mf-geoadmin3/pull/3149#issuecomment-232141055>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AATgvNiNvGPfuG3dnVoc67z0fQtIQ8Jrks5qU-BngaJpZM4HcMNV>
> .
>



-- 

oterral:
From @cedricmoullet:  
2) Good idea. I dont' know why we haven't think that before.

@davidoesch  @cedricmoullet  @gjn 
We could show the preview of the description in the modal window (where we can add links/image/video) so the user can see live what he adds.

davidoesch:
Hi all

- we display only raw
- if someone wants to add iframe code he is a pro: he can add directly the code raw into the window. 
- add video/photo: added the UI and descripition in slide 5 https://docs.google.com/presentation/d/18sa5HyAFUA8DXjBLNJRhXH-zi0g0J6yEVfbns0bsBKg/edit#slide=id.g154e446a6e_0_0 
- add URL: added the UI  in slide 7 https://docs.google.com/presentation/d/18sa5HyAFUA8DXjBLNJRhXH-zi0g0J6yEVfbns0bsBKg/edit#slide=id.g154e637244_0_2 

(Don't talk about "html"in a UI, since in 2016 no one understands it anymore)

cedricmoullet:
Ok for me.

2016-07-13 10:02 GMT+02:00 David Oesch <notifications@github.com>:

> Hi all
>
>    - we display only raw
>    - if someone wants to add iframe code he is a pro: he can add directly
>    the code raw into the window.
>    - add video/photo: added the UI and descripition in slide 5
>    https://docs.google.com/presentation/d/18sa5HyAFUA8DXjBLNJRhXH-zi0g0J6yEVfbns0bsBKg/edit#slide=id.g154e446a6e_0_0
>    - add URL: added the UI in slide 7
>    https://docs.google.com/presentation/d/18sa5HyAFUA8DXjBLNJRhXH-zi0g0J6yEVfbns0bsBKg/edit#slide=id.g154e637244_0_2
>
> (Don't talk about "html"in a UI, since in 2016 no one understands it
> anymore)
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/geoadmin/mf-geoadmin3/pull/3149#issuecomment-232285003>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AATgvNl4tlxdj1Dd7J0hmdsz-DjPBEHhks5qVJuZgaJpZM4HcMNV>
> .
>



-- 

cedricmoullet:
What do we do with slides 7 ? (links)

2016-07-13 10:23 GMT+02:00 Cédric MOULLET <cedric.moullet@gmail.com>:

> Ok for me.
>
> 2016-07-13 10:02 GMT+02:00 David Oesch <notifications@github.com>:
>
>> Hi all
>>
>>    - we display only raw
>>    - if someone wants to add iframe code he is a pro: he can add
>>    directly the code raw into the window.
>>    - add video/photo: added the UI and descripition in slide 5
>>    https://docs.google.com/presentation/d/18sa5HyAFUA8DXjBLNJRhXH-zi0g0J6yEVfbns0bsBKg/edit#slide=id.g154e446a6e_0_0
>>    - add URL: added the UI in slide 7
>>    https://docs.google.com/presentation/d/18sa5HyAFUA8DXjBLNJRhXH-zi0g0J6yEVfbns0bsBKg/edit#slide=id.g154e637244_0_2
>>
>> (Don't talk about "html"in a UI, since in 2016 no one understands it
>> anymore)
>>
>> —
>> You are receiving this because you were mentioned.
>> Reply to this email directly, view it on GitHub
>> <https://github.com/geoadmin/mf-geoadmin3/pull/3149#issuecomment-232285003>,
>> or mute the thread
>> <https://github.com/notifications/unsubscribe/AATgvNl4tlxdj1Dd7J0hmdsz-DjPBEHhks5qVJuZgaJpZM4HcMNV>
>> .
>>
>
>
>
> --
> Twitter: http://twitter.com/cedricmoullet
> Linked In: http://www.linkedin.com/in/cedricmoullet
>
> <http://twitter.com/cedricmoullet>
>



-- 

oterral:
New design added.

@davidoesch can you tell which icons you want (icones on top of description textarea):

http://fontawesome.io/icon/picture-o/
http://fontawesome.io/icon/film/
http://fontawesome.io/icon/link/

or others : http://fontawesome.io/icons/

cedricmoullet:
I really, really like !
It makes sense to put the "description editing tool" at the top of the description field, like in the standard HTML editors.
Regarding icons, i think the 3 you propose make sense.

davidoesch:
add a link
- [ ] change "Text to display:" to "link description"
- [ ] per default set the text to "More info ...", no underline in result for link
- [ ] when "add" is clicked, modal windows closes
- [ ] icon [http://fontawesome.io/icon/link/](url)

add image/video
- [ ] only one button for image /video (as google mymaps): every icon less is better, combine it one with a multiple lines as defined in https://docs.google.com/presentation/d/18sa5HyAFUA8DXjBLNJRhXH-zi0g0J6yEVfbns0bsBKg/edit#slide=id.g154e446a6e_0_0 
- [ ] icon  [http://fontawesome.io/icon/camera/] camera - it fits both: image and  
- [ ] change "Add image:" to " Add Photo or video"
- [ ] change "Image URL:" to "Photo URL"
- [ ] make sure that the image width and iframe width is set to 100%
- [ ] when "add" is clicked, modal windows is closed

Profile: in tooltip result for line
- [ ] icon [http://fontawesome.io/icon/line-chart/](line-chart)

description: Don't agree with 
> It makes sense to put the "description editing tool" at the top of the description field, like in the standard HTML EDITORS

-> the user we target here are not devs. they have never seen an editor. (except word but even there they don't interact) : important things first: the description field not the icons
-> icons are to small on top of the desc field (think mobile and big humbs)
therefore: keep it as in the spec and make the icons in the button bigger 

Question: if I want to edit an existing link/image/video: i either have to delete ALL content incl geometry and start from scratch or try to edit HTML....outch....any ideas to improve this ?

oterral:
@davidoesch I add the A and B markers to measure feature [Test](https://mf-geoadmin3.dev.bgdi.ch/teo_draw/?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&X=108545.89&Y=626075.07&dev3d=true&zoom=0&debug&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege,KML%7C%7Chttps:%2F%2Fpublic.dev.bgdi.ch%2Fvz10_-oJTQ62-h8csD1gEg&layers_visibility=false,false,false,false,true&layers_timestamp=18641231,,,,)

I haven't added the marker to the profile because  I think it's totally useless

davidoesch:
A and B Markers : your are correct--- it is useless, but also in the map.
what request do we want to cover here (BAZL)?
"we send the client a flightroute- from start to landing but how does he know on which side of the line does he need to start? and corresponds with the height below?  from right to left or left to roght"

a possible answer : "use measure (not line) fly from the end with no numbers to the end with the total numbers" " add manually A  and B

options I see:
- add a 0 at the beginning
- add ticks (as wanderland on the routes) with distance number on the line (as measure in in my gmap)

other toughts?

oterral:
I've 2 solutions :  

   - Adding arrows like in this [example](http://openlayers.org/en/master/examples/line-arrows.html). 
Pro: The direction can be more clear.
Cons: Not everyone wants arrows

  - Allow multi feature selection on click. We can  already place a marker or a text snapped on the begininng /end of a line, then, with the multi selection, we could select the common vertexes of  the marker and the line and move it together.
Pros: Fill a lot of usecase (not only bazl usecase)
Cons: Needs a bit more work for the drawer.

What do yout think @davidoesch ?

davidoesch:
For the arrow: if replacing the arrows with km / distance numbers? Like in gmap measure? Possible ?

oterral:
Everything is possible.
But doing like google you will get an overloaded map with tooltips everywhere, which hide map informations.

 The 2nd solution is also far less time-comsuptive.

davidoesch:
For the first one I assume it looks like this
screenshot_20160715-142310

Only for measure



For the second one : 
This would be like in the spec . Without the final total of the distance showed at the end. To see the total of measure as it is now and in the gaps example above makes sense.... So ditching the number for a to.b I'm not convinced

oterral:
@davidoesch @gjn I've added the measure like in Google. [Test](https://mf-geoadmin3.dev.bgdi.ch/teo_draw/?lang=fr&topic=ech&bgLayer=ch.swisstopo.swissimage&X=300580.41&Y=760300.00&dev3d=true&zoom=5&debug&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege&layers_visibility=false,false,false,false&layers_timestamp=18641231,,,)

davidoesch:
I like it... What b do @cedricmoullet  and @gjn think?

A bug: if you close on mobile the profile. Only the points b remain instead of the line.
screenshot_20160719-172533

oterral:
Thx for the bug report.

oterral:
PR ready for review , only some translations are missing.

rc_user
```diff
@@ -1,6 +1,6 @@
 source rc_dev
 export USERNAME=$(whoami)
-export API_URL=//mf-chsdi3.dev.bgdi.ch/$USERNAME
+export API_URL=//mf-chsdi3.dev.bgdi.ch/teo_color
```
loicgasser: please revert

davidoesch:
@oterral let me know when I should do translation

oterral:
@davidoesch  I've forwarded the email on your perso mail

davidoesch:
done

oterral:
@davidoesch some translations are missing line 386 of the doc  and further

gjn:
Remember to self and others: feedback dialog uses drawing elements. Needs testing as well.

davidoesch:
translation done

gjn:
Feedback dialog works fine with my tests.

Anyhow, I detected something else. When I create a drawing with _current_ map.geo admin, I can't edit it on your branch using the admin link. It's simply not visible. Anyhow, the sharing works and I can interact normally with the kml.

This would mean that old drawings can't be edited anymore - which is a no-go in my books. Could you verify and adapt?

apache/app.mako-dot-conf
```diff
@@ -38,6 +38,7 @@ Alias ${apache_base_path}/src ${apache_base_directory}/src
 Alias ${apache_base_path}/ ${apache_base_directory}/prd/
 RewriteRule ${apache_base_path}/src/layersConfig http:${api_url}/rest/services/all/MapServer/layersConfig [P]
 RewriteRule ${apache_base_path}/src/services http:${api_url}/rest/services [P]
+RewriteRule ${apache_base_path}/(?:src\/)?img\/maki/(.*) https:${api_url}/color/255,0,0/$1
```
gjn: This we need to adress for mapOnS3 as well...@loicgasser

src/components/UrlUtilsService.js
```diff
@@ -12,7 +12,7 @@ goog.provide('ga_urlutils_service');
         // from Angular
         // https://github.com/angular/angular.js/blob/master/src/ng/directive/input.js#L3
         var URL_REGEXP =
-            /^(ftp|http|https):\/\/(\w+:{0,1}\w*@)?(\S+)(:[0-9]+)?(\/|\/([\w#!:.?+=&%@!\-\/]))?$/;
+        /^(ftp|http|https):\/\/(?:\w+(?::\w+)?@)?[^\s/]+(?::\d+)?(?:\/[\w#!:.?+=&%@\-/[\]$'()*,;~]*)?$/;
```
gjn: This doesn't look like anything in https://github.com/angular/angular.js/blob/master/src/ng/directive/input.js#L3 anymore. So maybe comment should be removed.

gjn:
When opening the draw, there's a strange artefact:
image

Is it the tooltip div that has no content?

src/components/draw/DrawDirective.js
```diff
@@ -27,889 +28,798 @@ goog.require('ga_permalink');
...
+      var tooltipElement = $document[0].createElement('div');
+      tooltipElement.className = 'ga-draw-help';
+      var tooltip = new ol.Overlay({
+        element: tooltipElement,
+        offset: [15, 15],
+        positioning: 'top-left',
+        stopEvent: false
+      });
+      return tooltip;
+    };
```
gjn: Looks to me like this div is always visible, even when empty. Could we avoid that?

src/components/draw/DrawDirective.js
```diff
@@ -27,889 +28,798 @@ goog.require('ga_permalink');
...
+    var updateSelectHelpTooltip = function(overlay, type) {
+      if (!overlay) {
+        return;
+      }
+      var helpMsgId = 'select_no_feature';
+      if (type) {
+        helpMsgId = 'select_feature_' + type;
+      }
+      overlay.getElement().innerHTML = $translate.instant(helpMsgId);
+    };
```
gjn: The 3 update help tooltip functions could be refactored as all 3 functions contain quite some common code. Minor though.

src/components/draw/DrawDirective.js
```diff
@@ -27,889 +28,798 @@ goog.require('ga_permalink');
...
+        select.on('change:active', function(evt) {
+          var active = evt.target.get(evt.key);
+          if (active) {
+            if (!gaBrowserSniffer.mobile) {
+              deregPointerEvts = map.on([
+                'pointerdown',
+                'pointerup',
+                'pointermove'
+              ], function(evt) {
+                helpTooltip.setPosition(evt.coordinate);
```
gjn: Worth debouncing the setPosition? Or too clunky?

src/components/draw/DrawDirective.js
```diff
@@ -27,889 +28,798 @@ goog.require('ga_permalink');
...
-          // create/update the file on s3
-          ////////////////////////////////////
-          var save = function(evt) {
-            if (layer.getSource().getFeatures().length == 0) {
-              //if no features to save do nothing
-              return;
+        ////////////////////////////////////
+        // Popup content management
+        ////////////////////////////////////
+        var togglePopup = function(feature, clickCoord) {
```
gjn: It's more than toggle functionality now. Maybe the function should be renamed. Minor.

src/components/draw/DrawDirective.js
```diff
@@ -27,889 +28,798 @@ goog.require('ga_permalink');
...
+        // interaction so no need to activate/deactivate it.
+        // Activate/deactivate the select interaction is enough.
+        var modify = new ol.interaction.Modify({
+          features: select.getFeatures(),
+          style: scope.options.selectStyleFunction
+        });
+        modify.on('modifystart', function(evt) {
+          if (evt.mapBrowserPointerEvent.type != 'singleclick') {
+            body.addClass(cssModify);
+            mapDiv.addClass(cssGrabbing);
```
gjn: Why is there need to apply these classes to 2 different html elements?

src/components/UrlUtilsService.js
oterral: Correspond to the 1.4.8 version :  https://github.com/angular/angular.js/blob/v1.4.8/src/ng/directive/input.js#L15

The more recent regex are too permissive I think.

src/components/draw/DrawDirective.js
```diff
@@ -27,889 +28,798 @@ goog.require('ga_permalink');
...
+          // Set the layer to modify if exist, otherwise we use an empty
+          // layer
+          layer = createDefaultLayer(map, useTemporaryLayer);
+          if (!useTemporaryLayer) {
+
+            // If there is a layer loaded from public.admin.ch, we use it for
+            // modification.
+            map.getLayers().forEach(function(item) {
+              if (gaMapUtils.isStoredKmlLayer(item)) {
+                layer = item;
```
gjn: So the last one from public wins? Theoretically, we could have multiple, no?

gjn: To be sure, it's border case...

src/components/draw/DrawDirective.js
oterral: The element is created but not added the html document. It's only added when the OpenLayers Overlay is added to the map.

src/components/draw/DrawDirective.js
```diff
@@ -27,889 +28,798 @@ goog.require('ga_permalink');
...
-            }));
+            // Attach events to the source to update the KML file stored
+            // and to remove overlays when necessary
+            var source = layer.getSource();
+            unSourceEvents = [
+              source.on('addfeature', saveDebounced),
+              source.on('changefeature', saveDebounced),
+              source.on('removefeature', function(evt) {
+                saveDebounced();
+                // Used when the feature is removed outside the draw directive.
```
gjn: How/When can this happen?

src/components/draw/DrawDirective.js
oterral: We can't. With debounce the tooltip doesn't  follow the mouse nicely.

src/components/draw/DrawDirective.js
```diff
@@ -27,889 +28,798 @@ goog.require('ga_permalink');
...
-                    }
-                    updateMeasureTooltips(distTooltip, areaTooltip, geom);
-                  }
-                }
-              });
+          updateHelpTooltip(helpTooltip, tool.id, false);
+
+          if (!gaBrowserSniffer.mobile) {
+            unDrawEvts.push(map.on('pointermove', function(evt) {
+              helpTooltip.setPosition(evt.coordinate);
```
gjn: We have the same event binding in the select change:active function. Not sure we wheter we need it in both places.

If yes, might be factorized.

src/components/draw/DrawDirective.js
```diff
@@ -27,889 +28,798 @@ goog.require('ga_permalink');
...
-              deactivateTool(lastActiveTool);
-              scope.$applyAsync();
+          draw = new ol.interaction.Draw(tool.drawOptions);
+          var isFinishOnFirstPoint;
+          unDrawEvts.push(draw.on('drawstart', function(evt) {
+            // Remove the first features created if wanted
+            var src = layer.getSource();
+            if (tool.maxFeatures &&
+                src.getFeatures().length >= tool.maxFeatures) {
+              src.removeFeature(src.getFeatures()[0]);
```
gjn: Why is this needed?

src/components/draw/DrawDirective.js
```diff
@@ -27,889 +28,798 @@ goog.require('ga_permalink');
...
+            ol.Observable.unByKey(deregFeatureChange);
+            deactivateTool(lastActiveTool);
+            scope.$applyAsync();
+
+            var featureToAdd = evt.feature;
+            var geom = featureToAdd.getGeometry();
+
+            // According to #3319, it seems LineString can be created with one
+            // point (or with an array of exact same points). If it's the case
+            // we ignore it.
```
gjn: This could be factored into a dedicated function, which would also allow to bail out earlier (as soon as non-unique coords are detected)

src/components/draw/DrawDirective.js
```diff
@@ -27,889 +28,798 @@ goog.require('ga_permalink');
...
-              var styles = tool.style(featureToAdd);
-              featureToAdd.setStyle(styles);
-              select.getFeatures().push(featureToAdd);
-
-              // Add final measure tooltips
+            if (geom instanceof ol.geom.Polygon && !isFinishOnFirstPoint) {
+              // The sketchFeatureArea is automatically closed by the draw
+              // interaction even if the user has finished drawing on the
+              // last point. So we remove the useless coordinates.
+              var lineCoords = featureToAdd.getGeometry().getCoordinates()[0];
```
gjn: Minor: could use `geom` here.

src/components/draw/DrawDirective.js
oterral: The cssModify needs to be apply on all the application to hide the popup when we move the feature.

 And the grab cursor css is applying only on the map div to override the previous map cursor css.  

cssGrab, cssGrabbing and cssPointer are only map's cursor stuff I don't think it's necessary to add it to the body.

src/components/draw/DrawDirective.js
```diff
@@ -27,889 +28,798 @@ goog.require('ga_permalink');
...
-              activate();
-            } else {
-              deactivate();
+            if (lastActiveTool) {
+              featureToAdd.set('type', lastActiveTool.id);
             }
-          });
+            featureToAdd.getGeometry().set('altitudeMode', 'clampToGround');
 
+            // Set the definitve style of the feature
```
gjn: typo

src/components/draw/DrawDirective.js
oterral: I think we define in the past that only one file can be modified.

src/components/UrlUtilsService.js
gjn: Yes, we filter out more. So I think its better.

src/components/draw/DrawDirective.js
gjn: ok. I'm fine with that.

src/components/draw/DrawDirective.js
gjn: ok.

src/components/draw/DrawDirective.js
gjn: Makes sense. But this means we can have multiple public.geo kmls on the map, and when we go into drawing, we'll create a new one with all the elements of all the other drawings.

To me, this makes sense - but I'm not sure it's wanted behaviour.

src/components/draw/DrawDirective.js
oterral: When we remove the feature from the trash button in the popup.

src/components/draw/DrawDirective.js
```diff
@@ -27,889 +28,798 @@ goog.require('ga_permalink');
...
+          })
+        ];
 
-          // Remove interactions on destroy
-          scope.$on('$destroy', function() {
-            deactivate();
+        // Remove interactions on destroy
+        scope.$on('$destroy', function() {
+          deactivate();
+          dereg.forEach(function(item) {
```
gjn: dereg only ever contaions one element. So we could use it directly, instead of putting it in an array. Minor.

src/components/draw/DrawDirective.js
```diff
@@ -27,889 +28,798 @@ goog.require('ga_permalink');
...
+          if (mapDiv.hasClass(cssGrabbing)) {
+            mapDiv.removeClass(cssGrab);
+            return;
+          }
+          var hoverSelectableFeature = false;
+          var hoverNewVertex = false;
+          var hoverVertex = false;
+          var selectableFeat, newVertexFeat;
+          var type;
+          var classes = cssPointer + ' ' + cssGrab;
```
gjn: This variable is not used.

src/components/draw/DrawDirective.js
oterral: It was like this before but I'm not sure why this option `maxFeatures` has been added. I don't find a tool which uses that.

src/components/draw/DrawDirective.js
```diff
@@ -27,889 +28,798 @@ goog.require('ga_permalink');
...
-            dropdown.css('left', bt.offset().left -
-                (dropdown.outerWidth() - bt.outerWidth()) + 'px');
-          });
-        }
-      };
+        ////////////////////////////////////
+        // Utils functions
+        ////////////////////////////////////
+        // Change cursor style on mouse move, only on desktop
+        var updateCursorStyle = function(evt) {
```
gjn: Function could be renamed. It's a little bit more than just Cursor Style adaptions.

src/components/draw/DrawDirective.js
```diff
@@ -27,889 +28,798 @@ goog.require('ga_permalink');
...
+                  (itemLayer.getStyle &&
+                  itemLayer.getStyle() == scope.options.selectStyleFunction);
+            }
+          );
+          if (selectableFeat) {
+            // Get the type of the feature
+            type = selectableFeat.getId().split('_')[0];
+
+            if (newVertexFeat) {
+              // We try to found if the newVertex is snapped on an existing
```
gjn: `We try to find out if....`

src/components/draw/DrawDirective.js
gjn: I think we can remove it as maxFeatures is never set anyway.

src/components/draw/DrawDirective.js
oterral: hmm no. The last public.geo kml layer   is modifiable, the others stay the same.

gjn: Hmm. I have some strange effects mixing kml's. Maybe best shown 1-on-1

apache/app.mako-dot-conf
loicgasser: For old KML only

oterral: I can replace the urls when we parse the KML instead

gjn: No, new kml would not work on s3. 

The reason it's in apache like this is offline, no? If so, replacing won't work, no?

oterral: It's not needed for appcache , appcache doesn't save maki icons.

I will change the urls when parsing.

src/components/draw/DrawDirective.js
oterral: I agree but the helpMsgId are tooltip dependent so it's hard to factorize .

src/components/draw/DrawDirective.js
oterral: I could factorize but it will add 2 pointermove events on select mode (move the tooltip and update the cursor) and one in draw mode (move the tooltip). With this small duplication of code we add only one pointermove event in each mode.

gjn: ok. let's keep it that way.

oterral:
Things to do : 

- [ ] delete last point
- [ ] change old maki urls during parsing
- [ ] fix feedback text

davidoesch:
a small correction for the support of external kml
for polygons: use the square and square-o fontawesome

oterral:
@gjn modifications ready, include the delete of last point

gjn:
Great. I'll continue my review over the weekend.

oterral:
> for polygons: use the square and square-o fontawesome

@davidoesch in fact I don't think it's a good idea to use these icons because it's looks like a checkbox.

davidoesch:
You got a point there... Leave it as it is?

oterral:
Yes I think.

oterral:
@davidoesch  profile butto replace by a link. And help for 'delete last point' added. I sent you an email for the translation.

davidoesch:
@oterral if there is a description: why not add it to the south of the icon so it does not cover the icon
image

mariokeusen:
I could finally play around a bit with the current new draw:
I really like it and it works properly imo. Only found one little issue (i tested with IE11)
The "link" window does not close if I click the x-button. 
draw2_remove_link_window
Apart from that: great work!

oterral:
@mariokeusen bug fixed , can you confirm please. https://mf-geoadmin3.int.bgdi.ch/teo_draw/51e8929/1472460646/index.html?topic=ech&lang=fr&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege&layers_visibility=false,false,false,false&layers_timestamp=18641231,,,

davidoesch:
marios bug avove: tested : ok
what is missing
@oterral if there is a description: why not add it to the south of the icon so it does not cover the icon
image

oterral:
@davidoesch Could we do this in another PR ?

oterral:
I you add the text to the south you hide it behind the popup.

davidoesch:
Let's do This in another PR

gjn:
Impressive work, well composed and designed.

I couldn't review everything in detail - but let's go with this. We need to test quite well during deploy, especially IE and safari.

Issue #3153, opened 2016-02-18
[topic change] wildruhezonen-layer is missing in wildruhezonen topic when coming from ech


mariokeusen:
UC: open app with standard ech catalog, look at the layer selection, change topic to wildruhezonen
Result: in ech the layer Wildruhezonen is present in the layer selector (but not loaded), after the change of the topic this layer has disappeared and is NOT loaded to the map (as necessary on the start of the topic=wildruhezonen)
Expected: When switching to the Wildruhezonen-topic, both layers Wildruhezonen and Wildschutzgebiete must be loaded to the map
![bugontopicchangewrz](https://cloud.githubusercontent.com/assets/4577749/13135278/7fac1e20-d60f-11e5-9e12-5b80279db598.PNG)

Remark: if you start the other way round (with map.wildruhezonen.admin.ch), it's the same problem: when you switch to ECH, the layer Wildruhezonen disappears from the layer selector. I suppose this is a general issue, in case you have the same layer in the layer selector of the previous and of the next topic on topic change.

mariokeusen:
In case of Wildruhezonen this is a pretty severe one, therefore prio 1

mariokeusen:
and this is probably related to https://github.com/geoadmin/mf-geoadmin3/issues/3122

loicgasser:
```json
{
defaultBackground: "ch.swisstopo.pixelkarte-farbe",
plConfig: null,
selectedLayers: [
"ch.bafu.wrz-wildruhezonen_portal",
"ch.bafu.wrz-jagdbanngebiete_select"
],
backgroundLayers: [
"ch.swisstopo.swissimage",
"ch.swisstopo.pixelkarte-farbe",
"ch.swisstopo.pixelkarte-grau"
],
langs: "de,fr,it,en,rm",
activatedLayers: [ ],
id: "wildruhezonen"
},
```

You can see that these layers should be indeed both selected per default from the service.

gjn:
Agree, prio 1. I'm having a look.

gjn:
on-garanteed handling of the same event. on topic change, we now do alot of things, among others A) remove all layers, as part of #3122 and B) add default topic layers.

It's just that the order of execution of A) and B) is not garantueed.

Fix is coming and will be deployed as part of sonnendach release this afternoon.

mariokeusen:
Ok, thanks for the quick fix!

gjn:
It's on prod.

Issue #3122, opened 2016-02-05
on topic change : reset layer selection


davidoesch:
on topic change : reset layer selection - do not copy existing active layers.

as discussed.

loicgasser:
+1 it's too messy with preselected layers now.

mariokeusen:
+1 had another customer complaining about this topic-change behaviour

gjn:
Also remove wms/kml layers?

davidoesch:
IMHO yes. This is restart of the  site. clean slate. user can always go back with the browser history

gjn:
No, go back with browser history does not work. We have stateless SAP page - no browser history of state.

Do we keep a drawing?

davidoesch:
Browser back: topic change works. Drawing deletion as well : clean slate.

gjn:
`Browser back: topic change works.` ? which browser?

gjn:
ah, i think you mean the topic reset by clicking on the bund logo, right? yes, this trigger a new page load.

But the topic switch via the topic menu does not trigger a page load and therefore there's no 'go back' from browser history to previous topic.

gjn:
[Testlink](https://mf-geoadmin3.dev.bgdi.ch/gjn_topicchange/)

See #3145 for PR.

davidoesch:
@mariokeusen can you test as well? On mobile it works as discussed.

mariokeusen:
Works fine for me (tested on chrome). On this occasion I would also update the top-layers in the default-ech-catalog according to the stats from 2015:
![toplayers15](https://cloud.githubusercontent.com/assets/4577749/13101183/923e394a-d544-11e5-9207-caa426c9c422.PNG)

Not sure if it's really helpful to take the tlm-map, because it's "half" a bg-layer. I'd go with the following top 3 "data" layers:

- wanderwege
- zeitreihen
- swissboundaries-land

What do you think @davidoesch @gjn ?

davidoesch:
Carefull with stats. One step after the other- first we see how we solved the related tickets. Then we go ahead with a global solution for all topics

mariokeusen:
Ok, I'd drop those proposal layers anyway. But I know: usability....

AFoletti:
It's minor, but when you change the topic there are some "leftovers" of the old one in the URL. Namely layers_visibility and layers_timestamp when switching away from eCH

PR #3154, opened 2016-02-18
Assure right order of execution on topic change

gjn:
Fix for https://github.com/geoadmin/mf-geoadmin3/issues/3153

Will be part of this afternoon's deploy for sonnendach.

### DIFF
--- a/src/components/layermanager/LayermanagerDirective.js
+++ b/src/components/layermanager/LayermanagerDirective.js
@@ -278,6 +278,7 @@ goog.require('ga_urlutils_service');
           scope.filteredLayers.forEach(function(l) {
             scope.removeLayer(l);
           });
+          $rootScope.$broadcast('gaPostTopicChange', newTopic);
         });
 
         // Change layers label when topic changes
--- a/src/components/map/MapService.js
+++ b/src/components/map/MapService.js
@@ -2238,7 +2238,7 @@ goog.require('ga_urlutils_service');
           gaTime.allowStatusUpdate = true;
           registerLayersPermalink(scope, map);
           // Listen for next topic change events
-          $rootScope.$on('gaTopicChange', addTopicSelectedLayers);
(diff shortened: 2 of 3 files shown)


gjn:
This is maybe not the correct fix, but it's working.

Probably more correct would be to not remove the new topic layers in the layer manager, but that would mean first refactoring [this code](https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/map/MapService.js#L2076) to be usable in more than one location.

loicgasser:
issue with some prio?

gjn:
The refactoring? I'll take care of it.

oterral:
I think there is general problem. We should keep in mind this should work with only the map and the gaTopic component (without poermlink stuff).

And yes the LayerManager shouldn't  listen for the gaTopicChange event.

gjn:
Fully agree. It's architectural problem.

`gaTopicChange` should never be listened to by independant components that have nothing to do with topic changes. Only application controllers should listen to this event and dispatch them to more component specific messages/events. Application logic (if in A) happens this, then B) should react as follows and C) in this way) should be outside of components.

So in this case, the application controller should listen to `gaTopicchange` and dispatch a (new) `removeAllLayers` message to the layermanager component.

We have other such violations in our code. We should try to address them.

Also, mapservice file is too big now. Should be split up into components proper and application specific controllers/code.

oterral:
Yes we should split the MapService.

> So in this case, the application controller should listen to gaTopicchange and dispatch a (new) removeAllLayers message to the layermanager component.
No need to dispatch, at least to the layermanager because it is automatically updated when a layer is added or removed from the map, it doesn't care if the topic change or not.

gjn:
It's because of this requirement: https://github.com/geoadmin/mf-geoadmin3/issues/3122 . Application state needs to be initialized. But I agree, this shouldn't be done in the layer manager, but in a application controller.

PR #3164, opened 2016-02-22
[ON HOLD] Fix #3146: Send the geojson feature to the parent window

### LINKED ISSUE #3146: Post message of kml id/name on click
Currently, in iframe mode, we post a message containing the id of the clicked element to the parent page.

This functionality should be extended. When a kml is integrated in the page and the user clicks on an placement of the kml, the name/id of the clicked element should be send.

@oterral Possible? Effort?

oterral:
Fix #3146

### DIFF
--- a/src/components/tooltip/TooltipDirective.js
+++ b/src/components/tooltip/TooltipDirective.js
@@ -493,7 +493,8 @@ goog.require('ga_topic_service');
                 '</div>';
               var name = feature.get('name');
               var featureId = feature.getId();
-              var layerId = feature.get('layerId') || layer.get('bodId');
+              var layerId = feature.get('layerId') || layer.get('bodId') ||
+                  layer.getId();
               var id = layerId + '#' + featureId;
               htmlpopup = htmlpopup.
                   replace('{{id}}', id).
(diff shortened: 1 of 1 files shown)


gjn:
Does this change the message send in any way? Because we have hydrodaten website that already uses this, which would need adaption.

oterral:
Yes that changes (we pass the entire  feature with GeoJSON format).  Hydrodaten website owner must be warned

gjn:
Can we have old and new in parallel? So we can deploy both without risk to have hydrodaten broken?

This should be a cearly defined interface that doesn't change often. It will be used by multiple clients in the future.

Do you think current state is ok going forward?

oterral:
I think sending the feature with the geometry, properties and htmlpopup content, we fill 90% of usecases.

Maybe we can send an object with a feature propertiy in case we need to add more values like the layer or click informations.

oterral:
Living with the 2 in parallel seems complicated. I don't think we can.

gjn:
I think https://github.com/geoadmin/mf-geoadmin3/issues/3146 is enough for the use case we want. And it doesn't break the GeoJSON. So I close this one.

PR #3174, opened 2016-02-25
Avoid using git once snapshot is created

loicgasser:
Using the same approach as in https://github.com/geoadmin/mf-chsdi3/pull/1920/files
When creating a snapshot while deploying to dev we currently remove `.git` folder which is around 100Mb (800 Mb for chsdi).

The problem is that in https://github.com/geoadmin/mf-geoadmin3/blob/master/deploy/hooks/post-restore-code#L21
we use `make all` and the related git branch is false. This fixes this issue. I don't know how to test that other than merging it.
PR coming for chsdi.


loicgasser:
wait just saw something

loicgasser:
ok now

PR #3175, opened 2016-02-26
Allow execution of all saucelab tests in one browser/platform

gjn:
Running all Saucelabs tests on all browsers takes 25 minutes. For Pull Request builds, this is probably
too long.

This PR adds the possibility to execute all saucelabs tests on a single environment (browser and platform). It's meant to be used on the Pull Request builder on Jenkins only, but might be handy for manually exectued checks as well.

Jenkins is already adapted to use this. So quick review/merge would be great.


gjn:
Execution time of jenkins Pull Request job is 7 minutes now instead 29 minutes.

On main (job chsdi), which is run nighlty, we still execute everything (29 minutes).

PR #3176, opened 2016-02-26
Include version in feedback as parameter

gjn:
This is the front-end solution for https://github.com/geoadmin/mf-chsdi3/pull/1948. The PR in chsdi3 should be merged first.

It's the correct handling of what we did in https://github.com/geoadmin/mf-geoadmin3/pull/3090/files by putting the version not in the feedback message, but as a paremeter to the request.


gjn:
@oterral Care for a quick review? Needed for tomorrow.

PR #2917, opened 2015-11-18
Manage auto reprojection for external wms

mariavkourouni:
Fix for https://github.com/geoadmin/mf-geoadmin3/issues/2773

[Test](http://mf-geoadmin3.dev.bgdi.ch/kom_wms2/?dev3d=true)

here there are test urls

eurosoil: http://eusoils.jrc.ec.europa.eu/wms/wms.htm 
- [x] http://eusoils.jrc.ec.europa.eu/wrb/wms_Threats.asp?&SERVICE=WMS&VERSION=1.1.1&REQUEST=GetMap&LAYERS=PESERA&STYLES=&SRS=EPSG:3035&BBOX=1988372,1400000,6411627,5400000&FORMAT=image/png&WIDTH=1200&HEIGHT=900

eea http://www.eea.europa.eu/code/gis 
- [x]  http://discomap.eea.europa.eu/arcgis/services/Land/CLC2000_Dyna_WM/MapServer/WMSServer?request=GetCapabilities&service=WMS

BadenWürtemberg http://rips-dienste.lubw.baden-wuerttemberg.de/rips/ripsservices/apps/uis/metadaten/suche.aspx?suche=infoMapAccessService 

- [x]  http://rips-gdi.lubw.baden-wuerttemberg.de/arcgis/services/GDI/GDI_0100001500100003/MapServer/WMSServer?REQUEST=GetCapabilities&SERVICE=WMS  --> parsed but empty, so not included?

OSM http://wiki.openstreetmap.org/wiki/WMS#OSM_WMS_Servers
- [x] http://wiki.openstreetmap.org/wiki/WMS#OSM_WMS_Servers --> instead of this: "http://ows.terrestris.de/osm/service" & "http://ows.terrestris.de/osm-gray/service"

NASA realtime satellite products
 - [x] http://neowms.sci.gsfc.nasa.gov/wms/wms?request=GetCapabilities --> parsed and not empty, but it seems that there is no information for Switzerland, so not included?

EUMETSAT realtime satellite products
- [x] http://eumetview.eumetsat.int/geoserver/wms?service=wms&version=1.3.0&request=GetCapabilities

Issue #2414, opened 2015-06-11
Import: Redesign


davidoesch:
based on the design/usability feedback

One Import widget
- [ ] See geocat feed #3023
online ressources
- [ ] one TAB for ONLINE sources
  - [x] WMS 
    - [ ] WMS with Auth (manage permalink -> no pwd in permalink / ask user/pwd at start)
    - [x] GetFeatureInfo with TXT format
  - [x] WMTS
    - [ ] WMTS with Auth (manage permalink -> no pwd in permalink / ask user/pwd at start)
      Pulldown content
  - [ ] provide a github based config file to add, change and delete and translate title for the URL Links, based on a google spreadsheet based approach similar as for the translation.Variables see https://docs.google.com/spreadsheets/d/1Lg3BrJ_FHaUddFAL7-QzeJrxQ2lLwRMgLqImyugariA/edit?usp=sharing see as well geocat https://github.com/geoadmin/mf-geoadmin3/issues/2414
    - source id (eg 1)
    - URL to getCap ( eg http://wms.geo.admin.ch/?REQUEST=GetCapabilities&SERVICE=WMS&VERSION=1.0.0&lang=de)
    - Copyright Text ( Geoportal Bund)
    - Copyright URL ( www.geo.admin.ch)
    - DE (eg. wms.geo.admin.ch)
    - FR (eg. wms.geo.admin.ch)
    - IT (eg. wms.geo.admin.ch)
    - EN (eg. wms.geo.admin.ch)
  - [x] add existing WMS pulldown content
  - [x] WM(T)S with other projections (#2773)
    - [x] DE BW_ according to list in  http://rips-dienste.lubw.baden-wuerttemberg.de/rips/ripsservices/apps/uis/metadaten/suche.aspx?suche=infoMapAccessService   - DE BW http://www.webatlasde.de/arcgis/services/Maps4BW/MapServer/WMSServer?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetCapabilities
    - [x] DE BAyern: according to list in http://vermessung.bayern.de/opendata 
    - [ ] FR Cadastre: according to https://www.cadastre.gouv.fr/scpc/pdf/Guide_WMS_fr.pdf for all border communes http://inspire.cadastre.gouv.fr/scpc/[codeINSEE].wms?service=WMS&request=GetCapabilities&
      replace [codeINSEE] according to http://www.insee.fr/fr/methodes/nomenclatures/cog/ 
[...]

PR #3180, opened 2016-03-01
Prevent WMS import zoom to extent to go outside the portal

Jenselme:
More or less the same patch that I did for KML.

### DIFF
--- a/src/components/importwms/ImportWmsDirective.js
+++ b/src/components/importwms/ImportWmsDirective.js
@@ -283,7 +283,7 @@ goog.require('ga_urlutils_service');
     };
   });
 
-  module.directive('gaImportWmsItem', function($compile) {
+  module.directive('gaImportWmsItem', function($compile, gaMapUtils) {
 
     /**** UTILS functions ****/
     // from OL2
@@ -301,7 +301,7 @@ goog.require('ga_urlutils_service');
 
     // Zoom to layer extent
     var zoomToLayerExtent = function(layer, map) {
-      var extent = layer.extent;
+      var extent = gaMapUtils.intersectWithDefaultExtent(layer.extent);
       var view = map.getView();
       var mapSize = map.getSize();
 
@@ -325,11 +325,14 @@ goog.require('ga_urlutils_service');
             layerExtentCenter[0] + width / 2,
             layerExtentCenter[1] + height / 2
           ];
+          extent = gaMapUtils.intersectWithDefaultExtent(extent);
 
-          var res = view.constrainResolution(view.getResolutionForExtent(extent,
-              mapSize), 0, -1);
-          view.setCenter(layerExtentCenter);
-          view.setResolution(res);
+          if (extent) {
+            var res = view.constrainResolution(
+                view.getResolutionForExtent(extent, mapSize), 0, -1);
+            view.setCenter(layerExtentCenter);
+            view.setResolution(res);
+          }
           return;
         }
       }


oterral:
DO you have an example of WMS ? 

I thought we had fix this with modifying this [line](https://github.com/ioda-net/mf-geoadmin3/blob/prevent-wms-extent-outside-portal/src/components/importwms/ImportWmsDirective.js#L246)

Jenselme:
> I thought we had fix this with modifying this line

I have this line on my code and the problem with imported WMS extent. My problems occurs when I click on the WMS tool. See this screenshot: http://jenselme.perso.centrale-marseille.fr/visible/wms_import_extent.png

> DO you have an example of WMS ? 

In our portal I use http://www.geoservice.apps.be.ch/geoservice/services/a4p/a4p_basiswms_d_fk_s/MapServer/WMSServer? but it works correctly on map.geo.admin.ch since its extent is part of Switzerland. Sadly, I don't have a WMS to reproduce it on map.geo.admin. Maybe a French one ?

oterral:
I think i can test, we have tested with an american wms one time.

What is the projection of your map ? and the extent of this projection? any chance to have temporary link where I can test?

oterral:
If the extent is outside the projection extent, the layer has its property `extent` to undefined and so the zoom button shouldn't appear.

Jenselme:
You can reproduce the behavior on https://map.geojb.ch Import http://www.geoservice.apps.be.ch/geoservice/services/a4p/a4p_basiswms_d_fk_s/MapServer/WMSServer?, layer UP5.

I think that part of the problem is that our portal uses the swiss coordinate (epsg 21781) but is restricted to a region of Swisstzerland. That would explain why extent is not undefined and why I need to use `intersectWithDefaultExtent`. Here are the relevant values:
- epsg = "EPSG:21781"
- epsg_extent = [420000, 30000, 900000, 350000]
- portal extent = [550000, 210000, 620000, 245000]

oterral:
Ok I see but I'll prefer to keep the extent logic in the `getLayerExtentFromGetCap ` function , I think that makes sense if [here](https://github.com/ioda-net/mf-geoadmin3/blob/prevent-wms-extent-outside-portal/src/components/importwms/ImportWmsDirective.js#L244) you replace `proj.getExtent()` by the ` gaGlobalOptions.defautlExtent`  . What do you think ?

Jenselme:
Except that in our case, we exit the function [here](https://github.com/ioda-net/mf-geoadmin3/blob/prevent-wms-extent-outside-portal/src/components/importwms/ImportWmsDirective.js#L231) hence your solution will not work.

gjn:
@oterral What do we do?

oterral:
For our use cases I think it's good.

gjn:
@davidoesch @cedricmoullet We don't have a use case for this, but the fix does make sense if components are used for a subset of Switerzland. It affects external wms layers, but I didn't detect any side effects:

[Testlink](https://mf-geoadmin3.dev.bgdi.ch/iodawms/?topic=ech&lang=en&bgLayer=ch.swisstopo.pixelkarte-farbe&X=261563.38&Y=626273.37&zoom=4&layers=WMS||LANDSAT%201975||http:%2F%2Frips-gdi.lubw.baden-wuerttemberg.de%2Farcgis%2Fservices%2Fwms%2FUIS_0100000017900001%2FMapServer%2FWMSServer||0||1.3.0||true)

It would be good to have it.

davidoesch:
there are 31000 uses cases for this http://www.europeandataportal.eu/data/en/dataset?res_format=WMS 

tested it with our wms .

Can't wait until  we have https://github.com/geoadmin/mf-geoadmin3/pull/2917 in https://github.com/geoadmin/mf-geoadmin3/issues/2414

gjn:
Ok, I didn't found any to reproduce the bug that this PR adresses. Also @Jenselme didn't found any for map.geo.admin.ch (only for their protal, which is not whole of Switzerland)

Would be great to have it reproducable in map.geo.admin.ch...so if you know how, let us know :=)

oterral:
I think reducing the default extent (on Bern for example) then use this wms (http://ogc.heig-vd.ch/mapserver/wms ) in importWMS then use the zoomToExtent button of a layer. The map should not move outside the Bern area.

gjn:
This is good to me. I'll merge it. Thanks alot @Jenselme .

PR #3189, opened 2016-03-04
Fix #3188 and some profile display improvments

### LINKED ISSUE #3188: profile is not displayed when drawing new line feature
Situation:
* open map.geo.admin.ch -> draw & measure
* draw a new line
* click on tab profile

the profile tab is empty.
image

 If you add an additional vertex it will be drawn.
image

issue has been reported by customer via webgis mailbox.

oterral:
Fixes #3188  and the same bug for measure directive.

Other things:

  -  introduces promises in ProfilService
  -  bigger profile for measure features
  -  no more text truncated (name on top right and Y value on top left)
  -  fixes reducing of size on each appearance of profile.

[Test](https://mf-geoadmin3.dev.bgdi.ch/teo_prof)

### DIFF
--- a/src/components/measure/MeasureDirective.js
+++ b/src/components/measure/MeasureDirective.js
@@ -43,7 +43,7 @@ goog.require('ga_measure_service');
                 update(evt.target);
               });
             });
-            newFeature.changed();
+            update(newFeature);
           }
         };
         scope.$watch('feature', useFeature);
--- a/src/components/draw/style/draw.less
+++ b/src/components/draw/style/draw.less
@@ -164,19 +164,25 @@
 
 
 .ga-draw-popup {
-  
+
   &.ga-profile-active {
     right: 0px;
+
+    .ga-popup-content {
+      overflow: visible;
+    }
   }
 
   &.ga-tabs {
     [ga-profile] {
-      margin-top: 0px;
+      height: 145px;
     }
   }
+
   [ga-profile] {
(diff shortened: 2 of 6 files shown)


src/components/measure/MeasureDirective.js
```diff
@@ -43,7 +43,7 @@ goog.require('ga_measure_service');
                 update(evt.target);
               });
             });
-            newFeature.changed();
+            update(newFeature);
```
gjn: strange that this stopped working with an ol3 update. @oterral @fredj Do you know the reason?

gjn: This alone would have fixed the bug.

gjn:
Thanks alot. Nice improvement.

I'll cherry-pick this in the release branch and deploy it (afternoon).

src/components/measure/MeasureDirective.js
oterral: don't know  why the first call to `changed()` is ignored, the 2nd trigegrs correctly the event.

gjn: probably timing issue. We bind to `change` event just before, then try to call it.

fredj: @gjn no clue ... do you use the latest ol3 version? https://github.com/openlayers/ol3/releases/tag/v3.14.2

Issue #3152, opened 2016-02-17
kml/json option to not show tooltip on click only highlight


davidoesch:
#gjin: to be carefully thought about

cedricmoullet:
I can see the following approaches:
- New permalink parameter: layers_click_behaviour 0: standard, 1: event only, 2: tooltip+event only 3: highlight+event only, 4: tooltip and highlight+event, per layer
- New permalink parameter: layers_no_tooltip: 0 or 1 per layer
- New permalink parameter: no_tooltip: true of false

gjn:
KISS for me. Option 3. (always number lists)

Also, a) non-official, b) not documented, c) must be added by hand and maybe d) can be revoked/changed any time.

davidoesch:
Number 3: but document it not in the official help - at least I a way we won't forget that this feature exits when we do the next migration.

gjn:
This can be closes as the new parameter (notooltipo=true) parameter exists. See referenced PR and https://codepen.io/geoadmin/pen/yOBzqM?editors=0010

PR #3199, opened 2016-03-11
IFrame API

gjn:
This is a reduced alternative for https://github.com/geoadmin/mf-geoadmin3/issues/3146 and replaces https://github.com/geoadmin/mf-geoadmin3/pull/3164. It also establishes a first version of an iframe API.

It adds a service to handle iframe messaging to parent pages. Message is an object of the form
```
{
  type: messageType, //the message type
  payload: data //the data for this type
}
```

The PR also adds the notooltip parameter. If true, no tooltip is ever shown. It's application wide setting and applied for all layers. It's also working on mobile, desktop and embeded.

Existing message is adapted to make it structured. It adds the following messages:

- Selection of kml object
- Saving of kml file (as in https://github.com/geoadmin/mf-geoadmin3/pull/3317)
- Application change state (permalink is updated)

For details, see [Testlink (on codepen)](http://codepen.io/gjn19/pen/VajQyo)

### DIFF
--- a/src/components/tooltip/TooltipDirective.js
+++ b/src/components/tooltip/TooltipDirective.js
@@ -26,7 +26,8 @@ goog.require('ga_topic_service');
   module.directive('gaTooltip',
       function($timeout, $http, $q, $translate, $sce, gaPopup, gaLayers,
           gaBrowserSniffer, gaMapClick, gaDebounce, gaPreviewFeatures,
-          gaMapUtils, gaTime, gaTopic, gaIdentify, gaGlobalOptions) {
+          gaMapUtils, gaTime, gaTopic, gaIdentify, gaGlobalOptions
+          gaPermalink) {
         var mouseEvts = '';
         if (!gaBrowserSniffer.mobile) {
           mouseEvts = 'ng-mouseenter="options.onMouseEnter($event,' +
@@ -532,6 +533,12 @@ goog.require('ga_topic_service');
               var name = feature.get('name');
               var featureId = feature.getId();
               var layerId = feature.get('layerId') || layer.id;
+              if (layer.get('type') == 'KML') {
+                layerId = layer.label;
+                if (name && name.length) {
+                  featureId = name;
+                }
+              }
               var id = layerId + '#' + featureId;
               htmlpopup = htmlpopup.
                   replace('{{id}}', id).
@@ -566,6 +573,11 @@ goog.require('ga_topic_service');
 
             // Show the popup with all features informations
             var showPopup = function(html, value) {
+              // Don't show popup when notooltip paramter is active
+              if (gaPermalink.getParams().notooltip == 'true') {
+                return;
+              }
+
               // Show popup on first result
               if (htmls.length === 0) {
 
--- a/src/components/tooltip/TooltipDirective.js
+++ b/src/components/tooltip/TooltipDirective.js
@@ -89,7 +89,7 @@ goog.require('ga_topic_service');
             // onclick
             if (layer) {
               if (!vectorLayer || vectorLayer == layer) {
-                if (!featureFound && isFeatureQueryable(feature)) {
+                if (!featureFound) {
                   featureFound = feature;
                 }
               }
@@ -473,8 +473,9 @@ goog.require('ga_topic_service');
                     feature.setId(value.getId());
                     feature.set('layerId', layerId);
                     gaPreviewFeatures.add(map, feature);
-                    showPopup(value.get('htmlpopup'), value);
-
+                    if (value.get('htmlpopup')) {
+                      showPopup(value.get('htmlpopup'), value);
+                    }
                     // Store the ol feature for highlighting
                     featuresByLayerId[layerId][feature.getId()] = feature;
                   } else {
@@ -545,6 +546,9 @@ goog.require('ga_topic_service');
                   replace('{{descr}}', feature.get('description') || '').
                   replace('{{name}}', (name) ? '(' + name + ')' : '');
               feature.set('htmlpopup', htmlpopup);
+              if (!isFeatureQueryable(feature)) {
+                feature.set('htmlpopup', undefined);
+              }
               feature.set('layerId', layerId);
               showFeatures([feature]);
               // Iframe communication from inside out
--- a/src/components/tooltip/TooltipDirective.js
+++ b/src/components/tooltip/TooltipDirective.js
@@ -3,6 +3,7 @@ goog.provide('ga_tooltip_directive');
 goog.require('ga_browsersniffer_service');
 goog.require('ga_debounce_service');
 goog.require('ga_identify_service');
+goog.require('ga_iframe_com_service');
 goog.require('ga_map_service');
 goog.require('ga_popup_service');
 goog.require('ga_previewfeatures_service');
@@ -15,6 +16,7 @@ goog.require('ga_topic_service');
     'ga_browsersniffer_service',
     'ga_debounce_service',
     'ga_identify_service',
+    'ga_iframe_com_service',
     'ga_map_service',
     'ga_popup_service',
     'ga_previewfeatures_service',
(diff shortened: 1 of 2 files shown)


gjn:
@davidoesch @cedricmoullet Please verify. This is for wandern.ch.

cedricmoullet:
1. I had to update the codepen (id's of the span: http://codepen.io/cedricmoullet/pen/BKzPwY Is it correctly changed ?
2. Why do you work with the src code //mf-geoadmin3.int.bgdi.ch/gjn_kml_external/src/embed.html  ?

gjn:
That's certainly not the one I wanted to link to.

cedricmoullet:
I tried to create a drawing with map.geo.admin.ch. It's possible to click on a text, but not on polygon and symbol. Any idea why ? (see http://codepen.io/cedricmoullet/pen/BKzPwY )

cedricmoullet:
I got it -> because of tooltip

gjn:
Codepen is now correct.

cedricmoullet:
Also if a tooltip is available, nothing is shown for symbol and polyline/polygon: http://codepen.io/cedricmoullet/pen/BKzPwY Any idea why ?

gjn:
Yes, seems like our kmls (created by drawing) don't contain name for the kml and no name for Placement, as described in the PR description.

On the other hand, we read those values for other kmls for the popup.

Looks like we need to treat differernt kml's differently. That's bad.

davidoesch:
"name for the kml" 
-> shall we use by default the hash generated by geo.admin.ch
 "name for Placement"
-> here we have to add names while drawing (Line1, line2 like google mymaps does st
<Placemark>        <name>Point 1</name>
instead of 
 <Placemark id="linepolygon_1457714969707

gjn:
We don't put anything in <name> tag for Placemarks because AFAIK, the name tag is autoamatically displayed when displaying it in google maps.

I think quick brainstorm is needed to define what we want to support. Remember, this is `hidden` feature (we don't document it)

davidoesch:
Brainstorm on Wed. As soon as wandern. Ch uses it... It will be hard to hide it

oterral:
Are you sure you don't want to send all the feature in a GeoJSON format ? That fit all usecases.
(@loicgasser is right send an object doesn't work on iE9)

gjn:
yes. KISS principle. layer id and feature id should be enough.

davidoesch:
is this already on prod? since it is already "documented" on http://api3.geo.admin.ch/releasenotes/index.html

gjn:
It's nowhere on official documentation. Codepn exmaple above is under my private account, not official codepen.

davidoesch:
unbenannt
@gjn

gjn:
ouch. correct example (geojson layer, not kml layer), wrong text in api3 
page.

gjn:
@davidoesch We still need to decide if we want 

a) adapt what we send to parent window (use optional 'id' attribute instead of <name> tag)
or
b) adapt what we put in our own drawings

Because currently, our own drawings would not send anything when clicking an object because our placements have no <name> tag (because name tags are always displayed on google maps).

gjn:
I've adapted the PR. Now, when the Placemark has no name, we will transmit the 'id' attribute of the Placemark.

I've also added the notooltip permalink parameter. See PR description.

The codepen example has been updated to reflect these changes.

Anyway, we still have a related problem with drawings that we are creating with map.geo.admin.ch. If we load such a kml (red elements in the codepen example), most elements are not selectable, only the label element is. This is true outside of this PR as well. @oterral Do you know why this is the case? Could this easily be adapted?

For me, this PR is good to merge.

gjn:
ping @cedricmoullet @davidoesch Would be great to have this for next weeks deploy.

cedricmoullet:
1. I don't see in the code where the choice between "name" and "id" is made. Where it is ?
2. Any answer to the question you asked to @oterral ?

oterral:
Sorry guys. If a feature has no name and no description, it is not selectable because there is nothing to display in the tooltip.

src/components/tooltip/TooltipDirective.js
```diff
@@ -490,6 +490,12 @@ goog.require('ga_topic_service');
               var name = feature.get('name');
               var featureId = feature.getId();
```
gjn: @cedricmoullet Per default, it's the id for a feature.

src/components/tooltip/TooltipDirective.js
```diff
@@ -490,6 +490,12 @@ goog.require('ga_topic_service');
...
               var layerId = feature.get('layerId') || layer.get('bodId');
+              if (layer.get('type') == 'KML') {
+                layerId = layer.label;
+                if (name && name.length) {
```
gjn: @cedricmoullet For KML, and if the feature has a 'name' attribute, it will be the name.

gjn:
@oterral With the new 'notooltip' parameter, it would make sense to make those no-name/no-description objects selectable still.
Or in general, we make all kml objects selectable, but only display a tooltip if there is a name or description.

Thoughts?

oterral:
yes why not

gjn:
Ok. I've update the PR so that kml objects with not name/description are selectable - without a tooltip.

I've updated the codepen example accordingly.

@oterral Quick code review?

@davidoesch @cedricmoullet Good to merge?

davidoesch:
Testtes the codepen: ok from my side

src/components/tooltip/TooltipDirective.js
```diff
@@ -490,12 +491,21 @@ goog.require('ga_topic_service');
...
+                  featureId = name;
+                }
+              }
               var id = layerId + '#' + featureId;
               htmlpopup = htmlpopup.
                   replace('{{id}}', id).
                   replace('{{descr}}', feature.get('description') || '').
                   replace('{{name}}', (name) ? '(' + name + ')' : '');
               feature.set('htmlpopup', htmlpopup);
+              if (!isFeatureQueryable(feature)) {
```
oterral: If you want the same behavior in 3d you remove the test here : https://github.com/geoadmin/mf-geoadmin3/blob/369cb88b9c9aa3e341d2f6af1bbbce2318c6be50/src/components/tooltip/TooltipDirective.js#L347

oterral:
lgtm minor comment for 3d behavior

cedricmoullet:
Tested on mac (safari, ff, chrome), all fine.

davidoesch:
tested on Android  6.0.1  with chrome: ok

procrastinatio:
You are really opening the Pandora's box with iframe - parent's windows communication

gjn:
I've updated this PR with a couple of API possibilities and added structure to the messages passed.

For details, see [Testlink (on codepen)](http://codepen.io/gjn19/pen/VajQyo)

davidoesch:
Nice!-- lets see the pandoras box discussion

gjn:
Note: this will fix https://github.com/geoadmin/mf-geoadmin3/issues/3152

davidoesch:
As decided with @gjn , we make a silent release of this PR with the new structure. But we will remove the a) drawing save message and b) permalink update messages.

cedricmoullet:
To be sure: it doesn't affect hydrodaten, right ?

gjn:
Yes, for 98% of browsers.

We are not sure for IE9...I could not verify for IE9 because codepen doesn't work with IE9.

gjn:
I looked deeper in their code (which is written with the necessary checks), and I think IE9 will be able to handle this case as well.

cedricmoullet:
Thanks

2016-06-10 13:22 GMT+02:00 Gilbert Jeiziner <notifications@github.com>:

> I looked deeper in their code (which is written with the necessary
> checks), and I think IE9 will be able to handle this case as well.
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/geoadmin/mf-geoadmin3/pull/3199#issuecomment-225157643>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AATgvPguqa9ibG-ZafQmt30WX9U8mWd9ks5qKUjegaJpZM4HuliU>
> .
>



-- 

PR #3200, opened 2016-03-11
Adapt deploy

gjn:
This PR assures 3 things:

1) In worst case, varnish takes 70 seconds before it detects that apache if up again after we deployed to an instance. We therefore wait in the post-restore hook 70 seconds before we continue our deploy to other instances
2) If we detect that Apache is down after 70 seconds, we issue an error and interrupt further deploy onto other machines
3) We don't rebuild the project on all target instances. The idea is the build _for_ the target instances in the snapshot directory and simply rscyn the code to the target instances. This will heavily reduce the downtime of apache on the target machine (from 3+ minutes to <15s)

1) and 2) should be done for mf-chsdi3 deploy as well.

@loicgasser as discussed, please have a look and we can test on integration.
@danduk82 @procrastinatio Your thougths (I know it's sub-optimal process, but it's on the safer side)

### DIFF
--- a/deploy/hooks/post-restore
+++ b/deploy/hooks/post-restore
@@ -8,4 +8,25 @@ if [ -x /etc/init.d/apache2 ]; then
     sudo /etc/init.d/apache2 restart
 fi
 
-exit 0
+# We should check if apache is running and return error if it's not
+sleep 5
+sudo /etc/init.d/apache2 status
+
+apache_rc=$?
+
+if [ $apache_rc -ne 0 ]; then
+
+  echo "-----------Deploy Error----------"
+  echo "Apache is not running on the deployed instance. Aborting deployment. Please check  instance"
+  echo "-----------End of Deploy Error----------"
+
+else
+
+  # We wait to assure that varnishes had time to detect running apache again
+  # before we deploy to next instance
+  echo "Waiting 70 seconds before continuing. Give varnish time to rediscover this instance..."
+  sleep 70
+
+fi
+
+exit $apache_rc
--- a/deploy/hooks/post-restore-code
+++ b/deploy/hooks/post-restore-code
@@ -11,13 +11,11 @@ CODE_DIR=$2
 cd $CODE_DIR
 
 # ---------IMPORTANT---------
-# We assume that code deployed does not need re-build on target.
-# It should be rebuild in source machine with correct settings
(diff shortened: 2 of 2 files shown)


deploy/hooks/post-restore
```diff
@@ -8,4 +8,19 @@ if [ -x /etc/init.d/apache2 ]; then
     sudo /etc/init.d/apache2 restart
 fi
 
-exit 0
```
loicgasser: I would sleep after we are sure apache restarted no?

deploy/hooks/post-restore-code
```diff
@@ -17,8 +23,8 @@ if [ -f rc_$TARGET ]; then
         if [ -f rc_branch ]; then
             source rc_branch
         fi
+        make all
```
loicgasser: as discussed you need to build with the correct variables and if we disable the tests it won't be the case for sure.

loicgasser: sorry that's in chsdi

deploy/hooks/post-restore
gjn: 2 concerns:
a) give varnish time to detect instance
b) stop deployment onto other instances when something is wrong with this instance.

This is simple aproach, on the safe side. Treating 2 concerns separately.

loicgasser: I mean if apache doesn't restart after 20 secs you can aready exit earlier

And 70 secs is the worsth case scenario for Varnish to detect if the instance is running when we know for sure apache has started.

loicgasser: It's probably enough but not very accurate

gjn: yes, I could bail out with error directly when apache status is wrong. Restart is blocking.

And only wait 70 when apache was/is up.

gjn: I've adapted. Testing now.

gjn:
For a visible effect _without_ this patch during deploy, [see here](https://logs.bgdi.ch/kibana/#/dashboard/elasticsearch/deploy%20varnish%20effect)

image

We ended up with both back-end being recognized as down by varnish, even though both were really up. The reason was to fast deploys between both machines (and not enough time for varnish to recognize that 1 machine was back up again).

Issue #3206, opened 2016-03-15
White tiles


gjn:
On some browser, under some conditions, the application turns into a state were certain tiles always stay white. Zoomin and panning does not get rid of this effect. A re-launch of the application helps most of the time.

Thsi is very hard to re-produce, but customers continue to file reports to that effect.

Might be related to offline issues (and/or appcache)

gjn:
Analyse impact of 500 errors of tiles on ol3.

loicgasser:
@gjn Might be related https://github.com/openlayers/ol3/pull/4964

gjn:
@loicgasser Good find! Might indeed be related.

Anyhow, still not able to reproduce, even when erratically zooming in and out as described in the bug description. But it also indicates that the ol3 internal tilecache might influence this.

I'll do some tests.

gjn:
Bingo. I can easily reproduce when I reduce the cacheSize from the default 2048 to 16, [see here](https://mf-geoadmin3.dev.bgdi.ch/gjn_whitetiles_demo/). Zoom in wildly a couple of times, then you'll have those artefacts.

Putting the cacheSize to 0 does not help. In fact it gets worst.

I will tripple the cacheSize to mitigate the effect until PR in ol3 is merged. I count on @fredj!

PR coming.

gjn:
@fredj Just a side question: what happens in the tilecache if a specific tile gets a 503 error once? Will it be re-requested if I do a simple zoom in/zoom out or will it stay marked as errounous in the tilecache?

fredj:
@gjn the tiles will be re-requested; the response code is not stored

gjn:
@fredj Thanks for the info!

gjn:
Patch if deployed to prod. Anyhow, I keep this open for reference and wait for the ol3 patch to be integrated in mf-geoadmin3 (it's merged in ol3).

I'll downgrade to 2.

@davidoesch Please let me know if we get feedback regarding white tiles again.

davidoesch:
OK will inform Helpdesk

gjn:
See also 
https://github.com/openlayers/ol3/issues/4338
https://github.com/openlayers/ol3/pull/5270

davidoesch:
Hi Gilbert
We have intranet and internet report that this issue increase more often in the last week
![image](https://cloud.githubusercontent.com/assets/4577727/22773569/26d5a1ec-eea3-11e6-8258-d267c8110a06.png)

[...]

PR #3210, opened 2016-03-15
Triple ol3 cacheSize to mitigate White Tiles effect

gjn:
This is a temporary fix for https://github.com/geoadmin/mf-geoadmin3/issues/3206 until https://github.com/openlayers/ol3/pull/4964 is merged upstream in ol3.

It reduces the likelyhood of running into our famous white tiles problem.

Thanks alot @ahocevar for finding it in ol3 and making it reproducable, @fredj for fixing it and @loicgasser for the hint.

### DIFF
--- a/src/components/map/MapService.js
+++ b/src/components/map/MapService.js
@@ -688,6 +688,9 @@ goog.require('ga_urlutils_service');
                 dimensions: {
                   'Time': timestamp
                 },
+                // Temporary until https://github.com/openlayers/ol3/pull/4964
+                // is merged upstream
+                cacheSize: 2048 * 3,
                 projection: gaGlobalOptions.defaultEpsg,
                 requestEncoding: 'REST',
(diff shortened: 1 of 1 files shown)


gjn:
@cedricmoullet @davidoesch here's a [testlink](https://mf-geoadmin3.dev.bgdi.ch/gjn_whitetiles/), even though it will be hard to verify.

Please test on your mobiles as I think my fix increases the memory usage of the application, which could have slight negative effect on small memory devices. I've tested on my Jolla, and it seemed ok with 4 layers active.

cedricmoullet:
Could you please provide a test link ?

gjn:
Posted at the same time.

cedricmoullet:
LGTM

davidoesch:
Tested on a galaxy tab 2 , Android 4.4.2 very low mem /CPU. All seems fine. Let's cross fingers

PR #3216, opened 2016-03-16
Update coordinate of brückenmoostrasse 11 raron

alcapat:


gjn:
hmm. still not good? @alcapat

gjn:
jenkins test

gjn:
@davidoesch @cedricmoullet FYI, the swissearch data was updated. According to @ltclm 

I did a manual update this morning of the address data.
They wanted us to change the source of the geometry. The new geometry source is the building entrance and not the building center anymore.
The new data is deployed to dev and will be deployed automatically this weekend to int and prod.

davidoesch:
woah this quite a change @ltclm --- finally they agreed. it does not seem to work everywhere. what is th fallback? switch to center as here? https://s.geo.admin.ch/6aaba9a349

ltclm:
yes it is great, you can compare the old and the new position with this url:
https://s.geo.admin.ch/6aabe637e8

PR #3217, opened 2016-03-16
Remove duplicate def of gaLayerHideManager

loicgasser:


oterral:
@loicgasser you can remove both this service has been renamed .

loicgasser:
ok I was wondering right now cause I couldn't see it used anywhere else.

gjn:
I'm through as well. I think this is good to merge.

Jenkins failur is fixed in another PR.

PR #3223, opened 2016-03-16
Remove browserstack from default e2e tests

loicgasser:
Now we have saucelabs and it seems more stable to me. Browserstack has been failing a lot.
We keep it available but remove it per default from e2e tests and therefore Jenkins

PR #3225, opened 2016-03-17
IE browsers only for browserstack tests

gjn:
As an alternative to https://github.com/geoadmin/mf-geoadmin3/pull/3223, I propose to keep browserstack tests for IE on different platforms, as long as we don't support IE on saucelabs.

Note: it's minimal tests only and they didn't give any troubles in the past on browserstack.

@cedricmoullet @loicgasser Do you agree?


loicgasser:
+1

gjn:
@cedricmoullet Ready to be merged.

PR #3229, opened 2016-03-17
Adapt build

gjn:
This PR adapts the current build to better refelct the build in mf-chsdi3. It does

1) Deploy targets are taken per default from snapshot directory now (as in chsdi3)
2) Renaming of make targets to not confuse build targets and staging targets `debug` and `testdebug` instead of `dev` and `testdev`. `release` and `testrelease` instead of `prod` and `testprod`
3) Add a `make user` target as the main target to be used by the developers (as in chsdi3)

[Testlinkg](ttps://mf-geoadmin3.dev.bgdi.ch/gjn_build/) shows that a fresh checkout build does work find.

We specifcy the bash shell as the default shell in the `Makefile` now (to support `source` command), which is the only thing with _potential_ side effects. But I didn't find any.

@oterral @loicgasser Feedback and review welcome.

@davidoesch @cedricmoullet Internal changes - but needed to streamline our main build tool among projects.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,7 @@ GIT_BRANCH := $(shell if [ -f .build-artefacts/deployed-git-branch ]; then cat .
 GIT_LAST_BRANCH := $(shell if [ -f .build-artefacts/last-git-branch ]; then cat .build-artefacts/last-git-branch 2> /dev/null; else echo 'dummy'; fi)
 BRANCH_TO_DELETE ?=
 DEPLOY_ROOT_DIR := /var/www/vhosts/mf-geoadmin3/private/branch
+DEPLOYCONFIG ?=
 DEPLOY_TARGET ?= 'dev'
 LAST_DEPLOY_TARGET := $(shell if [ -f .build-artefacts/last-deploy-target ]; then cat .build-artefacts/last-deploy-target 2> /dev/null; else echo '-none-'; fi)
 OL3_VERSION ?= 34d8d77344ee0b653770f065c593d4ab7b5d102b # master, 2 mars 2016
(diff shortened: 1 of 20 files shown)


Makefile
```diff
@@ -46,6 +48,7 @@ DEFAULT_EPSG_EXTEND ?= '[420000, 30000, 900000, 350000]'
 DEFAULT_ELEVATION_MODEL ?= COMB
 DEFAULT_TERRAIN ?= ch.swisstopo.terrain.3d
 SAUCELABS_TESTS ?=
+USERNAME := $(shell whoami)
```
loicgasser: @procrastinatio suggested to me that we could a single rc_user file doing:

export USERNAME='$(shell whoami)'                                               
export API_URL=//mf-chsdi3.dev.bgdi.ch/$USERNAME                                
export APACHE_BASE_PATH=/$USERNAME                                              
export E2E_TARGETURL=https://mf-geoadmin3.dev.bgdi.ch/$USERNAME

-> we would the ignore all the other rc files that are user specific

I would simply add in makefile:
USER_SOURCE ?= rc_user

so that you can override this behaviour using:

`make user USER_SOURCE=rc_ltgal`

or simply

`source rc_ltgal && make all`

like before

loicgasser:
one small comment otherwise +1 for that

procrastinatio:
For the switch to S3, we should be able to build code from any branch/commit in  `debug`or `release` mode against any service.

Makefile
gjn: Indeed, a good idea. User specific is default, not main, as it is now.
I'll adapt.

gjn:
ok. I've adapted according to @procrastinatio idea. Using one rc_user file to source `make user`. That's nice.

Issue #3227, opened 2016-03-17
Build and update slipjs/fastclick


gjn:
We treat the same use case differently in our build. Both [slipjs](https://github.com/geoadmin/mf-geoadmin3/blob/master/Makefile#L251) and [fastclick](https://github.com/geoadmin/mf-geoadmin3/blob/master/Makefile#L259) are patched by a local patch during the build.

In the gitreposity, the src/lib/slipjs file is the patched version. src/lib/fastclick is the unpatched version.

This mismatch is probably re-matched again becasue we treat both differently on update: [slipjs](https://github.com/geoadmin/mf-geoadmin3/blob/master/Makefile#L511) is treated as standard npm module, while [fastklick](https://github.com/geoadmin/mf-geoadmin3/blob/master/Makefile#L523) is treated differently.

We should treat same things in the same way.

The current setup is also very annoying, because it updates `src/lib/slip.js` on every `make all` build, thus polluting the working space with unneeded changes.

@oterral Please find a minute to address this.

gjn:
I think I understand now. #3187 re-introduced the fastclick patch that was removed by an earlier fastclick update. But but it was committed to the minified version only.

It should be committed to the src version as well.

In any case, the way we handle external libs in the Makefile coukd be improved.

gjn:
This is done.

PR #3231, opened 2016-03-17
Apply patch when slip comes from node modules

loicgasser:
Currently we have a make target `slipjs` which is not required by any other make target.

slipjs is also a package.json dep and when we `npm install` initial debug file gets downloaded as it seems and revert the changes in the debug file.

Related commit:
https://github.com/geoadmin/mf-geoadmin3/commit/9d3e228fae85e2ce3ac04c44834432d705dc4afc#diff-9566e188f62d1912ae8f605cead2ca6c


gjn:
i think we should handle it the same way as fastclick as it's exactly the same use case. see #3227

loicgasser:
We just discussed a better way to handle this issue with @oterral 

1) all libs are commited in src/lib/
2) make all won't `npm` install per default
3) we keep the package.json to simplify libs updates only

I'll close this as it is a superficial fix and a better/more general fix will come soon.

gjn:
+1

PR #3234, opened 2016-03-18
Manage package.json deps

loicgasser:
Fix for https://github.com/geoadmin/mf-geoadmin3/issues/3227

So we decided to make a difference between

- dependencies (libs that are used front-end)
- devDependencies (libs that are used in dev only like testing framework, build tools etc...)

libs coming from dependecies are commited in the project under
`src/prod/`
These js deps can be updated via `make node_modules_prod` -> commit the changes

libs coming from devDependencies are always installed and never commited in the project.

So we'll proably be a bit faster for building the app and we now have clear strategy of what and when to do it.

@oterral please review :)


Makefile
```diff
@@ -500,36 +491,51 @@ test/karma-conf-dev.js: test/karma-conf.mako.js ${MAKO_CMD}
...
+ANGULAR_TRANSLATE_JS = angular-translate.js angular-translate.min.js
+ANGULAR_TRANSLATE_LOADER_JS = angular-translate-loader-static-files.js angular-translate-loader-static-files.min.js
+LOCALFORAGE = localforage.js localforage.min.js
+JQUERY = jquery.js jquery.min.js
+JQUERYXDOMAIN = jQuery.XDomainRequest.js  jquery.xdomainrequest.min.js
+D3 = d3.js  d3.min.js
+BOOTSTRAP = bootstrap.js bootstrap.min.js
+SLIPJS = slip.js
+.PHONY: nodemodulesprod
+nodemodulesprod: package.json
```
oterral: `node_modules_prod ` to be consistent with `node_modules`

oterral:
lgtm one minor comment.

I've tested ` make cleanall all`  then ` make nodemodulesprod` seems to works fine.

loicgasser:
Have noticed something I don't like:

npm --version
1.3.10

and latest version is 3.8.3!!!

so --only=dev and --only=production are ignored...

the older counterpart is

--dev and --production but it breaks the build. I leave anyway I guess.

Makefile
loicgasser: yes makes sense

loicgasser:
we start having issue with some libs as well:

npm WARN engine hawk@3.1.3: wanted: {"node":">=0.10.32"} (current: {"node":"v0.10.5","npm":"1.3.10"})

npm WARN engine is-buffer@1.1.3: wanted: {"node":">=0.12"} (current: {"node":"v0.10.5","npm":"1.3.10"})

npm WARN engine cryptiles@2.0.5: wanted: {"node":">=0.10.40"} (current: {"node":"v0.10.5","npm":"1.3.10"})

npm WARN engine hoek@2.16.3: wanted: {"node":">=0.10.40"} (current: {"node":"v0.10.5","npm":"1.3.10"})

npm WARN engine boom@2.10.1: wanted: {"node":">=0.10.40"} (current: {"node":"v0.10.5","npm":"1.3.10"})

We will have to migrate at some point.

gjn:
Thanks alot for this. Much needed. I like the separation of developmemt libraries and production libraries.

I'll do a more thorough review.

Makefile
```diff
@@ -58,6 +58,8 @@ HTMLMIN_CMD=${PYTHON_VENV}/bin/htmlmin
 GJSLINT_CMD=${PYTHON_VENV}/bin/gjslint
 FLAKE8_CMD=${PYTHON_VENV}/bin/flake8
 AUTOPEP8_CMD=${PYTHON_VENV}/bin/autopep8
+NODE_ENV=node_modules
```
gjn: Is only needed on the line below. Probably better to use node_modules.

Makefile
```diff
@@ -66,33 +68,34 @@ help:
...
+	@echo "- cleanall           Remove all the build artefacts"
+	@echo "- deploydev          Deploys current github master to dev. Specify SNAPSHOT=true to create snapshot as well."
+	@echo "- deployint          Deploys snapshot specified with SNAPSHOT=xxx to int."
+	@echo "- deployprod         Deploys snapshot specified with SNAPSHOT=xxx to prod."
+	@echo "- deploydemo         Deploys snapshot specified with SNAPSHOT=xxx to demo."
+	@echo "- deletebranch       List deployed branches or delete a deployed branch (BRANCH_TO_DELETE=...)"
+	@echo "- deploybranch       Deploys current branch to test (note: takes code from github)"
+	@echo "- deploybranchint    Deploys current branch to test and int (note: takes code from github)"
+	@echo "- deploybranchdemo   Deploys current branch to test and demo (note: takes code from github)"
+	@echo "- ol                 Update ol.js and ol-debug.js"
```
gjn: Not related, but this target is now called ol3cesium, and it updates ol3cesium, ol3 and cesium libraries.

Makefile
```diff
@@ -66,33 +68,34 @@ help:
...
+	@echo "- node_modules_prod  Update productive depedencies in package.json"
```
gjn: call it node_libs?

Makefile
```diff
@@ -395,6 +374,18 @@ define buildpage
 		--var "staging"="$(DEPLOY_TARGET)" $< > $@
 endef
 
+define applypatches
+	git apply --directory=src scripts/fastclick.patch;
+	git apply --directory=src/lib scripts/slipjs.patch;
```
gjn: I would form the patches in the same way (look at the directory that is different)

Makefile
```diff
@@ -500,36 +491,51 @@ test/karma-conf-dev.js: test/karma-conf.mako.js ${MAKO_CMD}
...
-	cp $(addprefix node_modules/d3/,$(D3)) src/lib/;
-	cp $(addprefix node_modules/bootstrap/dist/js/,$(BOOTSTRAP)) src/lib/;
-	cp node_modules/fastclick/lib/fastclick.js src/lib/;
-	cp node_modules/angular-mocks/angular-mocks.js test/lib/;
-	cp node_modules/expect.js/index.js test/lib/expect.js;
-	cp node_modules/sinon/pkg/sinon.js test/lib/;
-	cp node_modules/google-closure-compiler/contrib/externs/angular-1.4.js externs/angular.js;
-	cp node_modules/google-closure-compiler/contrib/externs/jquery-1.9.js externs/jquery.js;
+test/lib/angular-mocks.js:
+	cp -f node_modules/angular-mocks/angular-mocks.js $@;
```
gjn: Targets without a prerequisite are run only when the target file does not exist. But those target files actually do always exists, as we store them in git. Not even a `make cleanall all` would run it. Only deleting the target file will run it. Try changing the version in the package.json - the target will not be updated.

We need to run it whenever the version specified in the package.json is changed. So probably somthing like this:

```
test/lib/angular-mocks.js: node_modules/angular-mocks/angular-mocks.js
   cp -f $< $@
```

The same is true for all the below.

Makefile
```diff
@@ -500,36 +491,51 @@ test/karma-conf-dev.js: test/karma-conf.mako.js ${MAKO_CMD}
...
+test/lib/sinon.js:
+	cp -f node_modules/sinon/pkg/sinon.js $@;
+externs/angular.js:
+	cp -f node_modules/google-closure-compiler/contrib/externs/angular-1.4.js $@;
+externs/jquery.js:
+	cp -f node_modules/google-closure-compiler/contrib/externs/jquery-1.9.js $@;
+.PHONY: package_json_dev
+package_json_dev:
+	npm install --only=dev
+node_modules: package_json_dev externs/jquery.js externs/angular.js test/lib/sinon.js test/lib/expect.js test/lib/angular-mocks.js
```
gjn: I don't see the need for this rule.

Makefile
```diff
@@ -500,36 +491,51 @@ test/karma-conf-dev.js: test/karma-conf.mako.js ${MAKO_CMD}
...
+ANGULAR_TRANSLATE_JS = angular-translate.js angular-translate.min.js
+ANGULAR_TRANSLATE_LOADER_JS = angular-translate-loader-static-files.js angular-translate-loader-static-files.min.js
+LOCALFORAGE = localforage.js localforage.min.js
+JQUERY = jquery.js jquery.min.js
+JQUERYXDOMAIN = jQuery.XDomainRequest.js  jquery.xdomainrequest.min.js
+D3 = d3.js  d3.min.js
+BOOTSTRAP = bootstrap.js bootstrap.min.js
+SLIPJS = slip.js
+.PHONY: node_modules_prod
+node_modules_prod: node_modules
```
gjn: Why not have `package_json_dev` as prerequisite here?

package.json
```diff
@@ -35,5 +21,21 @@
     "fastclick": "1.0.6",
     "bootstrap": "3.3.6",
     "slipjs": "pornel/slip#0626afc"
```
gjn: Not sure, but whenever npm install is run, slipjs is updated. That generally not happening with other libraries. Is it because we specify a commit instead of a version? Could we address this?

test/lib/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
+!.gitignore
```
gjn: Why is this file needed?

Makefile
gjn: Ok, I see now that the test files were removed from the git repository and a cleanall removes them as well.

The update on package.json will still not happen without a clean-all. And having the source as the pre-requisite is not bad anyway.

Makefile
```diff
@@ -500,36 +491,51 @@ test/karma-conf-dev.js: test/karma-conf.mako.js ${MAKO_CMD}
...
+
+ANGULAR_JS = angular.js angular.min.js
+ANGULAR_TRANSLATE_JS = angular-translate.js angular-translate.min.js
+ANGULAR_TRANSLATE_LOADER_JS = angular-translate-loader-static-files.js angular-translate-loader-static-files.min.js
+LOCALFORAGE = localforage.js localforage.min.js
+JQUERY = jquery.js jquery.min.js
+JQUERYXDOMAIN = jQuery.XDomainRequest.js  jquery.xdomainrequest.min.js
+D3 = d3.js  d3.min.js
+BOOTSTRAP = bootstrap.js bootstrap.min.js
+SLIPJS = slip.js
```
gjn: All those variables above are used exactly once below. We could safe some lines putting the filenames below directly.

Makefile
gjn: If the above comment is done, this could be simple function where we pass a library name (applypatch(slipjs))

gjn:
Thanks alot @oterral and @loicgasser for this nice improvement.

Issue #3219, opened 2016-03-16
graticule selection: add WGS84 and LV95 as checkbox


davidoesch:
With WGS84 and LV95 as dedicated layer [1] in the viewer we give the user now the options to choose in the viewer the layers
between 2 graticules instead of the combination
[1] org.epsg.grid_4326 https://redmine.prod.bgdi.ch/issues/4269 
 org.epsg.grid_2056 https://redmine.prod.bgdi.ch/issues/4740 

Change UI so the Coord grid selection adds now only Lv95

cc @mariokeusen

cedricmoullet:
I would not add a checkbox, but document in the help how to do that with wms. It doesn't seems to be a feature requested by the majority...

Envoyé de mon iPhone

davidoesch:
it was a feature request. imho we can just stitch to lv95 grid. 
the proper solution would be to have it in the pdf as tick marks on th edge of the map as defined in the original spec for re3....

AFoletti:
now that the grids are (will be) standard layers, a special print setting should no more be needed.
+1 for Cedric's suggestion: document the grids in the help and let the user build his map with the grid as a layer in order to have it printed out

If we really need a print setting, I suggest we stick with lv95 only

davidoesch:
Stick to lv95 only

AFoletti:
wil be fixed on prod with next deploy

PR #3239, opened 2016-03-23
replaced lv03 and wsg84 with lv95 as grid from print checkbox

AFoletti:
this is in response to https://github.com/geoadmin/mf-geoadmin3/issues/3219

Must wait for merge since the lv95 grid layer is not in production yet

### DIFF
--- a/src/components/print/PrintDirective.js
+++ b/src/components/print/PrintDirective.js
@@ -725,7 +725,7 @@ goog.require('ga_time_service');
           'opacity': 1,
           'singleTile': true,
           'type': 'WMS',
-          'layers': ['org.epsg.grid_21781,org.epsg.grid_4326'],
+          'layers': ['org.epsg.grid_2056'],
           'format': 'image/png',
           'styles': [''],
           'customParams': {


gjn:
Gridlayer will be deployed on next deploy?

AFoletti:
yes

gjn:
@davidoesch Won't we surprise some people with this change?

davidoesch:
It is made sure that in the expert tooltip of obstacle layer both are available as default. For the normal user this is an improvement. He can add the other layers separately

gjn:
AFAIK, luftfahrhindernisse users use mainly pdf print, no?

Anyway, state of extended tooltip on production:https://api3.geo.admin.ch/rest/services/ech/MapServer/ch.bazl.luftfahrthindernis/5430/extendedHtmlPopup?lang=de

Either I do something wrong or the object is not visible and there's no grids shown.

AFoletti:
?
the tooltip looks perfectly fine by me. Coming from your link:
image

gjn:
Indeed. Works fine in Android/Chrome.

But not in Android/Ghostery browser. Strange...but let's ignore it.

Back to topic: I think bazl users are using print often. Didn't we introduce wgs84 grid just for them in the print?

davidoesch:
We introduced it 3 years ago for them in Print. Now they have seen that they need the 2 grids only in the expert tooltip. And Swissgrid is enough for them in expert tooltip. Makes all easier for us. BTW your link works perfectly for me as well on Android and brave browser

gjn:
Ok. Thanks for the explanations.

@AFoletti any way to test this before going live?

AFoletti:
Of course! Have a look at it [HERE](https://mf-chsdi3.dev.bgdi.ch/ltfoa/shorten/6b1acaf1dc)

gjn:
Looks good. Thanks alot for the Testlink and the fix.

I've noted in the deploy doc to deploy WMS.

davidoesch:
@mariokeusen can you test every zoom level portrait landscape with every paper size? So we have at least 2 lines vertical horizontal on it? Just to make sure ....

PR #3243, opened 2016-03-23
Instances for int on branch

gjn:


gjn:
@loicgasser having default config from working directory is better, I think.

Otherwise, we are not able to easily deploy old snapshos to new instances. There's no use case to deploy old snapshots to old (non-existant) instances.

loicgasser:
OK

PR #3244, opened 2016-03-23
added a predefined query filter for astra_alle layer, plus translations

AFoletti:


gjn:
Testlink?

AFoletti:
With [this link](https://mf-geoadmin3.dev.bgdi.ch/ltfoa/src/?lang=de&topic=vu&bgLayer=ch.swisstopo.pixelkarte-grau&layers_timestamp=,2011,2011,2011,2015&X=224250.00&Y=638750.00&zoom=2&layers=ch.astra.unfaelle-personenschaeden_alle,ch.astra.unfaelle-personenschaeden_getoetete,ch.astra.unfaelle-personenschaeden_fussgaenger,ch.astra.unfaelle-personenschaeden_fahrraeder,ch.astra.unfaelle-personenschaeden_motorraeder&layers_visibility=true,false,false,false,false&catalogNodes=1318) you can test this PR and the corresponding CHSDI one (https://github.com/geoadmin/mf-chsdi3/pull/1976)
Both need to be merged for next deploy

gjn:
We should be able to test this with generated wmts tiles.

AFoletti:
What do you mean? Those astra layers are supposed to switch to WMS

gjn:
Oh. They were kind of slow, that's why I thought tiles would be created.

AFoletti:
yes, performance is suboptimal now. But we are on dev infra and dev WMS
@ltclm , your thought? Tiles are rapidly done if needed

ltclm:
i would use the wms for the go live, if we have a bad feedback we can switch to wmts afterwards.
if we use wmts we have to generate the timestamps and for the default view to generate a 'current' timetamp with all the data.

gjn:
Any way to be able to see performance on production environment before we go live?

AFoletti:
we could push to prod a new version of wms-bgdi with only the astra layer changes. Would require a bit of tinkering and attention but it's doable

gjn:
@AFoletti @ltclm Will the layer be similar in performance than the luftfahrthindernisse layer in production?

gjn:
This can be merged once https://github.com/geoadmin/mf-chsdi3/pull/1976 is merged.

loicgasser:
LGTM

src/components/query/QueryService.js
```diff
@@ -89,6 +89,18 @@ goog.provide('ga_query_service');
             operator: '>=',
             value: twoWeeksAgo
           }]
+        }],
+        'ch.astra.unfaelle-personenschaeden_alle': [{
+          id: 'astra_alle_lastyear_casualties',
+          filters: [{
+            attrName: 'accidentyear',
```
loicgasser: please add an index in DB

src/components/query/QueryService.js
```diff
@@ -89,6 +89,18 @@ goog.provide('ga_query_service');
...
+            operator: '=',
+            value: '2015'
+          }, {
+            attrName: 'severitycategorycode',
```
loicgasser: please add index in DB

Issue #3246, opened 2016-03-24
Offline issues


gjn:
Offline issues still persist for some clients. If found a few articles that might explain the issues:

[Fact sheet](http://appcache.offline.technology/)

[Why using prefer-online is a bad idea](http://labs.ft.com/2012/07/prefer-online-not-so-much/). I think the solution proposed here might be really useful.

[Appcache use cases and their problems](https://www.w3.org/community/fixing-appcache/2012/01/18/appcache_use_cases/)

[Gotchas on Firefox (scroll down)](https://developer.mozilla.org/de/docs/Web/HTML/Using_the_application_cache)

[More fun with appcache](https://eclecticdreams.com/blog/fun-times-with-appcache)

[Debugging appcache](https://jonathanstark.com/blog/debugging-html-5-offline-application-cache?filename=2009/09/27/debugging-html-5-offline-application-cache/)

[Debugging appcache in Firefox](http://flailingmonkey.com/application-cache-not-a-douchebag/)

Debugging appcache in Chrome with chrome://appcache-internals/

Then I have this message in Firefox45 when opening a page with appcache: `The Application Cache API (AppCache) is deprecated and will be removed at a future date.  Please consider using ServiceWorker for offline support.` So we might come to a state where appcache is not supported anymore on firefox.

gjn:
I again did some detailed analysis. Here are some results.
Initial state is working appcache store with same version on all browsers. Now the version of the cache is updated an newly deployed. The user then loads the page again with good internet connection.

**Chrome 49** (win7):
`checking` is triggered, `obsolete` is triggered (404 on the appcache file). This triggers an automatic page reload. Page with new version is loaded, then new appcache is loaded next. Events triggered in the followin order:
`checking`, `downloading`, `progress` (for each file to download), `cached`. The page is up-to-date without any problems.

A next load with the new version: `checking` (200 on appcache file with no change) and `noupdate` events are triggered.

**IE 11** (win7):
Exact same behaviour as Chrome 49.

**Safari 9** (latest ios), **Safari 5** (win7), **Safari** (on iphone 4, ios8):
Exact same behaviour as Chrome 49.

**Android Browser 4.2** (on Galaxy S4):
Exact same behaviour as Chrome 49.

**

**Firefox 45** (win7, the funny one):
[...]

PR #3248, opened 2016-03-24
FF: Assure reloading of page on appcache change

gjn:
This improves the update process of the appcache on firefox. With this, the appcache is updated on Firefox without any user interaction. We assume that 1) the appcache file is not cached in browser (older versions of appcache might trigger this, see history section of https://github.com/geoadmin/mf-geoadmin3/issues/3246) and that the internet connection is solid.

It might not fix all troubles we have, but some.

See https://github.com/geoadmin/mf-geoadmin3/issues/3246 for a deeper analysis.

[Testlink](https://mf-geoadmin3.dev.bgdi.ch/gjn_appcache99/)

### DIFF
--- a/src/js/MainController.js
+++ b/src/js/MainController.js
@@ -394,11 +394,15 @@ goog.require('ga_topic_service');
     }
 
     // An new appcache file is available.
-    if ($window.applicationCache) { // IE9
+    if ($window.applicationCache) {
       $window.applicationCache.addEventListener('obsolete', function(e) {
-        $window.location.reload();
+        // setTimeout is needed for correct appcache update on Firefox
+        setTimeout(function() {
+          $window.location.reload(true);
+        });
       });
     }
+
   });
 })();
 


gjn:
@oterral Could you review?

src/js/MainController.js
```diff
@@ -394,11 +394,14 @@ goog.require('ga_topic_service');
     }
 
     // An new appcache file is available.
-    if ($window.applicationCache) { // IE9
+    if ($window.applicationCache) {
       $window.applicationCache.addEventListener('obsolete', function(e) {
-        $window.location.reload();
+        setTimeout(function() {
```
oterral: Is the timeout use necessary ?

gjn: Yes. It's the fix for Firefox. Without this, the reload does happend, but the appcache update results in an error (thus requireing another manual reload).

With the timeout, it works directly.

oterral: Ok good.

oterral: May be add a cooment to explain why there is a setTimeout.

gjn: I do it.

oterral:
Lgtm

gjn:
Thanks for the review.

PR #3249, opened 2016-03-24
Simplify targets and update slipjs

oterral:
Modification of @loicgasser PR https://github.com/geoadmin/mf-geoadmin3/pull/3234 according to comments of @gjn .

I've added a `devlibs`  and `libs` target 

The `libs` (old `node_modules_prod`) target forces a `npm install --prod` command and update the js librairies in `src/lib` 
The `devlibs` target uses `npm install --dev` only if a dev file (`externs/[angular|jquery].js` or `test/lib/*.js`) doesn't exist.

The `devlibs` allows to run multiple  `make dev` or `make prod ` without running npm install each time.

@loicgasser @gjn what do you think?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,8 +58,7 @@ HTMLMIN_CMD=${PYTHON_VENV}/bin/htmlmin
 GJSLINT_CMD=${PYTHON_VENV}/bin/gjslint
 FLAKE8_CMD=${PYTHON_VENV}/bin/flake8
 AUTOPEP8_CMD=${PYTHON_VENV}/bin/autopep8
-NODE_ENV=node_modules
-CLOSURE_COMPILER=${NODE_ENV}/google-closure-compiler/compiler.jar
+CLOSURE_COMPILER=node_modules/google-closure-compiler/compiler.jar
 
 
 .PHONY: help
@@ -92,8 +91,8 @@ help:
 	@echo "- deploybranch       Deploys current branch to test (note: takes code from github)"
 	@echo "- deploybranchint    Deploys current branch to test and int (note: takes code from github)"
 	@echo "- deploybranchdemo   Deploys current branch to test and demo (note: takes code from github)"
-	@echo "- ol                 Update ol.js and ol-debug.js"
-	@echo "- node_modules_prod  Update productive depedencies in package.json"
+	@echo "- ol3cesium          Update ol3cesium.js, ol3cesium-debug.js, Cesium.min.js and Cesium folder"
+	@echo "- libs               Update js librairies used in index.html, see npm packages defined in section 'dependencies' of package.json"
 	@echo "- translate          Generate the translation files (requires db user pwd in ~/.pgpass: dbServer:dbPort:*:dbUser:dbUserPwd)"
 	@echo "- help               Display this help"
 	@echo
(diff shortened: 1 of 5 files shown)


gjn:
From code review, I like the improvement.

I was also thinking about bundling the node libraries with browserfy or webpack - but that's maybe for later.

I'll test deeper.

Makefile
```diff
@@ -92,8 +91,8 @@ help:
 	@echo "- deploybranch       Deploys current branch to test (note: takes code from github)"
 	@echo "- deploybranchint    Deploys current branch to test and int (note: takes code from github)"
 	@echo "- deploybranchdemo   Deploys current branch to test and demo (note: takes code from github)"
-	@echo "- ol                 Update ol.js and ol-debug.js"
-	@echo "- node_modules_prod  Update productive depedencies in package.json"
+	@echo "- ol3cesium          Update ol3cesium.js, ol3cesium-debug.js, Cesium.min.js and Cesium folder"
+	@echo "- libs               Update js librairies used in index.html, see npm packages defined in section 'dependencies' of package.json"
```
gjn: Maybe we should prefix the manually updated targets with 'update'. So `updateol3cesium` and `updatelibs`. But that's minor.

Makefile
```diff
@@ -491,42 +491,28 @@ test/karma-conf-dev.js: test/karma-conf.mako.js ${MAKO_CMD}
...
-ANGULAR_TRANSLATE_LOADER_JS = angular-translate-loader-static-files.js angular-translate-loader-static-files.min.js
-LOCALFORAGE = localforage.js localforage.min.js
-JQUERY = jquery.js jquery.min.js
-JQUERYXDOMAIN = jQuery.XDomainRequest.js  jquery.xdomainrequest.min.js
-D3 = d3.js  d3.min.js
-BOOTSTRAP = bootstrap.js bootstrap.min.js
-SLIPJS = slip.js
-.PHONY: node_modules_prod
-node_modules_prod: node_modules
+test/lib/angular-mocks.js test/lib/expect.js test/lib/sinon.js externs/angular.js externs/jquery.js:
```
gjn: If you add `package.json` as prerequisite here, the dev libraries will be updated when you change their version in the package.json.

And IMO, `libs` target should auto-update when changing package.json (would need specific target to have both forced/manual update and auto-update on package.json change)

gjn:
Added a couple of comments.

Another thing: I do a `make all`, project is build. Then I do a `make all` again without changing any files or environment variables, yet the project rebuilds. Same is true for `make prod` sequence. I think this is becaue the annotation is always done. We should try to improve it and strictly re-build when something changes only.

Makefile
oterral: I thought too but it doesn't work. Adding `package.json` doesn't change anything

gjn: You have to add it as prerequisite (after the colon).

Makefile
oterral: Since the description is up to date, i prefer to leave as it is. Short name is fatser to type:)

oterral:
Cna we merge this one (that merges on ltgal branch) and fixe the make prod in another PR?

gjn:
If @loicgasser agrees, I've nothing against merging this onto his branch. I like it.

loicgasser:
+1

PR #3252, opened 2016-03-29
[WIP] Add print support MapFish V3

mariavkourouni:
[Testlink](https://mf-geoadmin3.dev.bgdi.ch/kom_print_v3/)

Issues with MapFish Print v3:
- [ ] PDF size: no way (?) to control/compress images. Automatic resampling of images (if setting to 300 DPI, a 1-2 MB tile set will result in a 20 MB PDF)
- [ ] Parallel tiles download: for a A4 landscape map, **v3** is requesting 24 tiles, instead of 108 for **v2**. Don't think it is the issue, all requests are distributed on 1-2 secondes.
- [ ] Text translation. Simple way of doing it in iReport. For now solved client side.
- [ ] Conditional text.
- [ ] Legend layout
- [ ] Referer forward
- [ ] rotation support
- [ ] open legends which are pdf links as seprate pdfs (see schneesport topic in v2)
- [ ] scale pulldown
- [ ] disclaimer text to be adapted


gjn:
@davidoesch @cedricmoullet Please use the testlink above to test print v3. It uses a dedicated print service (from https://github.com/geoadmin/service-print).

Currently, it's really slow and I doubt that we can use like this in production. There are known problems that might impact its performance. One is currently being fixed by the mapfish print team and our friend @Jenselme in https://github.com/mapfish/mapfish-print/pull/396. @mariavkourouni will update as soon as this is available in master.

In addition, we are in contact with external company to analyse our setup (template, configuration, requests) to maybe further improve the performance.

gjn:
@mariavkourouni Could you make sure that the release link works as well? (the one without /src/

davidoesch:
I won't comment so far on the result just on the time it takes to generate the PDF: it is way too slow. The time passed the user clicked on the generate PDF BUTTON and the download completes has to be at least the same as it is currently in prod. This is a condition sine qua non . speed matters.

gjn:
There's no debate. We won't accept anything that is by factors slower than now.

davidoesch:
In this case: fix speed before we discuss functions

procrastinatio:
Did deploy https://github.com/geoadmin/service-print/pull/2 onto http://service-print.dev.bgdi.ch/printserver/

gjn:
@mariavkourouni @procrastinatio Can you create a new testlink please?

mariavkourouni:
[Testlink](https://mf-geoadmin3.dev.bgdi.ch/kom_print_v3/)

davidoesch:
@gjn : still signficant slower.  for a benchmark link [1] 

stuff to fix besides speed added above in ticket description

[1] https://mf-geoadmin3.dev.bgdi.ch/kom_print_v3/?topic=schneesport&lang=en&bgLayer=voidLayer&layers=ch.swisstopo.pixelkarte-farbe-pk50.noscale,ch.swisstopo.hangneigung-ueber_30,ch.swisstopo-karto.hangneigung,ch.bafu.wrz-jagdbanngebiete_select,ch.bafu.wrz-wildruhezonen_portal,ch.bazl.heliports-gebirgslandeplaetze,ch.swisstopo-karto.schneeschuhrouten,ch.swisstopo-karto.skitouren,ch.swisstopo.skitourenkarte-50.metadata,ch.bav.haltestellen-oev&layers_visibility=true,false,true,true,true,false,true,true,false,false&layers_opacity=0.85,0.5,0.2,0.6,0.6,1,0.8,0.8,1,1&catalogNodes=2352&X=203640.00&Y=660740.00&zoom=5

gjn:
@mariavkourouni Any results for the speed/performance tests you did?

procrastinatio:
snowsport A4 landscape 1:50'000, 150 dpi
v2: 2.78+/- 0.35   (api3.geo.admin.ch), 578 tiles requests in about <2s
v3: 9.28 +/- 1.43  (service-print.dev.bgdi.ch), 145 tile requests in <6s.

gjn:
2 different environments, no?

procrastinatio:
Maybe, but both are **c3.xlarge** instances and have the same **mf0** template. If on a dedicated and customized instance you cannot do better than on a same instance also used by other applications, it is not very encouraging.

gjn:
What about RAM dedicated to tomcat and the webapp?

procrastinatio:
@gjn 7GB (c3.xlarge)

gjn:
@mariavkourouni Please create a snapshot release.

mariavkourouni:
@gjn it's released.

mariavkourouni:
not yet released, I will confirm

mariavkourouni:
ok now I can confirm, it's released

PR #3256, opened 2016-03-30
Config a list to define which test must be done by a specific browser

alcapat:
Use different class instead of big one for checker_test. 
Setup a list of tests by browser (more easier to read)

### DIFF
--- a/test/saucelabs/checker_test.py
+++ b/test/saucelabs/checker_test.py
@@ -64,6 +64,32 @@ important_layers = [
     'ch.bfe.sachplan-uebertragungsleitungen_kraft']
 
 
+def runCheckerTest(driver, url):
+    print 'Checker tests starts!'
+
+    try:
+        assert "dev" in url
+        url_4_api = URL_API_DEV
+    except Exception as e:
+        try:
+            assert "int" in url
+            url_4_api = URL_API_INT
+        except Exception as e:
+            try:
+                assert "ci" in url
+                url_4_api = URL_API_CI
+            except Exception as e:
+                url_4_api = URL_API_PROD
+                print '-----------'
+                print str(e)
+                print "Current url: " + url
+
+    for ckFunc in checkerFunctions:
+        ckFunc(driver, url, url_4_api)
+
+    print 'Checker tests completed'
+
+
 def wait_url_changed(driver, old_url, timeout=DEFAULT_WAIT):
     UrlHasChanged = False
     t0 = time.time()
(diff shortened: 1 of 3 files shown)


test/saucelabs/test.py
```diff
@@ -155,35 +128,28 @@ def parse_args(args):
...
+                command_executor='http://' +
+                saucelabs_user +
+                ':' +
+                saucelabs_key +
+                '@ondemand.saucelabs.com:80/wd/hub',
+                desired_capabilities=current_desired_cap)
+            driver.implicitly_wait(DEFAULT_WAIT_FOUND)
+
+            if driver.name == "MicrosoftEdge":
+                print 'Force set version, strange...'
```
loicgasser: what is the reason for that?

loicgasser: meaning you define a certain version with IE edge when creating the driver and the version gets changed?

test/saucelabs/test.py
```diff
@@ -56,29 +56,6 @@ def parse_args(args):
     top_browser = {'platform': "Windows 7", 'browserName': "firefox",
```
loicgasser: This should probably be part of `desired_cap_list` as well no?

test/saucelabs/checker_test.py
```diff
@@ -81,222 +107,217 @@ def wait_url_changed(driver, old_url, timeout=DEFAULT_WAIT):
...
-        print "  Test python Translations"
-        driver.get(url_4_api + 'testi18n?lang=de')
-        assert "BLN" in driver.page_source
-        assert "Jagdbanngebiete" in driver.page_source
-        driver.get(url_4_api + 'testi18n?lang=en')
-        assert "ILNM" in driver.page_source
-        assert "Swiss game reserves" in driver.page_source
-        driver.get(url_4_api + 'testi18n?lang=toto')
+
+def ShortenUrl(driver, url, url_4_api):
```
loicgasser: minor: avoid cap for the first letter of a function

loicgasser:
Some small comments otherwise I I like it. It's easy to configure tests for each browser. Great!

test/saucelabs/test.py
alcapat: I don't know why but with Edge you loose the 'version' information. 
So i force to rewrite it, not so nice, i'm agree, if you have an other solution ?

test/saucelabs/test.py
alcapat: we add the "top_browser" in the desired_cap_list

gjn:
@alcapat Thanks alot for the improvements.

PR #3257, opened 2016-03-30
Add translation parameter to wms request

gjn:
To have translated labels of wms services. That's a quick fix, and doesn't react on language changes by the user.

### DIFF
--- a/src/components/map/MapService.js
+++ b/src/components/map/MapService.js
@@ -713,7 +713,8 @@ goog.require('ga_urlutils_service');
           } else if (layer.type == 'wms') {
             var wmsParams = {
               LAYERS: layer.wmsLayers,
-              FORMAT: 'image/' + layer.format
+              FORMAT: 'image/' + layer.format,
+              LANG: gaLang.get()
             };
             if (timestamp) {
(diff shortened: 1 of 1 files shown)


gjn:
Also, we might need to extend this for all wms layer (also externally imported...)

gjn:
Note also, that this is not working in 3D, but it's not really relevant there as in 3D, the affected layers seem to be tiled. But I'm not sure about this.

gjn:
Addendum: we have same problem in 3D. Not solvable as fast, unfortunately.

davidoesch:
we dont' translate geodata/feature attributes/values... and we should stick to that

@gjn do you have an example of the issue? screenshot link?

ltclm:
if we extend the wms interaction to more then getmap in near future, i think we should do that. we have to be able to use the translated service (getfeatuteinfo, getlegend). at least for our own service.

gjn:
@ltclm [Here's an example](https://map.geo.admin.ch/?layers=ch.swisstopo.gkg500-papierkarte.metadata)

Quite a few shop layers have text in their `getmap` result, that is translated.

PR #3266, opened 2016-04-06
[WIP] Change base_URL for tiles from wmts.geo.admin.ch to use our s3 endpoint with direct access to S3 bucket

mariavkourouni:
@danduk82 

[Testlink](https://mf-geoadmin3.dev.bgdi.ch/kom_print_v2_s3_endpoint/src/)

### DIFF
--- a/src/components/print/PrintDirective.js
+++ b/src/components/print/PrintDirective.js
@@ -450,7 +450,8 @@ goog.require('ga_time_service');
             }
             angular.extend(enc, {
               type: 'WMTS',
-              baseURL: location.protocol + '//wmts.geo.admin.ch',
+              baseURL: location.protocol +
+                '//s3-eu-west-1.amazonaws.com/akiai4jxkwjqv5tgsaoq-wmts',
               layer: config.serverLayerName,
               maxExtent: layer.getExtent(),
               tileOrigin: tileGrid.getOrigin(),


danduk82:
We have changed this so that the requests that should get tiles from the normal s3 bucket, now get them directly from the s3 endpoint, instead of using the NAT-varnish etc stack 2 times per request.

The problem is that now the print gets 403 errors, because it tries to access to some /-1 coordinates. The pdf is created anyway, but the logs are full of these. 

Anyway, we should use the direct connection to S3 endpoint anyway.

@loicgasser , @gjn  and @procrastinatio What do you think?

gjn:
+1 from me.

Only impact I see is we loose the wmts tiles logs in both kibana and aws stats for all tiles requested by the print.

If @davidoesch and @cedricmoullet are ok with this, then we can merge this.

gjn:
@pfanguin @rfrancel @alcapat Can you please test the print using the testlink above. Use the print examples in squash TM, use border cases. Print outside switzerland, etc.

Note if you can detect something fishy in the resulting PDF.

davidoesch:
Sorry folks: a no go- we MUST deliver by law a stats of tiles (pixels) delivered to users. It is part of the stats we deliver every year to topo on which they base an invoice to the geo portal. ....

rfrancel:
I found two little culprits:
1. The first page of multiprint does not show the complete upper layer, but shows a part of the base layer (different in each print): https://mf-chsdi3.dev.bgdi.ch/shorten/6b1b34dcd9
2. The text in the scale and reference box goes out of the box (not in multiprint, only in single page PDFs)
Otherwise, I didn't see anything.

gjn:
@rfrancel Are those culprits also present in current production? (map.geo.admin.ch)

rfrancel:
no I did the same print in prod and I did not see them

gjn:
@davidoesch [Here's print wmts requests vs non-print wmts requests](https://logs.bgdi.ch/kibana/#/dashboard/elasticsearch/print%20wmts%20requests). Over last month, 5% of all wmts requests are coming from our print service. We would loose those in the logs.

Still a no-go?

davidoesch:
@gjn : you will get a decision by Bus and bov See Mail

danduk82:
@davidoesch after we switch to new logging infra, we will count them from S3 and cloudfront logs. But in the meantime, print request might still have download errors 503. Anyway, I am not sure if this will completely solve the print problem... (it should solve at least one of the print problems)

davidoesch:
It won't solve the Problem but reduces complexity... Already a huge step in stabilizing print

gjn:
Changed to WIP as it's under discussion.

procrastinatio:
Eventually moving this into the printproxy.

gjn:
We close this. @mariavkourouni Please snapshot/realease the branch.

mariavkourouni:
@gjn it's released.

mariavkourouni:
not yet released, I will confirm

mariavkourouni:
ok now I can confirm, it's released

PR #3268, opened 2016-04-07
Use safari on OS, version 9.0 (with smal screen reslolution)

alcapat:
I see that safari's test are failed this night. Check with configurator [1] to use a more recent version of this browser on OS. Change also screen resolution because "1280x1024" was not supported.

[1] https://wiki.saucelabs.com/display/DOCS/Platform+Configurator#/


gjn:
Too bad. Safari 5 is still sometimes used...

Issue #3276, opened 2016-04-12
[print] issue when printing drawings from KML generated with swissmapconverter


mariokeusen:
UC: open app, load a KML generated from an XOL with swissmapconverter. Print a PDF of the extent with the KML
Result: On the screen the lines are visualized smoothly, on the PDF the lines much fatter than on screen
Expected: looks more or less the same on screen and on the PDF
![print_kmlfromxol_symbology](https://cloud.githubusercontent.com/assets/4577749/14450779/b667ad52-0081-11e6-830c-791c5ee910b3.PNG)

oterral:
KML file please @mariokeusen

mariokeusen:
Requested that @helpdesk

mariokeusen:
I could reproduce the issue with Chrome and IE11, with the original KML from the customer.

PR #3280, opened 2016-04-12
Modify distances acoording to dpi

oterral:
Fix https://github.com/geoadmin/mf-geoadmin3/issues/3276

[Test](https://mf-geoadmin3.dev.bgdi.ch/teo_printwidth/?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&X=252454.82&Y=747104.44&dev3d=true&zoom=7&debug&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bafu.wrz-wildruhezonen_portal,ch.swisstopo.swisstlm3d-wanderwege,KML%7C%7Chttps:%2F%2Fpublic.dev.bgdi.ch%2Fceray2cFSBuuwT4I-_QYLQ&layers_visibility=false,false,false,false,true&layers_timestamp=18641231,,,,)

@mariokeusen can you test please ?

### DIFF
--- a/src/components/print/PrintDirective.js
+++ b/src/components/print/PrintDirective.js
@@ -271,9 +271,9 @@ goog.require('ga_time_service');
           fill = imageStyle.getFill();
           stroke = imageStyle.getStroke();
           var radius = imageStyle.getRadius();
-          var width = 2 * radius;
+          var width = adjustDist(2 * radius);
           if (stroke) {
-            width += stroke.getWidth() + 1;
+            width += adjustDist(stroke.getWidth() + 1);
           }
           size = [width, width];
           anchor = [width / 2, width / 2];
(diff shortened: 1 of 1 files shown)


src/components/print/PrintDirective.js
```diff
@@ -951,6 +951,14 @@ goog.require('ga_time_service');
       }
     };
 
+    // Change a distance according to the change of DPI
+    var adjustDist = function(dist) {
+      if (!dist) {
+        return;
+      }
+      return dist * 90 / getDpi($scope.layout.name, $scope.dpi);
```
gjn: What is this magic number?

oterral: The OGC standard output resolution is 90 DPI, it's used also by openlayers.

loicgasser: 90.7 to be accurate ;)

gjn: 90.7 is DPI in WMTS standard (used to calcualte scalefactor in WMTS GetCap).

Our WMTS Tiles are calculated with 96DPI and our WMS uses 96dpi as standard output. In print, we are using 150DPI.

Also, we use 72 DPI insided this print directive, I think to create the preview rectangle.

If I'm not completely mistaken, we should use 96 in this function here...or 72.

Issue #3208, opened 2016-03-15
Haltestellen Icons in print are too small


gjn:
Print https://map.geo.admin.ch/?topic=ech&lang=de&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bafu.wrz-wildruhezonen_portal,ch.swisstopo.swisstlm3d-wanderwege,ch.bav.haltestellen-oev,KML%7C%7Chttps:%2F%2Fpublic.geo.admin.ch%2FKWeuzcQWTR-Ym5AC75rFQg&layers_visibility=false,false,false,false,true,true&layers_timestamp=18641231,,,,,&X=156837.50&Y=540762.50&zoom=8

Icons are too small.

davidoesch:
same for the adressen... related to the resolution

rfrancel:
@davidoesch do you mean the Wohnungsregister?

davidoesch:
Yep, gwr

gjn:
This is a basic issue we have with pre-generated tiles. Sizes (icons sizes, line widths, etc) on the tiles are created for 96DPI (our standard WMS setting) - optimized for web and computer screens.

In print, we are using 150DPI. So, for a given zoom level in the WMTS pyramid, the icons, line width etc are .064 the size as they appear in the web-viewer. We also have this issues on screens with > 96DPI resolution.

We could address this by using the WMS and specifying the new dpi in the request. This seems to be supported by MapServer. But this means upscaling our WMS and assure that all layers are accessible via WMS. This requires quite some work (but migth be needed for other projects in the pipeline).

@rfrancel Maybe for test purposes, you can try to use WMS instead of WMTS for haltestellen print.

http://freegis.org/pipermail/mapserver-de/2012-January/004527.html
http://mapserver.org/development/rfc/ms-rfc-55.html

gjn:
The parameter `MAP_RESOLUTION` can be added to specify a different DPI for a given request:

[...]

PR #3283, opened 2016-04-13
Add MAP_RESOLUTION parameter for print WMS layers

gjn:
This PR addresses an issue we have when printing WMS layers. Per default, the printserver sends a `DPI` parameter along with the WMS request. But this parameter is ignored by our WMS server.

Instead, we need to use the `MAP_RESOLUTION` parameter, which our wms is supporting. This PR adds this parameter. 

For the description of the effect, please see https://github.com/geoadmin/mf-geoadmin3/issues/3208

Note: this change only affects WMS layers.

[Testlink](https://mf-geoadmin3.dev.bgdi.ch/gjn_print_wms/?topic=ech&lang=de&bgLayer=voidLayer&layers=ch.bfe.sachplan-geologie-tiefenlager,ch.swisstopo.geologie-gisgeol-punkte&X=243520.00&Y=644240.00&zoom=5&layers_opacity=0.75,1)

Here's how it looks in the browser:
image

Here's how it looks with current print in the pdf:
image

Here's how it looks with this PR in the pdf:
image

The new version resembles better the image in the browser. The icons and line widths have the correct pixel sizes.

@davidoesch @procrastinatio @ltclm Thoughts on this?

### DIFF
--- a/src/components/print/PrintDirective.js
+++ b/src/components/print/PrintDirective.js
@@ -433,7 +433,8 @@ goog.require('ga_time_service');
                 'EXCEPTIONS': 'XML',
                 'TRANSPARENT': 'true',
                 'CRS': 'EPSG:21781',
-                'TIME': params.TIME
+                'TIME': params.TIME,
+                'MAP_RESOLUTION': getDpi($scope.layout.name, $scope.dpi)
               },
               singleTile: config.singleTile || false
             });
@@ -729,7 +730,8 @@ goog.require('ga_time_service');
           'format': 'image/png',
           'styles': [''],
           'customParams': {
-            'TRANSPARENT': true
+            'TRANSPARENT': true,
+            'MAP_RESOLUTION': getDpi($scope.layout.name, $scope.dpi)
           }
         };
         encLayers.push(graticule);


oterral:
What's happening for external WMS, should  we add it too?

procrastinatio:
@oterral depends of the server, some support this kind of paramter, other not. **Geoserver** and **QGIS** server use a **DPI** parameter for this. It is *vendor specific*, not OGC.

gjn:
We add the MAP_RESOLUTION parameter to the external wms requests as well.

The printserver always sends both parameters, MAP_RESOLUTION and DPI.

procrastinatio:
It's a nice improvement, but we'll end with a big mix of layers which are scaled up (many WMS, KML) and layers (WMTS!) which are not.

gjn:
Yes - therefore, we need to think about using wms-only for printing. TBD.

davidoesch:
+1 for the above solution. might come handy for the future print service . I propose: for the BG layer we use always wmts, for the other wms....

gjn:
Thanks for the reviews.

PR #3284, opened 2016-04-14
Physical artefact for devlibs in Makefile

gjn:
This PR fixes the fact that repeated `make user` commands always re-creates the whole application (including annotation and compilation/minifying).

The culprit was the `devlibs` target, that was used as a pre-requisite on end targets. But `devlibs` never  really existed, so as a prerequisite, it was always considred as being out of date - thus triggering the build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,7 +119,7 @@ user:
 all: lint debug release apache testdebug testrelease deploy/deploy-branch.cfg fixrights
 
 .PHONY: release
-release: devlibs \
+release: .build-artefacts/devlibs \
 	prd/lib/ \
 	prd/lib/build.js \
 	prd/style/app.css \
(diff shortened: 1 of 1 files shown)


gjn:
@loicgasser Quick review/merge appreciated, because current state of build is awful during dev.

oterral:
@gjn I ve seen on tuesday that running `make testdebug` always relaunch app-whitespace build (with ng-annotate thing). could this PR also fixes this?

gjn:
Yes, absolutely.

In addition, annotate will only be run for the changed file(s) now - and not for all everytime it's needed.

oterral:
👍

loicgasser:
Tested and works as expected. A change in package.json will trigger a ng-annotate which is good.
+ 1 for me.

gjn:
Thanks for the reviews.

PR #3294, opened 2016-04-27
Use a common function to get the timeInstant param

oterral:
[Test](https://mf-geoadmin3.dev.bgdi.ch/teo_fixes/?lang=de&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.bafu.schutzgebiete-paerke_nationaler_bedeutung_perimeter,ch.bafu.schutzgebiete-paerke_nationaler_bedeutung,ch.bafu.showme-kantone_sturzprozesse,ch.bafu.showme-kantone_rutschungen,ch.bafu.showme-kantone_lawinen,ch.bafu.showme-kantone_hochwasser,ch.bafu.showme-gemeinden_sturzprozesse,ch.bafu.showme-gemeinden_rutschungen,ch.bafu.showme-gemeinden_lawinen,ch.bafu.showme-gemeinden_hochwasser,ch.swisstopo.geologie-geologischer_atlas,ch.swisstopo.geologie-geocover,ch.swisstopo.geologie-generalkarte-ggk200,ch.swisstopo.pixelkarte-farbe-pk500.noscale&layers_opacity=0.85,0.85,0.75,0.75,0.75,0.75,0.75,0.75,0.75,0.75,1,0.75,1,1&layers_visibility=false,false,false,false,false,false,false,false,false,false,false,false,true,true&X=206368.35&Y=649611.89&zoom=1)

### DIFF
--- a/src/components/query/QueryDirective.js
+++ b/src/components/query/QueryDirective.js
@@ -3,17 +3,20 @@ goog.provide('ga_query_directive');
 goog.require('ga_map_service');
 goog.require('ga_query_service');
 goog.require('ga_storage_service');
+goog.require('ga_time_service');
+
 (function() {
 
   var module = angular.module('ga_query_directive', [
     'ga_map_service',
     'ga_query_service',
-    'ga_storage_service'
+    'ga_storage_service',
+    'ga_time_service'
   ]);
 
   module.controller('GaQueryDirectiveController', function($filter, $rootScope,
       $scope, $timeout, $translate, gaLayers, gaLayerFilters, gaQuery,
-      gaMapUtils, gaStorage) {
+      gaMapUtils, gaStorage, gaTime) {
     var geojson = new ol.format.GeoJSON();
     var stored;
     $scope.queryType = 1; // Filter attributes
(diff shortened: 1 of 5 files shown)


oterral:
Wait before merge

oterral:
@loicgasser should be good now

src/components/query/QueryDirective.js
```diff
@@ -49,12 +53,16 @@ goog.require('ga_storage_service');
           return;
         }
         if (!paramsByLayer[filter.layer.bodId]) {
+          var params = {};
+          var timeInstant = getTimeInstantParam(filter.layer);
+          if (timeInstant) {
```
loicgasser: @gjn this PR is mainly because of that. We would send the time instant params even if the layer was not time enabled.

src/components/query/QueryDirective.js
```diff
@@ -22,10 +25,11 @@ goog.require('ga_storage_service');
...
     $scope.filters = [];
     $scope.featuresByLayer = {};
-    // return the year of the timestamp
-    var getYear = function(time) {
-      return (time && time.substr(0, 4) != '9999') ?
-          time.substr(0, 4) : undefined;
+
+    var getTimeInstantParam = function(layer) {
+      if (layer.timeEnabled) {
+        return gaTime.get() || gaTime.getYearFromTimestamp(layer.time);
```
loicgasser: gaTime.get could be put in `gaTime.getYearFromTimestamp` as we're always performing this test.

oterral: Yes it could but I prefer to keep this function as a simple parsing utilities function. May be we could need this without getting the current time.

src/components/query/QueryDirective.js
gjn: so basically only send time parameter when layer actually uses it.

loicgasser:
LGTM

src/components/query/QueryDirective.js
loicgasser: yes

loicgasser:
@gjn Those fixes are impacting the layers update of today's deploy. It would be good to have them.

PR #3301, opened 2016-05-03
Test pl chateau avenches return now 2 result, select the first

alcapat:
Fix saucelabs test for swisssearch.
Now the test "pl chateau avenches" return more than one result.
So we must select the first result to make the zoom in and compare the result on url (X, Y and Zoomlevel)


gjn:
Perfect. Thanks alot @alcapat !

PR #3305, opened 2016-05-09
Fix sauclabs search test, add new tests and refactoring

loicgasser:
This PR fixes the current issue with the search tests via the `swisssearch` permalink parameter.

The same address had 2 entries and as a result the dropDown search list was open instead of zooming directly on the expected location.

I introduced UI interactions with the input element of the search.

1. We first change the input value and trigger a search request
2. We check that the dropdown list is open
3. We select a result based on the location entry title
4. We check that the permalink contains the correct coordinates
5. That the dropdown is not visible anymore
6. We clear the input content and check that the input as been cleared

Similar tests are performed for the layer search and swisssearch permalink parameter.

We now have two use cases for the swisssearch permalink parameter.

1. Only one result -> we expect to zoom at the correct location
2. Several results -> we expect the dropDown to be open

Ideally I would like more tests like these ones.
We should use an extra layer of abstraction using classes for our components (or page element) that we could call from any test. See http://selenium-python.readthedocs.io/page-objects.html#page-elements for instance. That's what I started here with reusable functions that test the same thing.

ping @alcapat @pfanguin -> as a source of inspiration


test/saucelabs/swisssearch_test.py
```diff
@@ -1,223 +1,253 @@
...
-    UrlHasChanged = False
+def waitUrlChange(driver, pattern, find=True, timeout=DEFAULT_WAIT):
     t0 = time.time()
-    while not UrlHasChanged:
-        new_url = driver.current_url
-        if old_url != new_url:
-            UrlHasChanged = True
-        t1 = time.time()
-        if t1 - t0 > timeout:
+    newUrl = driver.current_url
```
gjn: We have many elements on the page that change their url in exactly the same way as the url in the address bar. Why don't we use those elements and profit from selenium internal detection loops and timeouts?

loicgasser:
Because it's a value test that is not avaiable in the native selenium waits.  See
http://selenium-python.readthedocs.io/waits.html

test/saucelabs/swisssearch_test.py
gjn: We can check for the href attribute of an <a> element, and the implicite waits with loops will happen, no?

gjn: hmm. maybe not. But it seems we did it differently with browserstack/js code. Might be worth checking out the js version...

loicgasser:
I ll check the js. This method was used prior to this Pr. I think this could be handled in the next PR. There is still a lot to do to create a full selenium api for geoadmin.

gjn:
The idea is to avoid to implement our own poll/wait loop. Instead use the selenium-proper implicit wait in the driver.find_element function; by using xpath selector with href attribute of the url wanted.

Anyway, i tkink this is a grewt refactoring and I want this in. Thanks alot @loicgasser for this proposal.

test/saucelabs/helpers.py
```diff
@@ -21,16 +21,27 @@ def bCheckIfUrlHasChanged(driver):
...
     t0 = time.time()
-    newUrl = oldUrl
-    while oldUrl == newUrl:
-        newUrl = driver.current_url
-        time.sleep(1)
-        ti = time.time()
-        tf = ti - t0
-        if tf > timeout:
-            raise Exception('%s has not changed in %.2f seconds' % (oldUrl, tf))
+    newUrl = driver.current_url
```
gjn: Btw: this will always fail on IE9 because there's not push history support there which we use to update the URL of the addreaa bar.

loicgasser:
I agree

Issue #3290, opened 2016-04-25
API URL via hidden permalink value


gjn:
I think to create testlinks for mf-chsdi3, it would make sense to be able to specify the API URL via hidden Permalink parameter. Therefore, you don't need to create a test-branch in mf-geoadmin3 to test mf-chsdi3 branches.

@rfrancel Could you give it a try?

rfrancel:
Yes, I'll have a look at it, thanks!

PR #3309, opened 2016-05-11
Frr api url param

rfrancel:
@gjn here is a [test  link](https://mf-chsdi3.dev.bgdi.ch/frr_tilematrixset/shorten/6bcb421dfb) with the chsdi branch as url parameter

linked to https://github.com/geoadmin/mf-geoadmin3/issues/3290

### DIFF
--- a/src/index.mako.html
+++ b/src/index.mako.html
@@ -726,6 +726,34 @@ itemscope itemtype="http://schema.org/WebApplication"
         var cacheAdd = '${version}' != '' ? '/' + '${version}' : '';
         var pathname = location.pathname.replace(/(index|mobile|embed)\.html$/g, '');
 
+        function getParam(name){
+          return decodeURIComponent((new RegExp('[?|&]' + name + '=' + '([^&;]+?)(&|#|;|$)').exec(location.search)||[,""])[1].replace(/\+/g, '%20'))||null;
+        }
+
+        var my_api_url = getParam('api_url');
+        var my_api_env = getParam('api_env');
+        var api_url_base = '${api_url}';
+
+        if (my_api_env){
+          if (my_api_env == 'prod'){
+            my_host = '//api3.geo.admin.ch';
+          } else {
+            my_host = '//mf-chsdi3.' + my_api_env + '.bgdi.ch';
+          }
+        } else {
+          my_host = '${api_url}';
+        }
+
+        if (my_host.slice(-1) == '/'){
+          my_host = my_host.slice(0, -1);
+        }
+
+        if (my_api_url){
+          api_url = my_host + '/' + my_api_url;
+        } else {
+          api_url = my_host;
+        }
+
         module.constant('gaGlobalOptions', {
           //dev3d to be removed once 3d goes live
           dev3d: true,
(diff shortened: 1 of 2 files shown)


src/index.mako.html
```diff
@@ -726,22 +726,34 @@
         var cacheAdd = '${version}' != '' ? '/' + '${version}' : '';
         var pathname = location.pathname.replace(/(index|mobile|embed)\.html$/g, '');
 
+        function getParam(name){
+          return decodeURIComponent((new RegExp('[?|&]' + name + '=' + '([^&;]+?)(&|#|;|$)').exec(location.search)||[,""])[1].replace(/\+/g, '%20'))||null;
```
gjn: Curios: where is the regexp coming from?

src/index.mako.html
```diff
@@ -726,22 +726,34 @@
...
+        }
+
+        var my_api_url = getParam('api_url');
+
+        if (my_api_url){
+          api_url = '//mf-chsdi3.dev.bgdi.ch/' + my_api_url;
```
gjn: We should think about integration, or other environments as well. So I would rather have here the domain of the ${api_url} variables per default. Then optionally, we can add a 2nd paramter where we can specify the environment. So we would have something like:

```
var my_api_env = getParam('api_env');
api_url = domain_from_url(${api_rul);
if (my_api_env) {
  api_url = '//mf-csdhi3.' + my_api_env + 'bgdi.ch/' + my_api_url;
}

gjn:
Well done. Exactly what I had in mind. I've added a comment for potential improvement.

Also, there's one thing we need to address:
LayersConfig and Topic services are _not_ served with the api_url (this is by design, because otherwise, offline won't work).

But what we could do is that if we detect an api_url parameter in the URL, we serve the LayersConfig and the Topic services from the api_url as well.

See https://github.com/geoadmin/mf-geoadmin3/blob/master/src/index.mako.html#L782 for layersConfig url configuration and https://github.com/geoadmin/mf-geoadmin3/blob/master/src/index.mako.html#L786 for topics url functionality.

src/index.mako.html
rfrancel: the regexp helps us find the parameter value. What is wrong with it? Is there another method you prefer I'd use?

rfrancel:
@gjn  I changed the PR according to your comments:
Here is a [test link](https://mf-geoadmin3.dev.bgdi.ch/frr_apiUrl_param/?api_env=int&topic=ech&lang=fr&X=189918.22&Y=659918.22&zoom=1&bgLayer=ch.swisstopo.swissimage) with the api_env parameter
and here is [one](https://mf-geoadmin3.dev.bgdi.ch/frr_apiUrl_param/?api_url=frr_shop_tooltips&topic=ech&lang=fr&X=190000.00&Y=660000.00&zoom=1&bgLayer=ch.swisstopo.swissimage&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bafu.wrz-wildruhezonen_portal,ch.swisstopo.swisstlm3d-wanderwege&layers_visibility=false,false,false,false&layers_timestamp=18641231,,,) woth a branch parameter.
I am not happy about the 'slice the url if it ends with /' but the base api_url does not return the same in my dev environment or in my deployed branch. 
//mf-chsdi3.dev.bgdi.ch in my env
//mf-chsdi3.dev.bgdi.ch/ in my deployed branch
Since we will probably be using these parameters in both cases, and that I could not find the reason for this difference, I did it like that..

src/index.mako.html
gjn: Nothing wrong with it. I just wanted to know where you got it from - or if you came up with it yourself.

gjn:
@rfrancel Very nice. Please squash the commits and merge.

src/index.mako.html
rfrancel: Ah no I found it here: http://stackoverflow.com/questions/11582512/how-to-get-url-parameters-with-javascript/11582513#11582513

rfrancel:
Thank you very much I'll merge when the tests are done

PR #3311, opened 2016-05-13
[WIP] Import redesign

### LINKED ISSUE #3547: local kml is not in selection window when added via drag and drop
reproduce on win10 chrome latest

drag and drop a local kml to viewport

image

expected result

image

work around
move map then the kml appears

mariavkou:
Fix #3547 

This PR redesign the import tool.

It creates new directives : 
- `ga-import` which is a simple container to make the content of the import popup testable.
- `ga-import-local` allows to read a local file (< 20mo).
- `ga-import-online` allows to load a file from a url.
- `ga-import-dnd` which allows to drag'n'drop a file/link then read/load the content dropped.
- `ga-tabs` is a basic directive to add tabs easily.
- `ga-wms-get-cap` parses then renders the content of a WMS GetCapabilties.

[Demo](https://mf-geoadmin3.int.bgdi.ch/kom_import/index.html)

Things to do: 
- [x] Translations
- [ ] Testing
- [ ] Add message when operation is cancelled by user
- [ ] Display parsing error message
- [ ] Add a directive for kml parsing (could wait the validity service)  
- [ ] Fix sauce labs test


gjn:
Only quickly tested. Works for local kml import and wms. But how do I import kml URL?

mariavkourouni:
@gjn kml url import is not integrated, actually this was a question from my side, if it should be integrated and how (with the wms and wmts? which button to press so it will be loaded...). And if it's possible, is there any way to have access to the mockup please: https://docs.google.com/presentation/d/1yR3kYfduTsEO3koIFFe2pdXCYIcSA1eWs6a4n0dIFQ0/edit#slide=id.ge9ace91ee_0_41

gjn:
@davidoesch Can we discuss this Friday?

davidoesch:
yep, book a meet

oterral:
@gjn @cedricmoullet @davidoesch 

Ready to be tested/reviewed . We have the exact same functionnalities as before.

gjn:
Testing on chrome. To me, we have feature parity! Well done.

The upload/parsing/etc button on the online tab does sometimes not make sense when switching between wms url and kml. There can be 'upload ok' after you pasted a new kml url. It's not clear that you need to press the button again.

Doing code review later.

gjn:
Also, needs rebase.

oterral:
The text is back to normal after 10 seconds.

src/components/import/ImportLocalDirective.js
```diff
@@ -112,7 +112,7 @@ goog.require('ga_file_service');
               scope.fileReader = evt.target;
             }
 
-          }).finally(function() {
+          })['finally'](function() {
```
gjn: very strange. IE9 specific? Or linter? or both?

gjn:
Could be back to normal if content of input field changes, no?

oterral:
Yes that makes sense

src/components/import/ImportLocalDirective.js
oterral: linter and IE,  finally is a reserved words

loicgasser:
Very nice. One thing If I can I would suggest, to make the tabs a little bigger and visible while at it.

davidoesch:
Hi feedback
CAN fixes:
- [ ] the width is to large: 
image

-> any proposal on how to minimize it, carriage return in pulldown ( until we have a topic) ?

otherwise the rest seems to work

oterral:
see ngeo import

Issue #3253, opened 2016-03-30
current timestamp support


ltclm:
The current timestamp should be detected as valid timestamp [1].
Current should be supported in the timeslider and it should be displayed / chosen by default if timeBehaviour is 'last' or 'current'.

There is a test database containing  time-enabled pk25.noscale, pk50.noscale and pk100.noscale configuration `bod_ltclm_noscale_master`

[layersconfig](https://mf-chsdi3.dev.bgdi.ch/dev_ltclm_noscale_timeenabled/rest/services/ech/MapServer/layersConfig)

[1] https://github.com/geoadmin/mf-geoadmin3/blob/01f6041d71dbe390738e91e48e1aa1a071d88923/src/components/map/MapService.js#L869

oterral:
Timeslider applies only ' real ' timestamp.

timeBehavior is never 'current' or am I missing something?

ltclm:
current is a recently introduced real timestamp in s3. 
this timestamp should be supported by the timeslider
```
ch.swisstopo.swissimage: {
		attribution: "CNES, Spot Image, swisstopo, NPOC",
		background: true,
		searchable: false,
		format: "jpeg",
		hasLegend: false,
		serverLayerName: "ch.swisstopo.swissimage",
		selectbyrectangle: false,
		attributionUrl: "http://www.swisstopo.admin.ch/internet/swisstopo/de/home.html",
		timeBehaviour: "last",
		tooltip: false,
		label: "SWISSIMAGE",
		highlightable: true,
		chargeable: true,
		timestamps: [
			"current",
			"20151231",
			"20140620",
			"20131107",
			"20130916",
			"20130422",
			"20120809",
			"20120225",
			"20110914"
		],
```

oterral:
When I meant real timestamp I meant a number

ltclm:
@loicgasser @gjn 
The "current" timestamp has to be replaced with a number for the timeslider. Assuming that ``current`` represents the current year we see two options to realize that:

**a) new timeBehaviour ``current``**
this option needs modifications on client and server side. The new layers config response would look like that:
```
ch.swisstopo.swissimage: {
        attribution: "CNES, Spot Image, swisstopo, NPOC",
        background: true,
        searchable: false,
        format: "jpeg",
        hasLegend: false,
        serverLayerName: "ch.swisstopo.swissimage",
        selectbyrectangle: false,
[...]

PR #3314, opened 2016-05-13
Add current timestamp support 

loicgasser:
Fix for https://github.com/geoadmin/mf-geoadmin3/issues/3253


ltclm:
this is just replacing ``current`` with ``currentYear`` for the timeslider label, 
the services identify, wmts, wms, etc. will be used with ``current``, right?

gjn:
I think we need to discuss this in group. Let's do this in the afternoon.

loicgasser:
This is only for the timeslider.

loicgasser:
@gjn Can we accept that as temporary solution?

It fixes all our current use cases and apparently the new use cases require more refactoring back and front end.

oterral:
can we have a test link  ?

gjn:
For me it's ok as temporary solution. As disussed in #3253, we need to analyse this in a broader context.

@oterral Good for you?

oterral:
as you want. I can't test.

PR #3317, opened 2016-05-19
[WIP] iFrame and drawing

cedricmoullet:
This is a POC to propose a way to use iFrame mode as drawing tools. With this approach, the parent window is informed everytime a KML is stored and get the link to the stored KML.

[Test codepen](http://codepen.io/cedricmoullet/pen/XdLRZJ)

### DIFF
--- a/src/components/draw/DrawDirective.js
+++ b/src/components/draw/DrawDirective.js
@@ -807,6 +807,11 @@ goog.require('ga_permalink');
                 layer.id = 'KML||' + layer.url;
               }
 
+              // Inform parent window that a KML has been saved
+              if (top != window) {
+                window.parent.postMessage(layer.url, '*');
+              }
+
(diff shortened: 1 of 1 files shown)


gjn:
What's the use case?

davidoesch:
I get a an error if I want to load https://gist.githubusercontent.com/davidoesch/f9d95a719c87d09c05b8/raw/9901107fedfb281e8a6be91e0039eb14a9fe8bf9/swissmetnet.kml

procrastinatio:
So, we have many demand for communication between iframe and parent window, possibly in both ways.. The goal is to develop another API ?

gjn:
I've update https://github.com/geoadmin/mf-geoadmin3/pull/3199 to include this in a simple message structure. I'll close this to move all discussion over there.

PR #3320, opened 2016-05-23
Fix #3319: Remove feature with bad geometries

### LINKED ISSUE #3319: Drawing writes KML with lines with only one point
Drawing writes KML with lines with only one point
Example

https://s.geo.admin.ch/6bf8508656 

proposal: download a couple of thousands of our KMLs
and test it against http://cite.opengeospatial.org/teamengine/about/kml22/2.2/site/ 

-> find possible other errors

oterral:
Fix #3319 

- Remove bad geometry on import
- Don't add bad  geometry on draw end

[Test](https://mf-geoadmin3.dev.bgdi.ch/fix_3319/?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&X=192880.00&Y=689840.00&zoom=5&dev3d=true&debug&layers=KML%7C%7Chttps:%2F%2Fpublic.geo.admin.ch%2FhxinPHmDT5-iF_5yOkP6Hg)

### DIFF
--- a/src/components/draw/DrawDirective.js
+++ b/src/components/draw/DrawDirective.js
@@ -464,8 +464,37 @@ goog.require('ga_permalink');
             }));
 
             unDrawEvts.push(draw.on('drawend', function(evt) {
+              // Unregister the change event
+              ol.Observable.unByKey(deregFeatureChange);
+              deactivateTool(lastActiveTool);
+              scope.$applyAsync();
+
               var featureToAdd = evt.feature;
               var geom = featureToAdd.getGeometry();
+
+              // According to #3319, it seems LineString can be created with one
+              // point (or with an array of exact same points). If it's the case
+              // we ignore it.
+              if (geom instanceof ol.geom.Polygon) {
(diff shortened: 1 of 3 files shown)


src/components/draw/DrawDirective.js
```diff
@@ -464,8 +464,37 @@ goog.require('ga_permalink');
...
               var featureToAdd = evt.feature;
               var geom = featureToAdd.getGeometry();
+
+              // According to #3319, it seems LineString can be created with one
+              // point (or with an array of exact same points). If it's the case
+              // we ignore it.
+              if (geom instanceof ol.geom.Polygon) {
+                var hasUniqueCoords = true;
+                var coords;
+                geom.getCoordinates()[0].forEach(function(item) {
```
gjn: Similar function is in kml service.

Maybe we can add a kml validation service and put those validations there. Then use it in this directive and in the kml service.

Thoughts?

gjn:
As discussed. Thanks for the fix.

PR #3318, opened 2016-05-20
Assure wms via permalink gets default version

gjn:
Before https://github.com/geoadmin/mf-geoadmin3/pull/2917, any WMS that was imported via the permalink and that didn't have a version specified in the permalink was automatically transformed to a version 1.3.0 wms. The above PR removed this. 

This parameter again assures that 1.3.0 is assumed if the version is left out.

@oterral Quick review/merge please.

PR #3326, opened 2016-05-25
Teo fix deploy

oterral:
Fix regression (explanations [here](https://github.com/geoadmin/mf-geoadmin3/blame/master/src/components/map/WmsService.js#L80)) introduced by https://github.com/geoadmin/mf-geoadmin3/pull/3318

### DIFF
--- a/src/components/StorageService.js
+++ b/src/components/StorageService.js
@@ -91,7 +91,8 @@ goog.require('ga_browsersniffer_service');
         try {
           $window.localStorage.getItem('testkey');
           localStorageSupport = true;
-        } catch (e) {}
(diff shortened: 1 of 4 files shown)


gjn:
My fix broke the wms id in permalink?

oterral:
Yes we already have this problem if the id in the permalink is changed before the re-ordering of layers (this re-ordering happens after all kmls are loaded), the re-ordering does a infinite loop and block the UI..

gjn:
Thanks for the info.

Didn't test all use cases. My bad.

gjn:
Might be worth a selenium test.

Issue #3272, opened 2016-04-10
iframe changes in preview are not taken into account


davidoesch:
go to
https://s.geo.admin.ch/6b2cae30ac

open Embed
go to preview
in preview zoom out and pan

copy result from preview URL window

result:
result in iframe is https://s.geo.admin.ch/6b2cae30ac

expected result:

zoom and pans in preview window are taken into account

oterral:
I can understand the problem (google map iframe doesn't update the link too) but to fix this that needs a dirty communication between iframe and the parent window. I'm not sure we want this.

As workaround we could disallow navigation in the iframe modal

gjn:
http://stackoverflow.com/questions/938180/get-current-url-from-iframe

we should be able to get url of iframe rather easy.

oterral:
Yes but we still need to know when the url has changed and for that we need to listen some events from the child window.
We could detect mouse/keyboard/touch/wheel events on iframe but it seems not very robust.

gjn:
Can't we read it when leaving the preview? No even on that?

gjn:
This is live.

PR #3327, opened 2016-05-25
Iframe copylink changes according to iframe preview

mariavkourouni:
Fix for   https://github.com/geoadmin/mf-geoadmin3/issues/3272

[Testlink](https://mf-geoadmin3.dev.bgdi.ch/kom_iframe/?topic=ech&lang=de&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.geologie-eiszeit-lgm-raster&X=191633.63&Y=759614.00&zoom=10&swipe_ratio=0.59)

### DIFF
--- a/src/components/share/ShareDirective.js
+++ b/src/components/share/ShareDirective.js
@@ -10,6 +10,18 @@ goog.require('ga_permalink');
     'pascalprecht.translate'
   ]);
 
+  function setIframeValue(input) {
+    var previewFrame = angular.element(document).
+                find('#gaEmbedModal iframe')[0];
+              if (previewFrame) {
+                var url = previewFrame.contentWindow.location.href;
+                var frame = input.val();
+                var prFrame = $(frame)[0];
+                prFrame.src = url;
+                input.val(prFrame.outerHTML);
+              }
+  };
+
   module.directive('gaShareCopyInput', function(gaBrowserSniffer, $translate) {
     return {
       restrict: 'A',
@@ -23,6 +35,7 @@ goog.require('ga_permalink');
             }
           }).on({
             focus: function() {
+              setIframeValue(element);
               this.setSelectionRange(0, 9999);
             }
           });
@@ -45,9 +58,12 @@ goog.require('ga_permalink');
         if (!isCopyAllow) {
           element.remove();
         }
+
         // Use clipboard API to copy URL in OS clipboard
         element.on('click', function() {
+
             var inputToCopy = $(attrs.gaShareCopyBt);
+            setIframeValue(inputToCopy);
             inputToCopy[0].setSelectionRange(0, 9999);
 
             // Execute the copy command
--- a/src/components/share/partials/share.html
+++ b/src/components/share/partials/share.html
@@ -92,7 +92,7 @@
           </div>
         </div>
         <iframe ng-if="loadIframe"
-                ng-src="{{embedValue}}" 
+                ng-src="{{embedValue}}"
                 width="{{iframeWidth}}"
                 height="{{iframeHeight}}"
                 frameborder="0">


src/components/share/ShareDirective.js
```diff
@@ -47,6 +47,17 @@ goog.require('ga_permalink');
         }
         // Use clipboard API to copy URL in OS clipboard
         element.on('click', function() {
+
+            // the copy link for the iframe-preview is refreshed
+            // according to the preview changes
+            var previewFrame = document.getElementById('previewFrame');
```
oterral: Don't use id in a directive. The goal of a directive is to make it re-usable.

```
previewFrame = element.find('iframe');
```

src/components/share/partials/share.html
```diff
@@ -91,8 +91,9 @@ <h4 class="modal-title" translate>embed_map</h4>
             </span>
           </div>
         </div>
-        <iframe ng-if="loadIframe"
-                ng-src="{{embedValue}}" 
+        <iframe id="previewFrame"
```
oterral: Remove the id.

src/components/share/ShareDirective.js
```diff
@@ -47,6 +47,17 @@ goog.require('ga_permalink');
...
+            if (previewFrame) {
+              var url = previewFrame.contentWindow.location.href;
+              var myframe = $($('.ga-embed-input')[1].value)[0];
```
oterral: You don't get an iframe here.

gjn:
@mariavkourouni Could you adapt according to the feedback? Would be great to have it for tomorrow.

mariavkourouni:
@gjn @oterral I adapted

gjn:
LGTM.

@oterral If good for you, please merge.

src/components/share/ShareDirective.js
```diff
@@ -47,6 +47,19 @@ goog.require('ga_permalink');
...
         // Use clipboard API to copy URL in OS clipboard
         element.on('click', function() {
+
+            // the copy link for the iframe-preview is refreshed
+            // according to the preview changes
+            var previewFrame;
+            [previewFrame] = angular.element(document).find('iframe');
+            if (previewFrame) {
+              var url = previewFrame.contentWindow.location.href;
+              var frame = $('.ga-embed-input')[1].value;
```
oterral: don't need to use ` .ga-embed-input `. Use the inputToCopy variable declared further.

```
var inputToCopy = $(attrs.gaShareCopyBt);
var previewFrame;
[previewFrame] = angular.element(document).find('iframe');
if (previewFrame) {
    var url = previewFrame.contentWindow.location.href;
    var frame = inputToCopy.val();
.
.
.
.


```

src/components/share/ShareDirective.js
```diff
@@ -47,6 +47,19 @@ goog.require('ga_permalink');
         }
         // Use clipboard API to copy URL in OS clipboard
         element.on('click', function() {
+
+            // the copy link for the iframe-preview is refreshed
+            // according to the preview changes
+            var previewFrame;
+            [previewFrame] = angular.element(document).find('iframe');
```
oterral: No enough robust, if there is another iframe elsewhere in the page. It will break

src/components/share/ShareDirective.js
```diff
@@ -47,6 +47,19 @@ goog.require('ga_permalink');
...
+            if (previewFrame) {
+              var url = previewFrame.contentWindow.location.href;
+              var frame = $('.ga-embed-input')[1].value;
+              var myframe = $(frame)[0];
+              myframe.src = url;
+              $('.ga-embed-input')[1].value = myframe.outerHTML;
```
oterral: same here use inputToCopy variable

src/components/share/ShareDirective.js
```diff
@@ -47,6 +47,19 @@ goog.require('ga_permalink');
...
         element.on('click', function() {
+
+            // the copy link for the iframe-preview is refreshed
+            // according to the preview changes
+            var previewFrame;
+            [previewFrame] = angular.element(document).find('iframe');
+            if (previewFrame) {
+              var url = previewFrame.contentWindow.location.href;
+              var frame = $('.ga-embed-input')[1].value;
+              var myframe = $(frame)[0];
```
oterral: Don't use myXXX as variable name.

src/components/share/ShareDirective.js
mariavkourouni: what would you suggest? should I put back the id?

oterral: Using an id seems inevitable. Use the id of the modal `#gaEmbedModal iframe`for now. I have the feeling that all this piece of code should be elsewhere. and not in this directive.

oterral:
There is still the problem on browser where the copy button doesn't appear. The content of the input text should be modified. 

@mariavkourouni can you try to modify the content of input text on focus evt?

src/components/share/ShareDirective.js
gjn: Why use the find on the `document`? We should use the find on the element itself. Then we assure than only the iframe which is a child of this element, is selcted.

`element.find('iframe')` should be enought (or `angular.element(element).find('iframe')`

oterral: We are in ShareCopyButton directive not in the ShareDirective

gjn: Yes, in that case, we should use the existing id "gaEmbedModal" and go from there. But that's not ideal...

@mariavkourouni Can you adapt?

mariavkourouni: Yes, I will adapt

oterral: @gjn yes not idea. see my comment on PR discussion

src/components/share/ShareDirective.js
```diff
@@ -10,6 +10,19 @@ goog.require('ga_permalink');
...
+  function setIframeValue(input) {
+    var previewFrame = angular.element(document).
+                find('#gaEmbedModal iframe')[0];
+              if (previewFrame) {
+                var url = previewFrame.contentWindow.location.href;
+                var frame = input.val();
+                var prFrame = $(frame)[0];
+                prFrame.src = url;
+                input.val(prFrame.outerHTML);
+                console.log(input.val());
```
gjn: console.log

gjn:
@mariavkourouni Small comment. Also, please squash it.

PR #3331, opened 2016-05-27
Adding retry logic for tiles and images

gjn:
This adresses and potentially fixes https://github.com/geoadmin/mf-geoadmin3/issues/3206

In OL2, the libary supported a number of retries per tiles. This came in handy when the network was not stable (2G, wifi holes, etc). OL3 removed this out of the box. There an PR in ol3 for this https://github.com/openlayers/ol3/pull/5270, but it lacks a little bit of subtility (no way to limit number of retries or to wait between retries).

This PR allows to define the maximum number of retries (currently 100) and the interval between 2 retries (currently 5 seconds). It supports wmts layers and integrated wms layers (single tile and tiled)

To test it, you need to have either a spotty internet connection or work with the developer console (F12 in chrome). Open the network tab and when tiles start loading, cut the internet connection [1] by selecting 'offline' from the dropdown list. The map is not complete and contains white tiles. After a while, re-enable the internet connection (no throttle option) and the tiles/images should appear again.

A couple of questions:
- I'm not sure about the current settings (interval, max_retries)
- Should be tested together with offline -> @oterral 
- Should be tested more with mobile device on the field.

[Testlink](https://mf-geoadmin3.int.bgdi.ch/gjn_reload/index.html?lang=de&topic=ech&bgLayer=ch.swisstopo.swissimage&X=199206.24&Y=689021.88&zoom=2&layers=ch.bakom.mobil-antennenstandorte-gsm,ch.bazl.luftfahrthindernis)

[1]
image

ping @oterral @davidoesch @procrastinatio

### DIFF
--- a/src/components/map/LayerRetryService.js
+++ b/src/components/map/LayerRetryService.js
@@ -1,0 +1,145 @@
+goog.provide('ga_layer_retry_service');
+
+goog.require('ga_map_service');
+
+
+(function() {
+
+  var module = angular.module('ga_layer_retry_service', [
+    'ga_map_service'
+  ]);
+
+ /**
+   * This service provide functionality to enable layers on
+   * an ol3 map to retry failed downloading attempts. This
+   * functionality existed in ol2.
+   */
+  module.provider('gaLayerRetry', function() {
+
+    var canObserve = function(l) {
+      if ((l instanceof ol.layer.Tile ||
+          l instanceof ol.layer.Image) &&
+          (l.getSource() instanceof ol.source.WMTS ||
+           l.getSource() instanceof ol.source.TileWMS ||
+           l.getSource() instanceof ol.source.ImageWMS)) {
+        return true;
+      }
+      return false;
+    };
+
+    var isRetryLayer = function(l) {
+      if (l.type == 'aggregate') {
+        var is = false;
+        angular.forEach(l.getLayers(), function(subLayer) {
+          is = is || isRetryLayer(subLayer);
+        });
+        return is;
+      }
+      return canObserve(l);
+    };
+
+    var reg = {};
+    var MAX_RETRIES = 3;
+    var RETRIES_INTERVAL = 5000;
+
+    this.$get = function($timeout, gaLayerFilters) {
+
+      var registerNewLayers = function(layers) {
+
+        angular.forEach(layers, function(layer) {
+          if (layer.type == 'aggregate') {
+            registerNewLayers(layer.getLayers());
+          } else if (canObserve(layer)) {
+            var end = 'tileloadend';
+            var error = 'tileloaderror';
+            var el = 'tile';
+            var src = layer.getSource();
+            var unreg = [];
+            if (!reg[layer.id]) {
+
+              if (src instanceof ol.source.ImageWMS) {
+                end = 'imageloadend';
+                error = 'imageloaderror';
+                el = 'image';
+              }
+
+              unreg.push(src.on(error, function(evt) {
+                if (evt[el]._retries == undefined) {
+                  evt[el]._retries = 0;
+                }
+                evt[el]._retries++;
+                if (evt[el]._retries <= MAX_RETRIES) {
+                  $timeout(function() {
+                    evt[el].state = 0;
+                    evt[el].load();
+                  }, RETRIES_INTERVAL * evt[el]._retries, false);
+                }
+              }));
+
+              unreg.push(src.on(end, function(evt) {
+                if (evt[el]._retries) {
+                  evt[el]._retries = 0;
+                  this.changed();
+                }
+              }));
+
+              reg[layer.id] = {
+                unregs: unreg,
+                src: src
+              };
+            }
+          }
+        });
+      };
+
+      var unregisterOldLayers = function(layers) {
+
+        var layerActive = function(layerId, ls) {
+          var hasIt = false;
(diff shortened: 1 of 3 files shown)


procrastinatio:
Small remarks:
* Max retries = 100 is way too much
* Generally, we delay further retries, something like 5, 10, 15, 30, 60, etc...

cedricmoullet:
If I remember correctly, we had in a previous life a max retry of 3.

davidoesch:
Will test it on Sunday in field

procrastinatio:
Some test with a toxic proxy, only delaying the response on one wmts host by 20-40s, and limiting the rate on antoher one.
Compared to master, it effectively retries the "missing" tiles, but it keeps asking for tiles, even for zoom levels which are not displayed, and for layers which have been removed (only tested 2D). So after a few minutes, the application is flooding for requests, even if it still state.

gjn:
@procrastinatio Thanks alot for the tests! Very nice.

We should reduce 100 -> 3-5
We should delay further retries as you suggested.
We should abandon retries for layers that are removed from the map.

Going down to zoom level / visible tiles will be hard to do, I guess. I can have a look how easy it could be done.

src/components/map/MapService.js
```diff
@@ -371,12 +371,51 @@ goog.require('ga_urlutils_service');
...
         gaNetworkStatus, gaStorage, gaTileGrid, gaUrlUtils,
         gaStylesFromLiterals, gaGlobalOptions, gaPermalink, 
         gaLang, gaTime) {
 
+      var MAX_RETRIES = 100;
+      var RETRIES_INTERVAL = 5000;
+
+      var retryOnError = function(msg, el, olsource) {
+        if (angular.isDefined(msg) &&
+            angular.isDefined(olsource)) {
```
oterral: Warning `angular.isDefined(null)` returns true.  I'm a fan of using ` if(msg && source)` instead.

src/components/map/MapService.js
```diff
@@ -371,12 +371,51 @@ goog.require('ga_urlutils_service');
...
+          olsource.on(msg, function(event) {
+            if (event[el]._retries == undefined) {
+              event[el]._retries = 0;
+            }
+            event[el]._retries++;
+            if (event[el]._retries <= MAX_RETRIES) {
+              $timeout(function() {
+                event[el].state = 0;
+                event[el].load();
+              }, RETRIES_INTERVAL);
```
oterral: Set the 3rd parameter to false to avoid triggering a digest cycle on each failure: `, RETRIES_INTERVAL, false) `

src/components/map/MapService.js
```diff
@@ -784,6 +830,7 @@ goog.require('ga_urlutils_service');
                 useInterimTilesOnError: gaNetworkStatus.offline,
                 extent: extent
               });
+              loadsuccess = 'tileloadend';
```
oterral: Can't  we have function taking  an olLayer as parameter wichich returns an object with all events :+1: 
```
function  getTIleLoadEvents(olLayer) {
    .
    .
    .
    returns {
      loadError:
      loadSuccess:
      load:   
    }
}

```

src/components/map/MapService.js
```diff
@@ -850,7 +897,8 @@ goog.require('ga_urlutils_service');
               return that.getCesiumImageryProviderById(bodId);
             };
           }
-
+          retryOnError(loaderror, loadel, olSource);
+          refreshAfterRetry(loadsuccess, loadel, olSource, olLayer);
```
oterral: Thinking of external WMS layer, shouldn't we use a general function (in gaMapUtils ? geDefineProperties? ) which register the event on the ol layer in parameter. 

This will make the functionnality more testable too.

procrastinatio:
I think I will install somewhere on of these toxic proxy (various timeout, errors, incomplete response, etc.)  to test all services for real conditions.

gjn:
Yes, this could be helpful.

gjn:
I've refactored this and put the whole logic in an isolated service. Retries is now limited to 3 (as in old ol3) and time between each re-try is increasing. In addition, retries are abandoned when the corresponding layer is removed or invisible.

It lacks tests, but I have similar troubles finding ways to trigger ol3 tile loading events. @oterral Care to review and give hints for the testing? I'd like to have this in - finally.

oterral:
@gjn for testing ol events hav you try to manually dispatch one with the `dispatchEvent` function
https://github.com/openlayers/ol3/blob/bdced571625ef2493377ba661fc48b8c0afa0e12/src/ol/source/urltile.js#L113

oterral:
Just to mention that 204 tiles are considered as an error. And we can't make the difference between a real error or a 204.

I've a proof of concept in this branch how we could [replace the error tile by a canvas](https://github.com/geoadmin/mf-geoadmin3/pull/3662) , we could extend this POC to manage retry in the `ol.GaImageTile` class.

gjn:
Since we abandoned the 204 fix using ol.GaImageTile, do we continue with this? @oterral ?

oterral:
I think this PR can still be useful . 

But I've seen that the offline detection seems broken on that branch. Click on offline checkbox in dev tools should display the offline header

and missing tests.

oterral:
As we don't have 204 anymore, shall we close it?

oterral:
@gjn

Issue #3303, opened 2016-05-04
icon in twitter handle dissapeared 


davidoesch:
Small thing, butwith one of the latest deploy, federal icon in twitter vanished... fb is ok
![unbenannt](https://cloud.githubusercontent.com/assets/4577727/15023462/da11aeb8-1230-11e6-851b-bc5426cb916c.png)

davidoesch:
it is _not_ fixed 
https://twitter.com/swiss_geoportal/status/737533335707082752

![unbenannt3](https://cloud.githubusercontent.com/assets/4577727/15667367/34a33d5c-2716-11e6-86b6-3c99b4d55b50.PNG)

procrastinatio:
Twitter and co need absolute path. And we exclusively deny all robots acces to `dev`and `int`(see https://github.com/geoadmin/mf-geoadmin3/pull/3338)

gjn:
Fixed

PR #3338, opened 2016-05-31
Decongestioned robots.txt path

procrastinatio:
For testing https://github.com/geoadmin/mf-geoadmin3/issues/3303, the twitter bot needs access to geoadmin, even on `dev`and `int`.

Unfortunately, the `robots.txt` is denying everything, even in branches:

```Disallow: /```

This PR sets the `robots.txt`correctly for each branches.

To test, simply remove all robots.txt files from the vhosts `mf-geoadmin3 (both in branches and users directory, remove you very own robot.tx,  restart apache2, praying nobody is building it's project in between, and test you branch with a real robot, eg. https://cards-dev.twitter.com/validator

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -295,6 +295,7 @@ guard-%:
 prd/robots.txt: scripts/robots.mako-dot-txt .build-artefacts/last-deploy-target
 	mkdir -p $(dir $@)
 	${PYTHON_CMD} ${MAKO_CMD} \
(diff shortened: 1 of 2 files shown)


gjn:
No, we don't want robots to parse our dev/int infrastructure. Otherwise, search results will start to pop up in google or other search engines that point to unofficial addresses.

For testing, we need to find other solutions.

PR #3339, opened 2016-05-31
Absolute path for twittercards and related

gjn:
Another try for https://github.com/geoadmin/mf-geoadmin3/issues/3303

As Marc mentioned in his PR, it's hard to test on dev/int because of blocking robots.txt. So we'll test on production...

@oterral Can you quickly review/merge?


oterral:
lgtm

Issue #3350, opened 2016-06-01
Search can't be deactivated


cedricmoullet:
1. Start https://map.geo.admin.ch/?topic=ech&lang=fr&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.pixelkarte-farbe-pk25.noscale,ch.swisstopo.lk25-papierkarte-zusammensetzung.metadata,ch.swisstopo.lk25-papierkarte.metadata
2. Search for "Bottens"
3. Place the cursor in the search field and clear the characters

Result
When everything is deleted,  search results appear and can't be removed anymore.
![image](https://cloud.githubusercontent.com/assets/319676/15716234/bf6b43c4-2821-11e6-8f19-e6ab673e081b.png)

gjn:
Fixed on next deploy.

PR #3351, opened 2016-06-01
Do not trigger a request if query is an empty str

loicgasser:
Fix for https://github.com/geoadmin/mf-geoadmin3/issues/3350

ping @gjn 

newval used to refer to the whole url which was actually never equal to an empty string...

[Test Link](https://mf-geoadmin3.dev.bgdi.ch/gal_emptystring/?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege,ch.swisstopo.pixelkarte-farbe-pk25.noscale,ch.swisstopo.lk25-papierkarte.metadata&layers_visibility=false,false,false,false,true,true&layers_timestamp=18641231,,,,,&X=159577.00&Y=535422.62&zoom=7)

### DIFF
--- a/src/components/search/SearchDirective.js
+++ b/src/components/search/SearchDirective.js
@@ -103,6 +103,8 @@ goog.require('ga_translation_service');
           }
           swisssearchActive = false;
           $rootScope.$broadcast('gaSwisssearchDone');
+        } else {
+          restat.reset();
         }
       };
 
--- a/src/components/search/SearchTypesDirectives.js
+++ b/src/components/search/SearchTypesDirectives.js
@@ -218,7 +218,6 @@ goog.require('ga_urlutils_service');
         var url = gaUrlUtils.append($scope.options.searchUrl,
                                     'type=' + $scope.type);
         url = $scope.typeSpecificUrl(url);
-
         $http.get(url, {
           cache: true,
           timeout: canceler.promise
@@ -300,10 +299,10 @@ goog.require('ga_urlutils_service');
 
       $scope.fuzzy = '';
 
-      $scope.$watch('options.searchUrl', function(newval) {
-        //cancel old requests
+      $scope.$watch('options.searchUrl', function() {
+        // cancel old requests
         cancel();
-        if (newval != '') {
+        if ($scope.options.query != '') {
           blockEvent = false;
           triggerSearch();
         } else {


gjn:
Yes, introduced by search tokens PR a couple of weeks ago. 

Changed to watch searchURL, because otherwise a change to a token string (from `limit: 1` to `limit: 10`) would not trigger a new search as the query variable does _ not_ contain the tokens. So this problem re-appears with this PR. I try to come up with something covering both cases.

loicgasser:
Then keep the watcher on the URL and test that $scope.options.query != ''

gjn:
+1
Am 01.06.2016 8:04 nachm. schrieb Gasser Loïc <notifications@github.com>:Then keep the watcher on the URL and test that $scope.options.query != ''

loicgasser:
That s what I came up with initially

cedricmoullet:
The regression is fixed.
But, when I put 1290, for example, in the search field and then click on
the item, then the result list stay. Maybe another regression.

2016-06-01 20:08 GMT+02:00 Gasser Loïc <notifications@github.com>:

> That s what I came up with initially
>
> —
> You are receiving this because you were assigned.
> <https://github.com/geoadmin/mf-geoadmin3/pull/3351#issuecomment-223077427>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AATgvILnZqOKH9zLcC6dvHJjnmOUHUwtks5qHcq3gaJpZM4IrvQ9>
> .
>



-- 

gjn:
Yes, can reproduce, but only for feature search results in combination with the shop layer in the testling. Adding lubis layerl searching for 2000 an selecting feature does close the search results.
Fascinating.

loicgasser:
yes tested on prod. Seems to behave the same way. not directly related to this issue.

gjn:
Might be related to tooltip popping up (different for shop layers as for other layers). Smells digest cycle / timing issue to me.
Am 01.06.2016 8:45 nachm. schrieb Gasser Loïc <notifications@github.com>:yes tested on prod. Seems to behave the same way. not directly related to this issue.

loicgasser:
It happens if when you have one feature result only.

loicgasser:
The situation is as follow:

When we click a result we always execute a `select` according to the search type.
In each select we change the global search scope using `$scope.options.valueSelected` together with `cleanLabel`.
See for instance:
https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/search/SearchTypesDirectives.js#L463

`$scope.options.valueSelected` then assigns the new cleaned result to `$scope.query` which value is watched and will trigger `startQuery` if the new value is different from the old one.
https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/search/SearchDirective.js#L87ed

But when we type `1208` the cleaned label string and the query string are exactly the same so that `startQuery` is never executed and as a result `reset` method of ResultStats is never called.
https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/search/SearchDirective.js#L136

An easy fix is to ensure that `reset` is called when a value is selected which is what I did.
It works as expected now and it doesn't seem to have any side effect.

loicgasser:
jenkins test

loicgasser:
ok now not that easy swissearch parameter fails. with an address now.

gjn:
Thanks for the explanation. Fix looks good to me.

Didn't understand last comment: swisssearch fails now?

loicgasser:
No it's fine now. I adapted the PR.

loicgasser:
selenium tests ;) finally useful

gjn:
Yes. Long live selenium. Safed us already a few times - mostly for swissearch tests.

Thanks alot for this.

PR #3362, opened 2016-06-06
Real official formula of hiking time

mariavkourouni:
[Demo](http://mf-geoadmin3.dev.bgdi.ch/kom_official_hikTime/)

### DIFF
--- a/src/components/profile/ProfileService.js
+++ b/src/components/profile/ProfileService.js
@@ -195,49 +195,50 @@ goog.require('ga_urlutils_service');
         };
 
         //Hiking time
-        //Official formula: http://www.wandern.ch/download.php?id=4574_62003b89
-        //Reference link: http://www.wandern.ch
+        //Official formula as agreed with the cantons
+        //ASTRA wander.ch FALV CODE
         this.hikingTime = function(data) {
           var wayTime = 0;
           if (data.length != 0) {
             for (var i = 1; i < data.length; i++) {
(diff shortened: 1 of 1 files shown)


davidoesch:
Hi all, before putting this on production: we need to have the to do the test as described in the mail to gilbert. Currently the calculated times are in sync with schweizmobil and swiss map online. So: do not put this in production

mariavkourouni:
@gjn @davidoesch  The '-25% to 25%' formula gives the "best" hiking time results (i.e. closer to the wandern.ch hiking times). Have a look at the mail for all the details.

procrastinatio:
I just did check the proposed hiking time with known and other commonly accepted reference time, and these are very slow I find, something like +20% slower.

mariavkourouni:
Yes actually the official formula is 27% slower than the wander.ch hik.times and the "-25% to 25%" formula is 25% slower than the wander.ch hik.times

gjn:
@davidoesch @mariavkourouni So is this the one to use? Can we merge?

mariavkourouni:
@gjn no, this is not the optimal formula in comparison to the wander.ch hiking times.
The optimal formula is the one which now on prod.

davidoesch:
-> we are awaiting feedback and decision from the different stakeholders. the we have to define WHEN we we would do a switch, if at all.

davidoesch:
close this one thsi has to be solved on mmgt Ebene

Issue #3158, opened 2016-02-20
add what3words in tooltip


davidoesch:
Using http://developer.what3words.com we could add this coordinate info in the position tooltip.
If no what 3 words I'd available show a dash.
Coverage of what 3 words is not yet clear
Swiss search should accept w3w as well....
Any thoughts on that?

vicchi:
Hi @davidoesch - in my day job I'm CTO for what3words; can you clarify what you mean by "coverage is not clear"? We have global coverage for 3 word addresses, so if you're reverse geocoding a long/lat to a 3 word address you should never have an "unavailable" 3 word address (you'll only get an invalid return if you're forwards geocoding a non-existent 3 word address).

davidoesch:
Hi vicchi , by coverage I mean : is there a density /heatmap of w3w for Switzerland available?

vicchi:
Ah! Now I see what you mean. No, there's not such a thing, but even if there was all you'd end up with is a pixellated polygon that corresponds to the country border.

To explain further; when I said in my previous post that we have global coverage, I was being quite literal. The what3words algorithm covers the entire surface of the globe, both land and sea, in a 3m x 3m grid. Each of these 57 trillion squares has a 3 word address assigned. So we have 3 word addresses for every 3m x 3m grid square for the entirety of Switzerland and for every other country in the world, as well as the seas, oceans, rivers, lakes and so on.

From this I hope you can see that a heat map or density map of Switzerland would effectively cover the entirety of the country's area, becoming slightly blocky and pixellated where the border exists.

Hopefully this makes things a little bit clearer and allays your concerns/questions? If not, feel free to ask as many questions as you like, either here or to gary at what3words dot com.

davidoesch:
I have to be more precise:
I'm aware that for every 3x3m you guys have a coordinate, my question was targeting- how many custom 3 words you already sold in switzerland ...

@vicchi: since we run our portal in 5 languages : we need Multingual support, that means w3w in up to 5 languages
- German
- French 
- Italian
- English
- Rumantsch Grischun (a nice to have , see http://www.pledarigrond.ch/rumantschgrischun/ and https://en.wikipedia.org/wiki/Romansh_language#Rumantsch_Grischun

vicchi:
Thanks @davidoesch ...

[...]

PR #3368, opened 2016-06-09
Add w3w coordinate in context popup

gjn:
This is for https://github.com/geoadmin/mf-geoadmin3/issues/3158

The PR adds [What3Words](http://what3words.com/) as location information to the context menu. It's language dependant: the same location has a different 3 word set for each language. Right Mouse Click will show you the context menu.

In addition, the searchbox and swisssearch are capable of interpreting 3 word sets. If someone enters a 3 word set, it will get recognized and the location will be marked in the map.

[Testlink](https://mf-geoadmin3.dev.bgdi.ch/gjn_3words/)
[Testlink with swisssearch](https://mf-geoadmin3.dev.bgdi.ch/gjn_3words/?swisssearch=einordnung.heil.aufstellen)

### DIFF
--- a/src/components/contextpopup/ContextPopupDirective.js
+++ b/src/components/contextpopup/ContextPopupDirective.js
@@ -2,17 +2,20 @@ goog.provide('ga_contextpopup_directive');
 
 goog.require('ga_networkstatus_service');
 goog.require('ga_permalink');
+goog.require('ga_what3words_service');
 (function() {
 
   var module = angular.module('ga_contextpopup_directive', [
     'ga_networkstatus_service',
     'ga_permalink',
+    'ga_what3words_service',
     'pascalprecht.translate'
   ]);
 
   module.directive('gaContextPopup',
       function($rootScope, $http, $translate, $q, $timeout, $window,
-          gaBrowserSniffer, gaNetworkStatus, gaPermalink, gaGlobalOptions) {
+          gaBrowserSniffer, gaNetworkStatus, gaPermalink, gaGlobalOptions,
+          gaLang, gaWhat3Words) {
         return {
           restrict: 'A',
           replace: true,
@@ -35,7 +38,7 @@ goog.require('ga_permalink');
             var map = scope.map;
             var view = map.getView();
 
-            var coord21781;
+            var coord21781, coord4326;
             var popoverShown = false;
 
             var overlay = new ol.Overlay({
@@ -62,6 +65,13 @@ goog.require('ga_permalink');
               return coord + ' ' + zone;
             };
 
+            var updateW3W = function() {
+              gaWhat3Words.getWords(coord4326[1],
+                                    coord4326[0]).then(function(res) {
+                scope.w3w = res;
+              });
+            };
+
             var handler = function(event) {
               if (scope.is3dActive) {
                 return;
@@ -84,7 +94,7 @@ goog.require('ga_permalink');
               coord21781 = (event.originalEvent) ?
                   map.getEventCoordinate(event.originalEvent) :
                   event.coordinate;
-              var coord4326 = ol.proj.transform(coord21781,
+              coord4326 = ol.proj.transform(coord21781,
                   'EPSG:21781', 'EPSG:4326');
               var coord2056 = ol.proj.transform(coord21781,
                   'EPSG:21781', 'EPSG:2056');
@@ -151,6 +161,8 @@ goog.require('ga_permalink');
                   scope.coord2056 = formatCoordinates(coord2056, 2);
                 });
 
+                updateW3W();
+
               });
 
               updatePopupLinks();
@@ -200,6 +212,9 @@ goog.require('ga_permalink');
 
             $rootScope.$on('$translateChangeEnd', function() {
               scope.titleClose = $translate.instant('close');
+              if (popoverShown) {
+                updateW3W();
+              }
             });
 
             // Listen to permalink change events from the scope.
--- a/src/components/contextpopup/partials/contextpopup.html
+++ b/src/components/contextpopup/partials/contextpopup.html
@@ -32,6 +32,10 @@
               <td>MGRS</td>
               <td>{{coordmgrs}}</td>
           </tr>
+          <tr>
+              <td><a href="http://what3words.com/" target="_blank">What 3 Words</a></td>
+              <td>{{w3w}}</td>
+          </tr>
           <tr ng-if="altitude">
               <td translate>elevation</td>
               <td>{{altitude | measure:'distance':[' m']:1}}</td>
--- a/src/components/search/SearchDirective.js
+++ b/src/components/search/SearchDirective.js
@@ -8,6 +8,8 @@ goog.require('ga_search_service');
 goog.require('ga_search_type_directives');
 goog.require('ga_topic_service');
 goog.require('ga_translation_service');
+goog.require('ga_what3words_service');
+
 (function() {
 
   var module = angular.module('ga_search_directive', [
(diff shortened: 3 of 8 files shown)


gjn:
Note: we have a conflict with our technical layer names. E.g. [https://mf-geoadmin3.dev.bgdi.ch/gjn_3words/?swisssearch=ch.swisstopo.zeitreihen](ch.swisstopo.zeitreihen)

gjn:
Adapted so we exclude 3words with 2 or fewer characters. The API itself says that for english, 4 letters is minium length and that for other languages, this might slighly be different. I think with 2, we are on the safe side.

davidoesch:
Fantastic -- works like a charm. 

Question: DESKTOP footer pulldown menu: for consistency we should it use there as well - perfomance of service?

src/components/contextpopup/ContextPopupDirective.js
```diff
@@ -62,6 +65,23 @@ goog.require('ga_permalink');
...
+              $http.get(w3wUrl, {
+                params: {
+                  key: w3wKey,
+                  position: coord4326[1] + ',' + coord4326[0],
+                  lang: gaLang.get()
+                }
+              }).success(function(response) {
+                var words = response.words;
+                if (words && words.length == 3) {
+                  scope.w3w = response.words[0] + '.' +
```
oterral: `scope.w3w = response.words.join('.');`

src/components/contextpopup/ContextPopupDirective.js
```diff
@@ -62,6 +65,23 @@ goog.require('ga_permalink');
...
             };
 
+            var updateW3W = function() {
+              $http.get(w3wUrl, {
+                params: {
+                  key: w3wKey,
+                  position: coord4326[1] + ',' + coord4326[0],
+                  lang: gaLang.get()
+                }
+              }).success(function(response) {
```
oterral: succes is deprecated, use then instead:

```
}).then(function(response) {
  var words = response.data.words;
```

src/components/contextpopup/ContextPopupDirective.js
```diff
@@ -62,6 +65,23 @@ goog.require('ga_permalink');
               return coord + ' ' + zone;
             };
 
+            var updateW3W = function() {
```
oterral: it seems to me that this function should be in a `gaW3W` service to make it testable and reusable.

I've seen you create a `gaSearchW3WAnalyser` which could be the same.

src/components/search/SearchDirective.js
```diff
@@ -151,10 +153,25 @@ goog.require('ga_translation_service');
...
+          // w3w word?
+          var w3w = gaSearchW3WAnalyser.getCoordinate(q);
+
           if (position) {
             gaMapUtils.moveTo($scope.map, $scope.ol3d, 8, position);
             gaMarkerOverlay.add($scope.map, position,
                                 [position, position], true);
+          //TODO: canceling requests that might be out there...
+          } else if (w3w) {
+            w3w.success(function(res) {
```
oterral: `success` is depecated use `then instead` : 
```
w3w.then(function(response) {
  var res = response.data;
```

src/components/contextpopup/partials/contextpopup.html
```diff
@@ -32,6 +32,10 @@
               <td>MGRS</td>
               <td>{{coordmgrs}}</td>
           </tr>
+          <tr>
+              <td><a href="http://what3words.com/">What 3 Words</a></td>
```
oterral: `target='_blank'`

src/components/contextpopup/ContextPopupDirective.js
```diff
@@ -62,6 +65,23 @@ goog.require('ga_permalink');
...
+                  position: coord4326[1] + ',' + coord4326[0],
+                  lang: gaLang.get()
+                }
+              }).success(function(response) {
+                var words = response.words;
+                if (words && words.length == 3) {
+                  scope.w3w = response.words[0] + '.' +
+                              response.words[1] + '.' +
+                              response.words[2];
+                }
```
oterral: We should also manage the case the response conatins an error message like RM : {"error":41,"message":"Invalid language"}`` 

May be displaying  a ` - ` or nothing.

oterral:
Nice feature. I've added minor comments.

gjn:
For desktop footer, it's not usable. Making API request on each move would get us blocked and it will be slow anyway.

gjn:
Yes, might be better to create a gaW3W service that takes care of both use cases. I'll adapt.

Thanks for the review.

procrastinatio:
Shouldn't we support all available languages ? For instance, swisstopo is also something like **бровки.петли.выручка**, or **iğne.nöbet.kanında**.

Then why this particular geocoding schema, there are many out there ? And it looks like the word databases are private, you have to use the API (privacy concern)

davidoesch:
Lang: we  link it to the application lang. 
Schema: correct, it is closed source but with open access via api.
Use case: a lot of huts and stables in the Alps do not have addresses - if you want to locate it: w3w is a good approach. Same for open-air apps- locating your tent with w3w is a use case.
Key: people should be able to directly  convert w3w positions to swiss grid/location. Lets see, if there will be any usage in the future.

gjn:
@oterral I've adapted to your comments.

But I have trouble with the tests. Could you have a look what's wrong with them? I get crazy otherwise...

gjn:
I was able to fix the tests. But I think the gaWhat3Words service could have better tests...but I was not able to with the http mock backend.

oterral:
I'm looking at the test

oterral:
@gjn test fixed

test/specs/What3WordsService.spec.js
```diff
@@ -0,0 +1,81 @@
...
+    beforeEach(function() {
+      
+      module(function($provide) {
+        $provide.value('gaTopic', {
+          get: function() {}
+        });
+        $provide.value('gaLang', {
+          get: function() {
+            return 'custom';
+          }
```
gjn: I missed those. Needed because gaWhat2Words needs them...

test/specs/What3WordsService.spec.js
```diff
@@ -0,0 +1,81 @@
...
+
+    it('gets coordinates', function(done) {
+      $httpBackend.expectGET(testUrl).respond({geometry: {lat: 51.484463, lng: -0.195405}});
+      var res = gaWhat3Words.getCoordinate('first.cörrect.tèst').then(function(resp) {
+        expect(resp.data.geometry.lat).to.eql(51.484463);
+        expect(resp.data.geometry.lng).to.eql(-0.195405);
+        done();
+      });
+      $httpBackend.flush();
+      $rootScope.$digest();
```
gjn: I missed the digest as well...

test/specs/contextpopup/ContextPopupDirective.spec.js
```diff
@@ -16,6 +16,12 @@ describe('ga_contextpopup_directive', function() {
       $provide.value('gaNetworkStatus', {
         offline: true
       });
+
+      $provide.value('gaLang', {
+        get: function() {
+          return 'de';
+        }
+      });
```
gjn: why no need for gaTopic here? gaLang needs gaTopic...

gjn:
Thanks alot. I think I understand them better now :=)

Please merge when you're happy with the architectural adaptions I did.

test/specs/contextpopup/ContextPopupDirective.spec.js
oterral: We load the entire ga module for each test. So the topic request is launched asynchronously.

oterral: Sorry I didn't see you mention the ContextPopup test.  Because here we replace the gaLang service by an object. So the real gaLang service is not loaded.

oterral:
lgtm

Issue #3378, opened 2016-06-21
JSON styling resolution dependent styles


loicgasser:
In the context of https://github.com/geoadmin/mf-chsdi3/pull/2133 we would like to use GeoJSON layers, because

1. Markers are big but point to a very small area -> identify requests can't detect the point if the user click in the middle of the marker.
2. As I already said it markers are big at low zoom levels the map is covered which is not ideal.

It's a good candidate for vector layers, but we would need to adapt the symbology according to the zoom level. (An important work for futrue MVT support as well.)

To do that we need to extend our JSON style parser.

ping @oterral @simonevonheute @AFoletti

loicgasser:
Alternatively we could use layer groups, but that sounds like a hack.

oterral:
OL manages style by resolution so it shouldn't be a big deal.

loicgasser:
I propose the following structure to include the style in the JSON parser for ol styles.

https://github.com/geoadmin/mf-geoadmin3/compare/gal_res_style?expand=1

oterral:
lgtm

PR #3379, opened 2016-06-21
Resolutions support in gaStylesFromLiterals

loicgasser:
https://github.com/geoadmin/mf-geoadmin3/issues/3378

ping @AFoletti You can use this branch for your test with the resolution dependent styles for the bike geojson (you need to change the BOD), it should be ready.
ping @oterral for the review.

### DIFF
--- a/src/components/StylesFromLiteralsService.js
+++ b/src/components/StylesFromLiteralsService.js
@@ -104,16 +104,40 @@ goog.provide('ga_stylesfromliterals_service');
           var values = properties.values;
           for (var i = 0; i < values.length; i++) {
             var value = values[i];
-            olStyle = getOlStyleFromLiterals(value);
-            this.styles[value.geomType][value.value] = olStyle;
+            var minResolution = value.minResolution !== undefined ?
+                value.minResolution : 0;
+            var maxResolution = value.maxResolution !== undefined ?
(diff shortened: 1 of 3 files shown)


src/components/StylesFromLiteralsService.js
```diff
@@ -104,16 +104,40 @@ goog.provide('ga_stylesfromliterals_service');
...
+              minResolution: minResolution,
+              maxResolution: maxResolution
+            };
+            if (!this.styles[value.geomType][value.value]) {
+              this.styles[value.geomType][value.value] = [olStyle];
+            } else {
+              this.styles[value.geomType][value.value].push(olStyle);
+            }
           }
         } else if (this.type === 'range') {
```
oterral: The content of the 2 `else if` seems very close. any chance to avoid this duplication of code ?

oterral:
A minor comment to avoid duplication of code otherwise lgtm.
I've not tested.

src/components/StylesFromLiteralsService.js
loicgasser: Sure, I ll think of something to avoid it.

loicgasser:
Thanks you for the review !

gjn:
Merging this for @AFoletti and @simonevonheute .

PR #3397, opened 2016-07-12
Add support for default EPSG other than 21781

Jenselme:
- Add suppport for secondary EPSG
- Use gaGlobalOptions.defaultEpsg instead of 'EPSG:21781'
- Use gaGlobalOptions.secondaryEpsg instead of 'EPSG:2056'
- Reframe by URL is done by gaGlobalOptions.defaultToSecondaryEpsgUrl
- Allow WMTS layers to be in a different projection than the default by
using an epsg attribute in layersConfig
- Convert permalink to default EPSG in it is contained in secondary EPSG
extent
- In context popop:
- display the coordinates in the default EPSG then in the secondary
EPSG
- use gaGlobalOptions.defaultEpsgContextPopupTitle and
gaGlobalOptions.secondaryEpsgContextPopupTitle to display the correct
projection title

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,11 @@ RESOLUTIONS ?= '[650.0, 500.0, 250.0, 100.0, 50.0, 20.0, 10.0, 5.0, 2.5, 2.0, 1.
 LEVEL_OF_DETAILS ?= '[6, 7, 8, 9, 10, 11, 12, 13, 14, 14, 16, 17, 18, 18]' #lods corresponding to resolutions
 DEFAULT_EPSG ?= EPSG:21781
 DEFAULT_EPSG_EXTEND ?= '[420000, 30000, 900000, 350000]'
+DEFAULT_EPSG_CONTEXT_POPUP_TITLE ?= CH1903 / LV03
+DEFAULT_TO_SECONDARY_EPSG_URL ?= https://geodesy.geo.admin.ch/reframe/lv03tolv95
+SECONDARY_EPSG ?= EPSG:2056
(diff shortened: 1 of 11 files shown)


gjn:
Can you desribe the use-case(s) that you want to support with this?

davidoesch:
context pop up: no changes can be made here ( Keep it simple) - but feel free to build a fork/branch on your side
converting permalink: longterm support (migration) ---no need from the FSDI to do so

To sum: we actually don't have the use case on our side for this PR

Jenselme:
@gjn this allows us to support portal in different EPSG with the same code base just by changing the configuration: we have a client that need portals in EPSG:21781 and one in EPSG:2056

loicgasser:
Makes sense to me as it is just an extra configuration parameter and does not impact UI or anything else.

gjn:
I would prefer that you would keep this downstream. I hope this is ok for you.

In any case, we'll need to switch to LV95 in the future.

davidoesch:
+1 downstream. Lv95 migration is on the do list as a project at IGKS

Gilbert Jeiziner <notifications@github.com> schrieb am Mi., 13. Juli 2016,
21:15:

Jenselme:
@gjn I hoped you'd have merged it to ease the transition to LV95. But if you're not interested, I'll keep it downstream. If you ever change you mind, you know the patch is ;-)

gjn:
Thanks for understanding.

You're PR will be helpful then - and it already showed a flaw, where we don't consistently apply the DEFAUTL_EPSG as defined in the Makefile. We still have hardcoded values...

PR #3399, opened 2016-07-13
Use same staging level as app for WMS

loicgasser:
As discussed, the app now always uses per default the corresponding WMS staging level.
This behaviour can be overriden using `wms_url` parameter.

Further discussion can be found here:
https://github.com/geoadmin/mf-chsdi3/pull/2167

[Use WMS default host](https://mf-geoadmin3.dev.bgdi.ch/gal_wms/?lang=fr&topic=ech&bgLayer=voidLayer&layers=ch.bafu.gewaesserschutz-biologischer_zustand_diatomeen&layers_timestamp=)
[Override WMS default host](https://mf-geoadmin3.dev.bgdi.ch/gal_wms/?wms_url=%2F%2Fwms-bgdi.int.bgdi.ch&lang=fr&topic=ech&bgLayer=voidLayer&layers=ch.bafu.gewaesserschutz-biologischer_zustand_diatomeen&layers_timestamp=)

@gjn @oterral 
What shall we do for the JS API?

### DIFF
--- a/src/index.mako.html
+++ b/src/index.mako.html
@@ -734,7 +734,9 @@ itemscope itemtype="http://schema.org/WebApplication"
           return decodeURIComponent((new RegExp('[?|&]' + name + '=' + '([^&;]+?)(&|#|;|$)').exec(location.search)||[,""])[1].replace(/\+/g, '%20'))||null;
         }
 
-        var wmsUrl = getParam('wms_url');
+        var adminUrlRegexp = new RegExp('${admin_url_regexp}');
+        var wmsUrl = location.protocol + getParam('wms_url');
+        wmsUrl = adminUrlRegexp.test(wmsUrl) ? wmsUrl : null;
 
         var apiParam = getParam('api_url');
         var apiEnv = getParam('api_env');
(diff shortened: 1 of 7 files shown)


src/index.mako.html
```diff
@@ -734,6 +734,8 @@
           return decodeURIComponent((new RegExp('[?|&]' + name + '=' + '([^&;]+?)(&|#|;|$)').exec(location.search)||[,""])[1].replace(/\+/g, '%20'))||null;
         }
 
+        var wmsUrl = getParam('wms_url');
```
gjn: I think we should treat it as api_url/api_env and add wms_urland wms_env.

Specificying a complete URL as parameter opens door for hijacking map.geo.admin.ch byu specifying a foreign wms URL altogether. (map.geo.admin.ch/?wms_url=show.funny.things/wms)

wms_url -> points to branch if specified (as api_url)
wms_env -> uses correct environment (dev int or prod)

(Maybe we should rename to wms_path and wms_staging - and api_path and api_staging)

Alternative to wms_path and wms_staging is to check with regexp that only correct domains are specified. In fact, this might be the best/most flexible solution.

gjn:
@loicgasser Could you address my point?

I think it's best to use a regexpt to validate the URL so it only points to domains we control

loicgasser:
So leave the URL and add a regexp?
I don't really know what you fear as there is no backend intercations.

gjn:
wms url is used to dispaly data in map.geo.admin. So when user can overwrite it via parameter, basically anything can be displayed in map.geo.admin.ch (map.geo.admin.ch/?wms_url=show.funny.things/wms)

Yes, URL with regexp check for correction domains should be enough.

loicgasser:
ok

loicgasser:
done redeploying the branch

gjn:
Perfect, thanks alot. Works as expected.

PR #3402, opened 2016-07-14
[WIP] build once for dev/int/prod

gjn:
As discussed in https://github.com/geoadmin/mf-geoadmin3/pull/3142, this PR tries to configure the project and the index.html so that only one build is required for dev/int/prod. It moves the application configuration (definition of back-ends mainly) from build time to run-time.

This is unfinished work but the basic idea is working: the information in the rc_* files is not needed anymore for the application build. Backend definitions happen on load and depend on the URL under which the appliation is served. Most backends can be overwritten by a parameters for flexibility.

There are a few things missing:
- appcache adaptions (I think they are needed)
- robots.txt needs adaptions as well
- maybe we need to add canonical name in html for SEO
- not all backends have `ci` environment. this needs to be treated.
- during dev, api should not only be staging dependant, but also point to user specific api (ltjeg points to ltjeg api)
- ....others

### DIFF
--- a/src/index.mako.html
+++ b/src/index.mako.html
@@ -34,7 +34,7 @@ itemscope itemtype="http://schema.org/WebApplication"
     <meta name="msapplication-square70x70logo" content="${versionslashed}img/touch-icon-bund-70x70.png"/>
     <meta name="msapplication-square150x150logo" content="${versionslashed}img/touch-icon-bund-150x150.png"/>
     <meta name="msapplication-square310x310logo" content="${versionslashed}img/touch-icon-bund-310x310.png"/>
-    <link rel="dns-prefetch" href="${api_url}"/>
+    <link rel="dns-prefetch" href="https://api3.geo.admin.ch"/>
 
     <link rel="apple-touch-icon" sizes="76x76" href="${versionslashed}img/touch-icon-bund-76x76.png"/>
     <link rel="apple-touch-icon" sizes="120x120" href="${versionslashed}img/touch-icon-bund-120x120.png"/>
@@ -50,8 +50,9 @@ itemscope itemtype="http://schema.org/WebApplication"
     <meta property="og:title" content="Swiss Geoportal"/>
     <meta property="og:description" content="geo.admin.ch ist die Geoinformationsplattform der Schweizerischen Eidgenossenschaft. // geo.admin.ch est la plateforme de géoinformation de la Confédération suisse."/>
     <meta property="og:type" content="WebApplication"/>
-    <meta property="og:url" content="${application_url}"/>
-    <meta property="og:image" content="${application_url}/${versionslashed}img/logo_geoportal.png"/>
+    <meta property="og:url" content=https://map.geo.admin.ch"/>
+    <!-- TODO: assure the logo is there withou version numbers -->
+    <meta property="og:image" content="https://map.geo.admin.ch/img/logo_geoportal.png"/>
 
     <!-- Twitter Card Infos -->
     <meta name="twitter:card" content="summary"/>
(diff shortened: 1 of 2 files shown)


src/index.mako.html
```diff
@@ -721,80 +722,64 @@
...
-        } else {
-          shopUrl = shopUrlBase;
-        }
+        // Setting backends
+        var apiUrl = setBackend('https://api3.geo.admin.ch', 'https://mf-chsdi3.', 'api_url');
+        var apiOverwrite = hasOverWrite;
+        var mapproxyUrl = setBackend('https://wmts{s}.geo.admin.ch', 'https://wmts{s}.', 'mapproxy_url');
+        var wmsUrl = setBackend('https://wms.geo.admin.ch', 'https://wms-bgdi.', 'wms_url');
+        var shopUrl = setBackend('https://shop.swisstopo.admin.ch', 'https://shop-bgdi.', 'shop_url');
+        var publicUrl = setBackend('https://publich.geo.admin.ch', 'https://public.', 'public_url');
```
gjn: typo

src/index.mako.html
```diff
@@ -721,80 +722,64 @@
...
+ 
+        var adminUrlRegexp = new RegExp('${admin_url_regexp}');
         var module = angular.module('geoadmin');
         var cacheAdd = '${version}' != '' ? '/' + '${version}' : '';
         var pathname = location.pathname.replace(/(index|mobile|embed)\.html$/g, '');
-
-        function getParam(name){
-          return decodeURIComponent((new RegExp('[?|&]' + name + '=' + '([^&;]+?)(&|#|;|$)').exec(location.search)||[,""])[1].replace(/\+/g, '%20'))||null;
+        var staging = 'prod';
+        var testres = /^mf-geoadmin3\.([a-z]+)\.bgdi.ch$/.exec(location.hostname);
```
loicgasser: That's ok when served behind apache but in S3 what would the expected behaviour be?
We'll probably have only a unique address so what should be the default env?

src/index.mako.html
```diff
@@ -721,80 +722,64 @@
...
-          } else if (shopUrl == 'prod'){
-              shopUrl = '//beta.shop.swisstopo.ch';
-          } else if (shopUrl == 'puzzle'){
-              shopUrl = '//shop-shopbgdi.openshift.puzzle.ch';
-          }
-        } else {
-          shopUrl = shopUrlBase;
-        }
+        // Setting backends
+        var apiUrl = setBackend('https://api3.geo.admin.ch', 'https://mf-chsdi3.', 'api_url');
```
loicgasser: Shouldn't we put
`//api3.geo.admin.ch`
`//mf-chsdi3.`

in a config file instead?

src/index.mako.html
```diff
@@ -721,80 +722,64 @@
...
-        var shopUrlBase = '${shop_url}';
-
-        if (shopUrl){
-          shopUrl = shopUrl;
-          if (shopUrl == 'dev' || shopUrl == 'int'){
-            shopUrl = '//shop-bgdi.'+ shopUrl + '.bgdi.ch';
-          } else if (shopUrl == 'prod'){
-              shopUrl = '//beta.shop.swisstopo.ch';
-          } else if (shopUrl == 'puzzle'){
-              shopUrl = '//shop-shopbgdi.openshift.puzzle.ch';
```
loicgasser: We won't be able to use this address anymore is it ok?

src/index.mako.html
loicgasser: We also need to handle user env right, so we may want to keep rc_user* files.

src/index.mako.html
```diff
@@ -34,7 +34,7 @@
     <meta name="msapplication-square70x70logo" content="${versionslashed}img/touch-icon-bund-70x70.png"/>
     <meta name="msapplication-square150x150logo" content="${versionslashed}img/touch-icon-bund-150x150.png"/>
     <meta name="msapplication-square310x310logo" content="${versionslashed}img/touch-icon-bund-310x310.png"/>
-    <link rel="dns-prefetch" href="${api_url}"/>
+    <link rel="dns-prefetch" href="https://api3.geo.admin.ch"/>
```
loicgasser: I don't think it's such a good idea to do that if domain names don't match...

src/index.mako.html
```diff
@@ -50,24 +50,25 @@
...
+    <meta name="twitter:image" content="https://map.geo.admin.ch/img/logo_geoportal.png"/>
+    <meta name="twitter:url" content="https://map.geo.admin.ch"/>
 
     <!-- Facebook Specific fb:admins and fb.appid -->
 
     <!-- Google+ Specific Tags -->
     <meta itemprop="name" content="geo.admin.ch"/>
     <meta itemprop="description" content="geo.admin.ch ist die Geoinformationsplattform der Schweizerischen Eidgenossenschaft. // geo.admin.ch est la plateforme de géoinformation de la Confédération suisse."/>
-    <meta itemprop="image" content="${application_url}/${versionslashed}img/logo_geoportal.png"/>
+    <meta itemprop="image" content="https://map.geo.admin.ch/img/logo_geoportal.png"/>
```
loicgasser: so we always point to prod for the images?

loicgasser:
So I propose that we always build with an extra robots file in `prd`, named `robots_prod.txt`

Then at Varnish level when we switch to prod

`map.geo.admin.ch/robots.txt` is redirected to `<s3url>/robots_prod.txt`.
and of course
`map.geo.admin.ch` is redirected to `<s3url>/index.html`.

loicgasser:
Also with this approach `src` is exposed in prod.

src/index.mako.html
gjn: Yes.

gjn:
+1 for robots.txt

src/index.mako.html
gjn: As discussed, `dev`

loicgasser:
I close this one as some of it as already been integrated in:
https://github.com/geoadmin/mf-geoadmin3/pull/3410

PR #3410, opened 2016-07-22
Upload geoadmin to s3

loicgasser:
Based on https://github.com/geoadmin/mf-geoadmin3/pull/3142
Fixes rebase conflicts
Start cleanup and new make targets

ping @procrastinatio


procrastinatio:
@loicgasser This is the wrong branch, there is no `s3manage.py` changes. I think it should be `mom_s3_upload_wip`

loicgasser:
Yep you can cherry pick the wip commit from this branch if you want. Just wanted a working version.

loicgasser:
https://mf-geoadmin3.infra.bgdi.ch/dev/mom_s3_upload_wip/26cd63d/1470157509/index.html
https://mf-geoadmin3.infra.bgdi.ch/dev/mom_s3_upload_wip/26cd63d/1470157509/src/index.html

loicgasser:
@gjn I got the unique build to work locally but I noticed we have a small issue with the caching of the 

topic service and layersConfig services

When we will deploy to int and then to prod we will do things as follow:

1) Deploy chsdi to int
2) Deploy geoadmin to S3 (one build only)

at this point layersConfig and topic service will be cached (e.g. int version)
Therefore geoadmin will always point to these static files when we make the Varnish switch.

The only way to point to these services in prod again would be to redeploy geoadmin, which defies the purpose of making a unique build.

Therefore I don't see any other solution but to get rid of this cache. Do you agree?

gjn:
Ouch, indeed a problem.

**We can't get rid of the cache**, it would break the offline as appcache needs it. Appcache can only store things which run under the same domain. And layersConfig and topic service is needed for the offline mode.

Hmm - we need to come up with something for this.

Maybe for deploy, we always create prod-only configs...not sure.

loicgasser:
I'll try to think of something, maybe we wan upload the cache files separately.

gjn:
That - or we re-write on varnish level...

loicgasser:
Okay I took care of most things, issues, adaptations (90 % ok) and it's rebased:

You can upload a branch using (defaults to master)

`make s3builddeploy DEPLOY_GIT_BRANCH=mom_s3_upload` (of course it won't work with master for now)

Open the output link.

You can list all projects using:

`make s3list`

You can get the info using:

`make s3info S3_VERSION_PATH=<branch/sha/version>` 

You can delete a repo using:

`make s3delete S3_VERSION_PATH=<branch/sha/version>`

It uses the unique build approach we discussed earlier.
Per default it points to dev, you can override this behaviour by using the probably temporary parameter `env=` (for testing purposes as s3 bucket is hosted under infra) and it will point to the desired env.

Specific URL definition for some services still overrides the env parameter.

Everything works as it used to when developing locally in the dedicated users envs.
Which is a good thing.

I added the `robots_prod.txt` file as discussed.

We need to discuss the cache issue and appcache.

Shall I keep the old deploy targets? They are not working as all rc files have been removed. We should maybe only keep the deploydev target and remove the others.

We still need to flush varnish manually, should probably be done when we switch Varnish target.

Please test and comment.

https://mf-geoadmin3.infra.bgdi.ch/mom_s3_upload_wip/778bfd0/1470321268/index.html
https://mf-geoadmin3.infra.bgdi.ch/mom_s3_upload_wip/778bfd0/1470321268/src/index.html

loicgasser:
https://mf-geoadmin3.infra.bgdi.ch/index.html

loicgasser:
Got it, you can test:

```bash
$ make s3builddeployint DEPLOY_GIT_BRANCH=mom_s3_upload
...
Uploading to mom_s3_upload/60fafad/1470933482/src/style/font-awesome-4.5.0/font/fontawesome-webfont.eot - application/vnd.ms-fontobject, gzip: False, cache headers: True
Upload completed at https://mf-geoadmin3.infra.bgdi.ch/mom_s3_upload/60fafad/1470933482/index.html
```
Test

Then:
```bash
$ make make s3activate S3_VERSION_PATH=mom_s3_upload/60fafad/1470933482
...
Please check it on https://mf-geoadmin3.infra.bgdi.ch/index.html
And https://mf-geoadmin3.infra.bgdi.ch/src/index.html
```

Makefile
```diff
@@ -93,28 +114,34 @@ help:
...
-	@echo "- deployint          Deploys snapshot specified with SNAPSHOT=xxx to int."
-	@echo "- deployprod         Deploys snapshot specified with SNAPSHOT=xxx to prod."
-	@echo "- deploydemo         Deploys snapshot specified with SNAPSHOT=xxx to demo."
-	@echo "- deletebranch       List deployed branches or delete a deployed branch (BRANCH_TO_DELETE=...)"
-	@echo "- deploybranch       Deploys current branch to test (note: takes code from github)"
-	@echo "- deploybranchint    Deploys current branch to test and int (note: takes code from github)"
-	@echo "- deploybranchdemo   Deploys current branch to test and demo (note: takes code from github)"
+	@echo "- s3deployint        Deploys a snapshot specified with SNAPSHOT=xxx to s3 int."
+	@echo "- s3deployprod       Deploys a snapshot specified with SNAPSHOT=xxx to s3 prod."
+	@echo "- s3builddeployint   Build branch DEPLOY_GIT_BRANCH (default to 'master') and deploy to S3 int (usage: make s3builddeployint DEPLOY_GIT_BRANCH=branch_to_deploy)"
```
oterral: Is this the equivalent of deploybranchint ?

loicgasser:
Ok I am done with all I can do before having the 2 buckets setup.
Please review.

loicgasser:
@oterral good question, yes I renamed to deploybranch and removed the other on as we are not supposed to deploy a branch to prod bucket.
The only difference is that the branch will be on S3 and the directory is now in your home.

gjn:
I'm testing this a little bit now. So I might overwrite the active version.

scripts/s3manage.py
```diff
@@ -0,0 +1,522 @@
...
+mimetypes.add_type('application/x-font-opentype', '.otf')
+mimetypes.add_type('application/vnd.ms-fontobject', '.eot')
+mimetypes.add_type('application/json', '.json')
+mimetypes.add_type('text/cache-manifest', '.appcache')
+mimetypes.add_type('text/plain', '.txt')
+
+NO_COMPRESS = [
+    'image/png',
+    'image/jpeg',
+    'image/ico',
```
oterral: Isn't it `image/x-icon`  instead?

oterral: And maybe `image/vnd.microsoft.icon` should be also added.

scripts/s3manage.py
```diff
@@ -0,0 +1,522 @@
...
+    'application/x-font-ttf',
+    'application/x-font-opentype',
+    'application/vnd.ms-fontobject',
+    'application/vnd.ms-fontobject']
```
oterral: Duplicated mimetype

src/index.mako.html
```diff
@@ -50,24 +50,24 @@
...
 
     <!-- Twitter Card Infos -->
     <meta name="twitter:card" content="summary"/>
     <meta name="twitter:site" content="@swiss_geoportal"/>
     <meta name="twitter:creator" content="@swiss_geoportal"/>
     <meta name="twitter:title" content="geo.admin.ch"/>
     <meta name="twitter:description" content="geo.admin.ch ist die Geoinformationsplattform der Schweizerischen Eidgenossenschaft. // geo.admin.ch est la plateforme de géoinformation de la Confédération suisse."/>
-    <meta name="twitter:image" content="${application_url}/${versionslashed}img/logo_geoportal.png"/>
-    <meta name="twitter:url" content="${application_url}"/>
+    <meta name="twitter:image" content="${s3basepath}${versionslashed}img/logo_geoportal.png"/>
```
oterral: Is `s3basepath` a relative path ?  if yes I don't think it's working , twitter needs probably the complete url.

src/index.mako.html
```diff
@@ -50,24 +50,24 @@
...
+    <meta name="twitter:url" content="//map.geo.admin.ch"/>
 
     <!-- Facebook Specific fb:admins and fb.appid -->
 
     <!-- Google+ Specific Tags -->
     <meta itemprop="name" content="geo.admin.ch"/>
     <meta itemprop="description" content="geo.admin.ch ist die Geoinformationsplattform der Schweizerischen Eidgenossenschaft. // geo.admin.ch est la plateforme de géoinformation de la Confédération suisse."/>
-    <meta itemprop="image" content="${application_url}/${versionslashed}img/logo_geoportal.png"/>
+    <meta itemprop="image" content="${s3basepath}${versionslashed}img/logo_geoportal.png"/>
```
oterral: Same here

src/index.mako.html
```diff
@@ -712,82 +712,66 @@
...
-        var apiParam = getParam('api_url');
-        var apiEnv = getParam('api_env');
-        var apiUrlBase = '${api_url}';
-
-        if (apiEnv){
-          if (apiEnv == 'prod'){
-            host = '//api3.geo.admin.ch';
-          } else {
-            host = '//mf-chsdi3.' + apiEnv + '.bgdi.ch';
+        pathname = '${s3basepath}' == '' ? pathname : '${s3basepath}';
```
oterral: Are `pathname` and `${s3basepath}`  not equal here ?

src/index.mako.html
```diff
@@ -712,82 +712,66 @@
...
-          shopUrl = shopUrlBase;
-        }
+        var defaultApiUrl = prtl + '${api_url}';
+        var apiUrl = setBackend(defaultApiUrl, prtl + '${api_tech_url}', 'api_url');
+        var mapproxyUrl = setBackend(prtl + '${mapproxy_url}', prtl + '${mapproxy_tech_url}', 'mapproxy_url');
+        var wmsUrl = setBackend(prtl + '${wms_url}',  prtl + '${wms_tech_url}', 'wms_url');
+        var shopUrl = setBackend(prtl + '${shop_url}', prtl + '${shop_tech_url}', 'shop_url');
+        var publicUrl = setBackend(prtl + '${public_url}', prtl + '${public_tech_url}', 'public_url');
+        var printUrl = setBackend(prtl + '${print_url}', prtl + '${print_tech_url}', 'print_url');
+        var apiOverwrite = !!(defaultApiUrl === apiUrl);
```
oterral: you don't need `!!`

src/index.mako.html
```diff
@@ -712,82 +712,66 @@
...
+        var prtl = location.protocol;
+        // Because s3 is hosted on infra
+        var supportedEnv = ['dev', 'int', 'prod', 'ci'];
+        var staging = '${staging}';
+
+        // Determines if one is using custom prodUrl (for user env)
+        var userEnvRegexp = new RegExp('\/\/.*(\/[a-z]*)');
+        // Force env via parameter for testing purposes (before Varnish)
+        var env = getParam('env');
+        function setBackend(prodUrl, techUrl, overWriteParam) {
```
oterral: overwriteParam

oterral:
Just a remark that could be done in another PR. I think all the js inside the index.mako.html file could be in a separate js file (like in the test, see Loader.spec.js)  since it's become bigger and bigger.

gjn:
+1 if sync loading and execution of said script can be garantueed on all browsers.

src/index.mako.html
loicgasser: Yes s3basepath is in the form `/{branch}/{sha}/{version}/`
Good catch. We may put back application URL, but this won't work until the Varnish redirects.

src/index.mako.html
loicgasser: They may not be when in src mode. Basiacally we always want to use `s3basepath` if it's defined.

src/index.mako.html
loicgasser: Indeed

src/index.mako.html
loicgasser: :ok: :

scripts/s3manage.py
loicgasser: I think you're right.

scripts/s3manage.py
loicgasser: :sob:

loicgasser:
Thank you for the review I will address your points.

src/index.mako.html
loicgasser: It's actually because we move index.html to the root. So pathname (dynamic) is subject to changes when we always want to point to the same resources.

loicgasser:
+1 for the loader as well.

src/index.mako.html
gjn: Yes, twitter (and other cards) need full URL

gjn: I would have no qualms to but this image somehwere else in S3 (date.geo or public.geo). It changes so rarely....

loicgasser: yes that would help. When can put it where we put the help.

scripts/s3manage.py
```diff
@@ -0,0 +1,522 @@
...
+        if s3_path.endswith('/'):
+            s3_path = s3_path[:len(s3_path) - 1]
+        if s3_path.count('/') != 2:
+            print('Bad version definition')
+            usage()
+            sys.exit(1)
+
+    bucket_name_env = 'S3_MF_GEOADMIN3_%s' % deploy_target.upper()
+    # TODO Remove me when switch has been performed
+    bucket_name_env = 'S3_MF_GEOADMIN3_INFRA'
```
gjn: If we remove this, we could test the deployment also on all other buckets (meaning, all commands) from the branch. Does this make sense to you?

I'd like to test a couple of things first before merging.

loicgasser: Sure that's the idea. We can't test anything as s3 native addresses are blocked at the moment and I need Andrea to make the necessary Varnish changes.

scripts/deploydev.sh
```diff
@@ -6,7 +6,9 @@ T="$(date +%s)"
 set -o errexit
 
 # adapt these for emergency deploys coming from branches
-GITBRANCH=master
+# TODO Change me back!
+GITBRANCH=mom_s3_upload
```
loicgasser: TO switch back

scripts/s3manage.py
```diff
@@ -0,0 +1,522 @@
...
+                        mimetype = get_file_mimetype(local_file)
+                        save_to_s3(local_file, remote_file, bucket_name, cached=cached, mimetype=mimetype)
+                        # Also upload chsdi metadata file to src folder if available
+                        if is_chsdi_cache:
+                            relative_file_path = relative_file_path.replace(version + '/', '')
+                            remote_file = os.path.join(s3_dir_path, 'src/', relative_file_path, file_name)
+                            save_to_s3(local_file, remote_file, bucket_name, cached=cached, mimetype=mimetype)
+
+    url_to_check = '//mf-geoadmin3.%s.bgdi.ch/' % deploy_target
+    # TODO Remove me!
```
loicgasser: Use int and prod url

scripts/s3manage.py
```diff
@@ -0,0 +1,522 @@
...
+            try:
+                s3client.copy_object(
+                    Bucket=bucket_name,
+                    CopySource=bucket_name + '/' + src_key_name,
+                    Key=j,
+                    CopySourceIfModifiedSince=datetime(2015, 1, 1),
+                    ACL='public-read')
+            except botocore.exceptions.ClientError as e:
+                print('Cannot copy {}: {}'.format(j, e))
+        # TODO Adapt get_url to final URLs
```
loicgasser: TO adapt

scripts/deploydev.sh
gjn: Yes, later. Because I want to test from branch (before merging) a little bit.

scripts/s3manage.py
gjn: We should be able to write onto any bucket from this branch.

Then we can start our varnish tests (dev, int ----merge --- prod)

Plan?

src/index.mako.html
```diff
@@ -51,23 +51,23 @@
     <meta property="og:description" content="geo.admin.ch ist die Geoinformationsplattform der Schweizerischen Eidgenossenschaft. // geo.admin.ch est la plateforme de géoinformation de la Confédération suisse."/>
     <meta property="og:type" content="WebApplication"/>
     <meta property="og:url" content="//map.geo.admin.ch"/>
-    <meta property="og:image" content="${s3basepath}${versionslashed}img/logo_geoportal.png"/>
+    <meta property="og:image" content="//public.geo.admin.ch/img/logo_geoportal.png"/>
```
gjn: I would put https directly in the URL - not sure if cards handle it correct. As well for the og:url part. It's like this currently on prod. I seem to remember troubles if its not full url.

src/index.mako.html
```diff
@@ -51,23 +51,23 @@
...
 
     <!-- Twitter Card Infos -->
     <meta name="twitter:card" content="summary"/>
     <meta name="twitter:site" content="@swiss_geoportal"/>
     <meta name="twitter:creator" content="@swiss_geoportal"/>
     <meta name="twitter:title" content="geo.admin.ch"/>
     <meta name="twitter:description" content="geo.admin.ch ist die Geoinformationsplattform der Schweizerischen Eidgenossenschaft. // geo.admin.ch est la plateforme de géoinformation de la Confédération suisse."/>
-    <meta name="twitter:image" content="//map.geo.admin.ch${versionslashed}img/logo_geoportal.png"/>
+    <meta name="twitter:image" content="//public.geo.admin.ch/img/logo_geoportal.png"/>
     <meta name="twitter:url" content="//map.geo.admin.ch"/>
```
gjn: full URL

src/index.mako.html
```diff
@@ -50,24 +50,24 @@
     <meta property="og:title" content="Swiss Geoportal"/>
     <meta property="og:description" content="geo.admin.ch ist die Geoinformationsplattform der Schweizerischen Eidgenossenschaft. // geo.admin.ch est la plateforme de géoinformation de la Confédération suisse."/>
     <meta property="og:type" content="WebApplication"/>
-    <meta property="og:url" content="${application_url}"/>
-    <meta property="og:image" content="${application_url}/${versionslashed}img/logo_geoportal.png"/>
+    <meta property="og:url" content="//map.geo.admin.ch"/>
+    <meta property="og:image" content="//cms.geo.admin.ch/img/logo_geoportal.png"/>
```
gjn: Please use full URL's here. I think the cards have troubles without fully qualified URLS. Use https. Same as below.

src/index.mako.html
```diff
@@ -50,24 +50,24 @@
...
     <!-- Twitter Card Infos -->
     <meta name="twitter:card" content="summary"/>
     <meta name="twitter:site" content="@swiss_geoportal"/>
     <meta name="twitter:creator" content="@swiss_geoportal"/>
     <meta name="twitter:title" content="geo.admin.ch"/>
     <meta name="twitter:description" content="geo.admin.ch ist die Geoinformationsplattform der Schweizerischen Eidgenossenschaft. // geo.admin.ch est la plateforme de géoinformation de la Confédération suisse."/>
-    <meta name="twitter:image" content="${application_url}/${versionslashed}img/logo_geoportal.png"/>
-    <meta name="twitter:url" content="${application_url}"/>
+    <meta name="twitter:image" content="//cms.geo.admin.ch/img/logo_geoportal.png"/>
+    <meta name="twitter:url" content="//map.geo.admin.ch"/>
```
gjn: Full URLS

src/index.mako.html
```diff
@@ -50,24 +50,24 @@
...
 
     <!-- Facebook Specific fb:admins and fb.appid -->
 
     <!-- Google+ Specific Tags -->
     <meta itemprop="name" content="geo.admin.ch"/>
     <meta itemprop="description" content="geo.admin.ch ist die Geoinformationsplattform der Schweizerischen Eidgenossenschaft. // geo.admin.ch est la plateforme de géoinformation de la Confédération suisse."/>
-    <meta itemprop="image" content="${application_url}/${versionslashed}img/logo_geoportal.png"/>
+    <meta itemprop="image" content="//cms.geo.admin.ch/img/logo_geoportal.png"/>
```
gjn: Same here

src/index.mako.html
```diff
@@ -822,15 +806,15 @@
           gaLayersProvider.wmtsGetTileUrlTemplate = '//wmts{s}.geo.admin.ch/1.0.0/{Layer}/default/{Time}/{TileMatrixSet}/{z}/{y}/{x}.{Format}';
           gaLayersProvider.wmtsMapProxyGetTileUrlTemplate =  gaGlobalOptions.mapproxyUrl + '/1.0.0/{Layer}/default/{Time}/{TileMatrixSet}/{z}/{x}/{y}.{Format}';
           gaLayersProvider.terrainTileUrlTemplate = '//3d.geo.admin.ch/1.0.0/{Layer}/default/{Time}/4326';
-          if (apiEnv || apiParam){
+          if (apiOverwrite){
```
gjn: I have a hunch that something is wrong with this. When we don't specify anything in the URL as parameters, the layersConfig _should_ come from the cached resource (under same domain). Only when we overwrite the api_url via parameter should this be used for the layersConfig. This is confirmed with tests on current integration, where we acutally call the API and not use the cached version under same URL.
This is important for the offline.

Same is true for the topics URL.

gjn: For reference, see current code on master: https://github.com/geoadmin/mf-geoadmin3/blob/master/src/index.mako.html#L825

src/index.mako.html
```diff
@@ -709,82 +709,66 @@
...
-          shopUrl = shopUrlBase;
-        }
+        var defaultApiUrl = prtl + '${api_url}';
+        var apiUrl = setBackend(defaultApiUrl, prtl + '${api_tech_url}', 'api_url');
+        var mapproxyUrl = setBackend(prtl + '${mapproxy_url}', prtl + '${mapproxy_tech_url}', 'mapproxy_url');
+        var wmsUrl = setBackend(prtl + '${wms_url}',  prtl + '${wms_tech_url}', 'wms_url');
+        var shopUrl = setBackend(prtl + '${shop_url}', prtl + '${shop_tech_url}', 'shop_url');
+        var publicUrl = setBackend(prtl + '${public_url}', prtl + '${public_tech_url}', 'public_url');
+        var printUrl = setBackend(prtl + '${print_url}', prtl + '${print_tech_url}', 'print_url');
+        var apiOverwrite = (defaultApiUrl === apiUrl);
```
gjn: For understanding: it's an overwrite, when we overwrite the backend via URL. Not sure that the above is correct.

src/index.mako.html
loicgasser: shall be good now. You can terst it on int again.

src/index.mako.html
```diff
@@ -709,82 +709,66 @@
...
-          shopUrl = shopUrlBase;
-        }
+        var defaultApiUrl = prtl + '${api_url}';
+        var apiUrl = setBackend(defaultApiUrl, prtl + '${api_tech_url}', 'api_url');
+        var mapproxyUrl = setBackend(prtl + '${mapproxy_url}', prtl + '${mapproxy_tech_url}', 'mapproxy_url');
+        var wmsUrl = setBackend(prtl + '${wms_url}',  prtl + '${wms_tech_url}', 'wms_url');
+        var shopUrl = setBackend(prtl + '${shop_url}', prtl + '${shop_tech_url}', 'shop_url');
+        var publicUrl = setBackend(prtl + '${public_url}', prtl + '${public_tech_url}', 'public_url');
+        var printUrl = setBackend(prtl + '${print_url}', prtl + '${print_tech_url}', 'print_url');
+        var apiOverwrite = !!(getParam('api_url') || env || staging == 'dev');
```
gjn: For dev main, we should also take cached version, no?

gjn:
Do we merge and go from there?

src/index.mako.html
loicgasser: not the case at the moment.

loicgasser:
Let me remove the TODO in deploydev then.

gjn:
ok. Do that, then we'll merge and start deploying to int from master. Repeat tests (mainly offline!) on integration and see.

Is @procrastinatio in the office?

loicgasser:
yes

src/index.mako.html
gjn: It is, at least here: 
https://mf-geoadmin3.dev.bgdi.ch/ltjeg/?topic=ech&lang=en&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege&layers_visibility=false,false,false,false&layers_timestamp=18641231,,,

And we build the cache on each build.

It's not the case on dev main now, because in dev main we probably have 
this branch...

I think we should stick to it

loicgasser:
I adapt README for param doc and I merge after?

gjn:
Good to have the Olympian back.

gjn:
I would like to have the cache on dev thing also in it.

loicgasser:
want to change the current behaviour?

loicgasser:
for users defaults -> no cache
for main -> cache
?

loicgasser:
it is not for users. We use apache. try with ltkom for instance.

gjn:
We always used cache for all.

Otherwise, offline would not work on users and main.

gjn:
ltkom also uses mf-geomadin3 cached for layersConfig and topics.

See code in current master: https://github.com/geoadmin/mf-geoadmin3/blob/master/src/index.mako.html#L833
It's always using mf-geaoadmin3 cached url _unless_  api is overwritten by parameter in URL.

Makefile
```diff
@@ -303,7 +340,14 @@ guard-%:
...
 	${PYTHON_CMD} ${MAKO_CMD} \
-	    --var "deploy_target=$(DEPLOY_TARGET)" $< > $@
+	    --var "version=$(VERSION)" \
+	    --var "deploy_target=dev" $< > $@
+
+prd/robots_prod.txt: scripts/robots.mako-dot-txt .build-artefacts/last-deploy-target
+	mkdir -p $(dir $@)
+	${PYTHON_CMD} ${MAKO_CMD} \
+	    --var "version=$(VERSION)" \
+	    --var "deploy_target=prod" $< > $@
```
gjn: For the robots target, we could remove the last-deploy-target pre-requisite (as its hard-coded now). And we should add the last-version prerequiste. The mako itselfl contains the version.

scripts/robots.mako-dot-txt
```diff
@@ -4,8 +4,8 @@ User-agent: *
 
 Disallow: /checker
 
-Sitemap: https://map.geo.admin.ch/sitemap_addresses.xml
-Sitemap: https://map.geo.admin.ch/sitemap_index.xml
+Sitemap: https://api3.geo.admin.ch/${version}/sitemap_addresses.xml
+Sitemap: https://api3.geo.admin.ch/${version}/sitemap_index.xml
```
gjn: Those urls are not correct. Use

Sitemap: https://api3.geo.admin.ch/${version}/sitemap?content=addresses
Sitemap: https://api3.geo.admin.ch/${version}/sitemap?content=index

loicgasser:
ok good to go. I let you check and merge.

scripts/robots.mako-dot-txt
loicgasser: yep

loicgasser:
sitemap urls addressed

Makefile
gjn: This is the only thing remaing I think.

gjn:
On last thing for the build of robots.txt and we're good to go.

Makefile
loicgasser: indeed

loicgasser:
done

loicgasser:
good last catches

gjn:
Ok, let's go with this. I'll merge, then @loicgasser will deploy from scratch to both int and prod buckets.

Then testing on integration.
Then testing with mf-geoadmin3.prod.bgdi.ch on production
Then switch map.geo.admin.ch over.

Keep Andrea in the loop over there.

PR #3434, opened 2016-08-16
Fix #3360: Use the default behavior for the favicon

### LINKED ISSUE #3360: favicon Android
On Android favicon do sometimes not appear

Use https://realfavicongenerator.net/favicon_checker?site=Map.geo.admin.ch&ignore_root_issues=on#.V1CXmMtCRnE

As test

oterral:
Fix #3360 

It seems the problem comes from the `.replaceState` use in gaPermalink service, [see this link](http://stackoverflow.com/questions/24185506/using-history-pushstate-in-firefox-make-my-favicon-disappear) .

Putting the favicon in the root directory is the only way I found to fix this bug.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -512,6 +512,7 @@ prd/embed.html: src/index.mako.html \
 prd/img/: src/img/*
 	mkdir -p $@
 	cp -R $^ $@
+	cp -f $@/favicon.ico $@/../
 
 prd/style/font-awesome-4.5.0/font/: src/style/font-awesome-4.5.0/font/*
 	mkdir -p $@
--- a/scripts/s3manage.py
+++ b/scripts/s3manage.py
@@ -220,7 +220,7 @@ def upload(bucket_name, base_dir, deploy_target):
     upload_directories = ['prd', 'src']
     exclude_filename_patterns = ['.less', '.gitignore', '.mako.']
     root_files = ('index.html', 'mobile.html', 'embed.html',
-                  'robots.txt', 'robots_prod.txt',
+                  'robots.txt', 'robots_prod.txt', 'favicon.ico',
                   'checker', 'geoadmin.%s.appcache' % version)
 
     for directory in upload_directories:
@@ -383,7 +383,7 @@ def activate_version(s3_path, bucket_name, deploy_target):
         files = list(bucket.objects.filter(Prefix='{}/geoadmin.'.format(s3_path)).all())
         if len(files) > 0:
             appcache = os.path.basename(sorted(files)[-1].key)
-        for j in ('robots.txt', 'checker', appcache):
+        for j in ('robots.txt', 'checker', 'favicon.ico', appcache):
             # In prod move robots prod
             src_file_name = 'robots_prod.txt' if j == 'robots.txt' and deploy_target == 'prod' else j
             src_key_name = '{}/{}'.format(s3_path, src_file_name)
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -447,6 +447,7 @@ prd/embed.html: src/index.mako.html \
 prd/img/: src/img/*
 	mkdir -p $@
 	cp -R $^ $@
+	mv $@/favicon.ico $@/../
```
loicgasser: cp -f ?

loicgasser:
Also now, if favicon is at the root you need to define it `s3manage.py` as a root file.

oterral:
Fixed according to comment 
[Test](https://mf-geoadmin3.int.bgdi.ch/teo_fixes/e0594fa/1471875451/favicon.ico)

loicgasser:
Add favicon in activate. Or we upload it once and for all.

loicgasser:
Adapt readme please as well. but +1 for the default GIT_BRANCH
Could you also rename: https://github.com/geoadmin/mf-geoadmin3/blob/master/scripts/clonebuild.sh#L11

oterral:
renamed

Makefile
```diff
@@ -512,6 +511,7 @@ prd/embed.html: src/index.mako.html \
 prd/img/: src/img/*
 	mkdir -p $@
 	cp -R $^ $@
+	cp -f $@/favicon.ico $@/../
```
loicgasser: I think you're supposed to copy from src/img/favicon.ico $@/prd/img/

loicgasser:
`favicon.ico` is not accessible in dev mode. You have to fix that as well.

Makefile
```diff
@@ -70,12 +70,11 @@ USER_NAME ?= $(shell id -un)
 GIT_COMMIT_HASH ?= $(shell git rev-parse --verify HEAD)
 GIT_COMMIT_DATE ?= $(shell git log -1  --date=iso --pretty=format:%cd)
 CURRENT_DATE ?= $(shell date -u +"%Y-%m-%d %H:%M:%S %z")
-DEPLOY_GIT_BRANCH ?= master
 S3_VERSION_PATH ?=
 S3_MF_GEOADMIN3_INFRA = mf-geoadmin3-infra-dublin
 S3_BASE_PATH ?=
 S3_SRC_BASE_PATH ?=
-CLONEDIR = /home/$(USER_NAME)/tmp/branches/${DEPLOY_GIT_BRANCH}
+CLONEDIR = /home/$(USER_NAME)/tmp/branches/${GIT_BRANCH}
```
loicgasser: GIT_BRANCH is taken from `.build-artefacts/last-deploy-git-branch` and therefore will be set to master most of the time.
My take would be to adapt, `clonebuilds.sh` directly instead or default S3DEPLOY_GIT_BRANCH to the current branch.

Makefile
oterral: All img folder is copied into prd/   so I just need to copy the favicon from prd/img to prd/

oterral:
we don't care of the dev mode if the favicon is in the root folder the dev mode doesn't need it.

oterral:
I have removed the git branch thing from this pr since it seems not ready.

Makefile
loicgasser: might as well copy the ones from src/img as well then. Like this we also have the favicon in dev mode.

loicgasser:
I guess we can live with having the favicon coming from dev.

PR #3435, opened 2016-08-16
Add tests

oterral:
- Add tests fo gaLayerFilters service
- Add tests fo gaProfile service (I had to rewrite a bit the service to make it more testable)

[Test](http://mf-geoadmin3.int.bgdi.ch/teo_tests/5112b46/1472455738/index.html)


src/components/map/MapService.js
```diff
@@ -972,13 +972,9 @@ goog.require('ga_urlutils_service');
...
             return false;
           }
-          var hasTooltip;
           var parentLayerId = this.getBodParentLayerId(olLayer);
           var bodId = parentLayerId || olLayer.bodId;
-          if (bodId) {
-            hasTooltip = this.getLayerProperty(bodId, 'tooltip');
-          }
-          return !!(bodId && hasTooltip);
+          return this.getLayerProperty(bodId, 'tooltip');
```
gjn: Doesn't this provoke an uncaught error when bodID is undefined or does not exist in layers array?

oterral: yes but if the test upper is correct the olLayer MUST have an existing bodId.

gjn: Not fully. Theoretically, parentLayerId could be something that does not exist in layers array (it's never checked, just assumed it's correct).

oterral: Yes but in that case that means the layersConfig is wrong so it's important that it triggers an error.

oterral: When we are in map.geo.admin.ch it's reasonable to assume that the layersConfig is fully correct.  The validity of the LayersConfig should be test during compilation or deploy.

gjn: And it is. There's tests for it on mf-chsdi3 side.

I concur :=)

oterral:
jenkins test

loicgasser:
you can assign it to me when you're done I would gladly review,.

oterral:
I'm done .... for today you can review commit by commit . thx

test/specs/map/MapService.spec.js
```diff
@@ -1271,127 +1271,242 @@ describe('ga_map_service', function() {
...
+        gaDefinePropertiesForLayer(layer);
+        layer.displayInLayerManager = true;
+      });
+
+      afterEach(function() {
+        stub.restore();
+      });
+
+      it('keeps queryable and visible bod layers', function() {
+        stub.returns(['attr1']);
```
loicgasser: For my own understanding this means that `gaLayers.getLayerProperty` will always return ['attr1'] ?
`

src/components/profile/ProfileService.js
```diff
@@ -256,11 +269,16 @@ goog.require('ga_urlutils_service');
...
+            var defer = $q.defer();
+            defer.resolve(emptyData);
+            return defer.promise;
+          }
           var coordinates = feature.getGeometry().getCoordinates();
 
           // TODO: manage all kind of geometry
-          if (feature.getGeometry() instanceof ol.geom.Polygon ||
-              feature.getGeometry() instanceof ol.geom.LinearRing) {
+          if (feature && (feature.getGeometry() instanceof ol.geom.Polygon ||
```
loicgasser: check for feature not needed as you added ` if (!feature || !feature.getGeometry())` just above

src/components/profile/ProfileService.js
```diff
@@ -256,11 +269,16 @@ goog.require('ga_urlutils_service');
...
 
         this.get = function(feature) {
+          if (!feature || !feature.getGeometry()) {
+            var defer = $q.defer();
+            defer.resolve(emptyData);
+            return defer.promise;
+          }
           var coordinates = feature.getGeometry().getCoordinates();
 
           // TODO: manage all kind of geometry
```
loicgasser: I am wondering this is mainly for the draw, right? so we don't need to handle multi geoms.

test/specs/Loader.spec.js
```diff
@@ -22,7 +13,8 @@ beforeEach(function() {
     var wmsUrl = '//wms.geo.admin.ch';
     var apacheBasePath = '/';
     var cacheAdd = '/' + version;
-    var pathname = location.pathname.replace(/(index|mobile|embed)\.html$/g, '');
+    var pathname = location.pathname.replace(/(context|index|mobile|embed)\.html$/g, '');
```
loicgasser: why context here?

test/specs/profile/ProfileService.spec.js
```diff
@@ -0,0 +1,251 @@
...
+          expect(profile.elPoints()).to.eql([0, 0]);
+          expect(profile.distance()).to.be(0);
+          expect(profile.hikingTime()).to.be(0);
+          done();
+          spy.restore();
+        });
+        $rootScope.$digest();
+      });
+
+      it('creates an chart from a feature', function(done) {
```
loicgasser: c chart

test/specs/profile/ProfileService.spec.js
```diff
@@ -0,0 +1,251 @@
...
+        $httpBackend.expectPOST(profileUrl).respond(goodResult);
+        gaProfile.create(feature).then(function(profile) {
+          expect(profile).to.be.an(Object);
+          expect(profile.create).to.be.a(Function);
+          expect(profile.update).to.be.a(Function);
+
+          // Test properties
+          expect(profile.data).to.eql(goodResult);
+          expect(profile.elevDiff()).to.be(-0.10000000000002274);
```
loicgasser: This is subject to change with the new node version (and therefore will change soon). Can we check `[0, 0.1]`

src/components/IdentifyService.js
```diff
@@ -95,6 +95,12 @@ goog.provide('ga_identify_service');
           if (order) {
             othersParams.order = order;
           }
+          if (offset) {
+            othersParams.offset = offset;
+          }
```
loicgasser: This probably part of https://github.com/geoadmin/mf-geoadmin3/issues/3447

loicgasser:
I found a regression in the query tool when you first open it in the conditions section.

1. open aviation topic
2. open query tool
3. open conditions

Result:
select attribute label is not displayed

test/specs/map/MapService.spec.js
oterral: Yes until you call `stub.restore()`

src/components/profile/ProfileService.js
oterral: yes you are right

src/components/profile/ProfileService.js
oterral: Maybe in the future in case we allow modification of external KML. It's just a reminder.

test/specs/Loader.spec.js
oterral: phantomjs open http://localhost:8081/context.html to launch test.

test/specs/profile/ProfileService.spec.js
oterral: I prefer to check the exact value. With that we are aware that something has changed in node module.

src/components/IdentifyService.js
oterral: yes

oterral:
I check for the bug. thx for the review @loicgasser

loicgasser:
you should install service-alt :)

oterral:
@loicgasser bug fixed

loicgasser:
LGTM, nice one, please merge

Issue #3441, opened 2016-08-19
Change KML Download Functionality


gjn:
To download a kml drawing, we currently have 2 approaches, depending on browser capabilities:

1) transform blob to local file (chrome, fireforx, Edge)
2) upload to api3 and fetch file from there (IE9/IE10, Safari).

We could replace both approaches by downloading the KML directly from public.geo.admin.ch, where we already store it!

The 'download as kml' functionality was introduced _before_ we started storing all kml's.

This would be an alternative fix for https://github.com/geoadmin/mf-geoadmin3/issues/3440 and would render https://github.com/geoadmin/mf-chsdi3/pull/2206. IT would allow us to remove the downloadkml service from mf-chsdi3 as well.

Marked as high prio as potential fast fix for  https://github.com/geoadmin/mf-geoadmin3/issues/3440

gjn:
The problem with this approache might be that we can't set a filename for the local, downloaded file. Per default it will look like [91S43_1tQsSqBPUVVbFPKw](https://public.geo.admin.ch/91S43_1tQsSqBPUVVbFPKw). It will be exactly as you would click on this link.

loicgasser:
Fixed in prod.

PR #3442, opened 2016-08-19
[WIP] Download kml from public

gjn:
Possible solution for https://github.com/geoadmin/mf-geoadmin3/issues/3441

Not tested, no testlink. Just showing approach as discussed in the issue.

@oterral @loicgasser

### DIFF
--- a/src/components/FileStorageService.js
+++ b/src/components/FileStorageService.js
@@ -81,6 +81,10 @@ goog.provide('ga_filestorage_service');
           return $http['delete'](getServiceUrl(adminId));
         };
 
+        this.getPublicUrl = function(id) {
+          return getPublicUrl(id);
+        };
+
       };
       return new FileStorage();
     };
--- a/src/components/draw/DrawDirective.js
+++ b/src/components/draw/DrawDirective.js
@@ -34,6 +34,20 @@ goog.require('ga_styles_service');
       gaLayerFilters, gaExportKml, gaMapUtils, $document, gaMeasure,
       gaStyleFactory, gaGeomUtils) {
 
+    var pp0 = function(s) {
+      return s.length === 2 ? s : '0' + s;
+    };
+
+    var dateFormat = function(d) {
+      var YYYY = d.getFullYear().toString();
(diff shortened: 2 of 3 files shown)


oterral:
Nice solution. only thing is we looase the name of the KML file `map.geo.admin.ch_KML_20160822091838.kml` is that a problem ?

loicgasser:
@davidoesch ?

davidoesch:
yep, this is an issue if it affects all browser. if onyl safari and IE are affected.:ok

loicgasser:
I mean new file name would be something like `JICZOrHIQ7yIcnGMooVMrQ.kml`
with this solution.

davidoesch:
for all Browsers or only IE and safari?

loicgasser:
For all

davidoesch:
From a user perspective, this is a clear no go - and hard to explain. the source of download is not in the name, hard to remember filename. and bad from marketing. perspective...

-> invest here to get a human readable filenname

loicgasser:
We'll go through CHSDI as before for now.
I don't know how feasible this is, direct download from S3 dosen't let us rename the file.
We'll keep it open and see if you can find a solution for a direct download and pretty file name.

gjn:
Hi

for the local filename, use the 'download="filename.kml' property on the 
<a> tag element. It's html5. Most
browsers support it - only IE9 and safari don't. But that's maybe 
acceptable.

G.

davidoesch:
If it is not supported for IE9 and safari... No issue. +1 for gin proposal

oterral:
It's only  supported by the last  IE EDGE. [see](http://caniuse.com/#search=download)

oterral:
and safari is commonly use.

gjn:
I've updated the PR and deployed it [for testing](mf-geoadmin3.int.bgdi.ch/gjn_kmldown/727bb5d/1473078350/index.html)

gjn:
For some reason, it doesn't work as expected, even in chrome and firefox.

davidoesch:
ok.. and it fails as well in IE9

gjn:
Let's put that to rest.

PR #3445, opened 2016-08-23
Use current branch by default

oterral:
@loicgasser  easier version


gjn:
For my understanding: this is for deploybranch only?

loicgasser:
yes

PR #3453, opened 2016-08-26
Allow multiple local KMLs to be loaded

Jenselme:
Loading more than one local KML caused an exception to be thrown and
broke the KML import tool for both local and distant KMLs.

To reproduce: import a local KML from your hard drive and click on the
'Load KML' button twice or try to load another KML.

This seems to come from the fact that before this commit, once a local
KML was loaded, the fileReader reference was always truethy which means
in $scope.cancel we were trying to abort the upload of an uploaded
file. This caused the exception.

### DIFF
--- a/src/components/importkml/ImportKmlDirective.js
+++ b/src/components/importkml/ImportKmlDirective.js
@@ -135,6 +135,7 @@ goog.require('ga_urlutils_service');
 
             // Read the file
             fileReader.readAsText($scope.file);
+            fileReader = null;
           }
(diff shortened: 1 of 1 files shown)


oterral:
Can't reproduce on chrome . Which config do you use ?

Jenselme:
@gjn Thanks for merging this.

@oterral Can't reproduce on Chrome either but the problem is there in firefox (just reproduced).

oterral:
Yes I've seen thx for the fix.

PR #3455, opened 2016-08-30
Switch to cloudfront backend

loicgasser:
VIP config for S3 has been uploaded at the root.
Not to be merged before tomorrow's deploy.


oterral:
LGTM,

so what do we do with this we merge and deploy it ?

oterral:
@loicgasser @gjn ?

loicgasser:
+1

PR #3458, opened 2016-08-31
Fix profile distance measurement

oterral:
[Test](https://mf-geoadmin3.int.bgdi.ch/teo_fixes/88bcc8a/1472629233/index.html)

@loicgasser quick review please

### DIFF
--- a/src/components/profile/ProfileService.js
+++ b/src/components/profile/ProfileService.js
@@ -133,7 +133,6 @@ goog.require('ga_urlutils_service');
             var denom = maxX >= 10000 ? 1000 : 1;
             this.unitX = maxX >= 10000 ? ' km' : ' m';
             $.map(data, function(val) {
-              val.dist = val.dist / denom;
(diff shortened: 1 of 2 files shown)


cedricmoullet:
Not related to this fix, but when measuring in a lake, the profile looks like empty and not flat:
image

gjn:
Same height everywhere? Has been like that for long time. Not related to draw V2.

cedricmoullet:
@davidoesch please decide if it's a bug

davidoesch:
not a bug it is a feature: lake is flat at 429m

src/components/profile/ProfileService.js
```diff
@@ -133,7 +133,6 @@ goog.require('ga_urlutils_service');
             var denom = maxX >= 10000 ? 1000 : 1;
```
loicgasser: not using it anymore

loicgasser:
one comment LGTM otherwise.

PR #3464, opened 2016-09-02
[WIP] Add image page

gjn:
This PR adds an img page that _could_ serve as basis for map.geo rendered map images (jpg/png) using different sizes and resolutions (dpi)

We have use cases where we need a rendered map, outside of the actual application, as an image:
- print
- non-javascript version of map.geo
- twitter cards
- etc

Print currently uses it's own map rendering engine, which means we need to translate the map.geo map into something that the print server understands - and the print service needs to recompose the map gain.  And to have real Wysiwig, the print rendering engine needs to support the same things as the application itself - something that might not be there when we introduce vector tiles. So I thought: why not use map.geo to create the rendered image?

Anyhow, there are some challenges. As you can see in the testlink, rendering for A0 and 300dpi is extremely slow. Also
- [ ] we would need to extend current solution to store the image somewhere (on S3?) to be consumed with other things.
- [ ] parametrisation (specify dpi and image size) should be possible with permalink parameters.
- [ ] Assure map is loaded before image is created/uploaded
- [ ] Many other things

[In the demo](https://mf-geoadmin3.int.bgdi.ch/gjn_img/4781147/1472819635/img.html?topic=ech&lang=en&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.astra.unfaelle-personenschaeden_alle,ch.bazl.luftfahrthindernis&layers_timestamp=,&X=192243.00&Y=661224.96&zoom=4), you can define some parameters and apply them. To safe the image, just right-click on the map.

Inspiration: http://openlayers.org/en/latest/examples/export-pdf.html
Some other refs:
https://github.com/niklasvh/html2canvas
http://stackoverflow.com/questions/10721884/render-html-to-an-image
https://github.com/tsayen/dom-to-image
http://stackoverflow.com/questions/35480112/canvas-todataurl-download-size-limit#35480643
http://www.html5canvastutorials.com/advanced/html5-canvas-get-image-data-url/
https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toBlob

### DIFF
--- a/src/components/permalink/PermalinkService.js
+++ b/src/components/permalink/PermalinkService.js
@@ -55,6 +55,27 @@ goog.require('ga_urlutils_service');
           var base = b;
           var params = p;
 
+          var getHrefFor = function(p, page) {
+            var pageHtml = page + '.html';
+            var pageRegex = new RegExp(page + '\\.html$');
+            var newParams = angular.extend({}, params);
+            if (angular.isDefined(p)) {
+              angular.extend(newParams, p);
+            }
+            if (angular.isDefined(newParams.mobile)) {
+              delete newParams.mobile;
+            }
+            var basePage = base.replace(/^http:/, 'https:').
+                replace(/(index|mobile)\.html$/, pageHtml);
+            if (!pageRegex.test(basePage)) {
+              if (!/\/$/.test(basePage)) {
+                basePage += '/';
+              }
+              basePage += pageHtml;
+            }
+            return basePage + '?' + gaUrlUtils.toKeyValue(newParams);
+          };
+
           this.getHref = function(p) {
             var newParams = angular.extend({}, params);
             if (angular.isDefined(p)) {
@@ -64,22 +85,11 @@ goog.require('ga_urlutils_service');
           };
 
           this.getEmbedHref = function(p) {
-            var newParams = angular.extend({}, params);
-            if (angular.isDefined(p)) {
-              angular.extend(newParams, p);
-            }
-            if (angular.isDefined(newParams.mobile)) {
-              delete newParams.mobile;
-            }
-            var baseEmbed = base.replace(/^http:/, 'https:').
-                replace(/(index|mobile)\.html$/, 'embed.html');
-            if (!/embed\.html$/.test(baseEmbed)) {
-              if (!/\/$/.test(baseEmbed)) {
-                baseEmbed += '/';
-              }
-              baseEmbed += 'embed.html';
-            }
-            return baseEmbed + '?' + gaUrlUtils.toKeyValue(newParams);
+            return getHrefFor(p, 'embed');
+          };
+
+          this.getImgHref = function(p) {
+            return getHrefFor(p, 'img');
           };
 
           // The main href is the embed permalink but without the name of
--- a/src/components/timestampcontrol/style/timestampcontrol.less
+++ b/src/components/timestampcontrol/style/timestampcontrol.less
@@ -16,6 +16,9 @@
   border-radius: 3px;
 }
 
-.embed .ga-timestamp-control {
-  bottom: 22px;
+.embed, .img {
+  .ga-timestamp-control {
+    bottom: 22px;
+  }
 }
+
--- a/src/index.mako.html
+++ b/src/index.mako.html
@@ -101,6 +101,7 @@ itemscope itemtype="http://schema.org/WebApplication"
           'offline': globals.offline,
           'online': !globals.offline,
           'embed': globals.embed,
+          'img': globals.img,
           'ga-draw-active': globals.isDrawActive,
           'ga-3d-active': globals.is3dActive
         }">
@@ -139,7 +140,7 @@ itemscope itemtype="http://schema.org/WebApplication"
     <div ng-controller="GaSeoController">
       <div ga-seo ga-seo-options="options" ga-seo-map="map"></div>
     </div>
-% if device != 'embed':
+% if device != 'embed' and device != 'img':
     <div ng-cloak translate-cloak id="header" class="navbar navbar-fixed-top" ng-class="{'host-not-prod': !globals.hostIsProd}">
       <a href="?topic={{topicId}}&lang={{langId}}">
         <div id="logo" class="pull-left ga-logo">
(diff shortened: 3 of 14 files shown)


gjn:
ping @davidoesch :)

davidoesch:
theLaw: We are not allowed to provide more than  "500'000 Pixel als statisches Einzelbild" [1] for free as  IMAGE. only PDF A4 A3.
Use case: we had one (1) request froma custoemr to get a static image service -- but since pixel need to be counted - the user has to use sth like phantomJS for stuff like this.
Mf-print 2 allows export as png, but it is not the same approach as above.

Confronting the user with dpi etc: is trick to maintain the scale and aspect ratio for him

finally -- could this be a replacement for mf-print3 ;-) : I know, mf- print can do way more, but if we stick to basic footer and layout. it might be a good replacment for the whole mf-print/tomcat enviro: 
-> cost savings?
-> PDF Quality?

[1]https://www.swisstopo.admin.ch/de/swisstopo/rechtsgrundlagen/nutzung-ohne-einwilligung.html

gjn:
Misunderstanding. This is not for end-clients. It's for intermediary use.

I've listed the use cases above: print, static page image (needed for good SEO), twitter card (imagine have small thumbnail of map representing the exact pl instead of the same logo for all)

davidoesch:
- print: +1 if we can replace mf-print
- static page image for twitter: really nice +1 ( so far I pre- stored them as in https://twitter.com/swiss_geoportal/status/770873066742280192 )

src/components/permalink/PermalinkService.js
```diff
@@ -82,6 +82,26 @@ goog.require('ga_urlutils_service');
...
+            var baseImg = base.replace(/^http:/, 'https:').
+                replace(/(index|mobile)\.html$/, 'img.html');
+            if (!/img\.html$/.test(baseImg)) {
+              if (!/\/$/.test(baseImg)) {
+                baseImg += '/';
+              }
+              baseImg += 'img.html';
+            }
+            return baseImg + '?' + gaUrlUtils.toKeyValue(newParams);
+          };
```
oterral: Lot of duplicate code with getEmbedHref.

Can't we use the same function for both of them with a page name in parameter ?

src/components/timestampcontrol/style/timestampcontrol.less
```diff
@@ -19,3 +19,7 @@
 .embed .ga-timestamp-control {
   bottom: 22px;
 }
+
+.img .ga-timestamp-control {
+  bottom: 22px;
+}
```
oterral:
```
.embed, .img {
   .ga-timestamp-control {
    bottom: 22px;
  }
}
```

src/index.mako.html
```diff
@@ -193,35 +194,59 @@
            ga-attribution-map="::map"
            ga-attribution-ol3d="::ol3d">
       </div>
-  % if device == 'embed':
+  % if device == 'embed' or device == 'img':
```
oterral: This `if` could be moved lin 227

src/index.mako.html
```diff
@@ -193,35 +194,59 @@
...
 
+% if device != 'img':
       <div id='buttonGroup' ng-cloak translate-cloak>
   % if device != 'embed':
         <div ga-geolocation ga-geolocation-map="::map" ga-geolocation-ol3d="::ol3d"></div>
   % endif
         <div id="zoomButtons"></div>
         <div ga-tilt3d ng-if="globals.dev3d"></div>
-  % if device != 'embed':
+  % if device != 'embed' and device != 'img':
```
oterral: `device` is already different from `img` , see line 234

src/index.mako.html
```diff
@@ -193,35 +194,59 @@
...
       <a ng-cloak translate-cloak id="bigMapLink" target="_blank" href="{{toMainHref}}">
         <div>
           <i class="fa fa-ga-logo-bund"></i>
           <span translate-values="{{host}}" translate>view_on_mapgeoadminch</span>
         </div>
         <div ng-show="time">{{time}}</div>
       </a>
+    % endif
+    % if device == 'img':
+      <form id=img-settings class="form">
```
oterral: Missing `"`

src/index.mako.html
```diff
@@ -193,35 +194,59 @@
...
+          <option value="a4" selected>A4</option>
+          <option value="a5">A5 (fast)</option>
+        </select>
+      <!-- <label>Resolution </label> -->
+        <select id="resolution">
+          <option value="75">75 dpi (fast)</option>
+          <option value="150">150 dpi</option>
+          <option value="300">300 dpi (slow)</option>
+        </select>
+        <button id="apply-rendering">Apply</button>
```
oterral: `id`s should be camelcase, but I'm not sure if it's  the case in all the app

src/index.mako.html
```diff
@@ -926,5 +953,39 @@
...
+        var height = Math.round(dim[1] * resolution / 25.4);
+        var size = /** @type {ol.Size} */ (mymap.getSize());
+        var extent = mymap.getView().calculateExtent(size);
+        var oldRes = mymap.getView().getResolution();
+
+        mymap.setSize([width, height]);
+        mymap.getView().fit(extent, /** @type {ol.Size} */ (mymap.getSize()));
+        var newRes = mymap.getView().getResolution();
+        mymap.render();
+        console.log("old res: " + oldRes + " - newRes: " + newRes);
```
oterral: Remove log

src/index.mako.html
```diff
@@ -926,5 +953,39 @@
...
+        var format = document.getElementById('format').value;
+        var resolution = document.getElementById('resolution').value;
+        var dim = dims[format];
+        var width = Math.round(dim[0] * resolution / 25.4);
+        var height = Math.round(dim[1] * resolution / 25.4);
+        var size = /** @type {ol.Size} */ (mymap.getSize());
+        var extent = mymap.getView().calculateExtent(size);
+        var oldRes = mymap.getView().getResolution();
+
+        mymap.setSize([width, height]);
```
oterral:
```
var map = angular.element(document.body).scope().map;
map.setSize([width, height]);
```

src/js/MainController.js
```diff
@@ -78,6 +78,14 @@ goog.require('ga_topic_service');
         viewport.removeClass(dragClass);
       });
 
+      // Expose map gloabally - TODO: remove this
+      $window.mymap = map;
```
oterral: No need of this with the comment l980

src/style/print.less
```diff
@@ -49,6 +49,14 @@
     height: initial !important;
   }
 
+  .img [ga-map] {
```
oterral:
```
.embed, .img {
  [ga-map] {
  }
}
}
```

gjn:
I've rebased this as well. This could serve as backend for a service that creates images out of permlinks. Could be usefull for something like the page overviews with have in our CMS.

src/style/print.less
```diff
@@ -43,12 +43,15 @@
     border: 1px solid lightgray;
   }
 
-  .embed [ga-map] {
-    width: initial !important;
-    top: initial !important;
-    height: initial !important;
+  .embed, img {
```
oterral: .img

danduk82:
ping @gjn and @davidoesch , what do you think? 
 shall we close this PR and re-discuss it in BGDIRE4 print project?

PR #3469, opened 2016-09-05
Use readable version number

gjn:
This changes the version format from seconds since 1970 to better readable timestamp.

Before: `1473087445`
Now: `20160905165747`

It's longer, but at least you get an idea when it was build.

[Testlink](https://mf-geoadmin3.int.bgdi.ch/gjn_v/ad93052/20160905165650/index.html)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ DEPLOY_TARGET ?= dev
 LESS_PARAMETERS ?= -ru
 KEEP_VERSION ?= 'false'
 LAST_VERSION := $(shell if [ -f .build-artefacts/last-version ]; then cat .build-artefacts/last-version 2> /dev/null; else echo '-none-'; fi)
-VERSION := $(shell if [ '$(KEEP_VERSION)' = 'true' ] && [ '$(LAST_VERSION)' != '-none-' ]; then echo $(LAST_VERSION); else date '+%s'; fi)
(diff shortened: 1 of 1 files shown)


gjn:
If the same number of digits is required, we could remove the `20` at the start and the seconds at the end. -> `1609051657`

gjn:
[Testlink](https://mf-geoadmin3.int.bgdi.ch/gjn_v/ad93052/20160905165650/index.html) added

loicgasser:
If we use the first formatting, it would generate the same string as for our snapshot creation.
Would it make sense to adapt it to have the same string for the snaphot and the version?

gjn:
Snapshot version is when snapshot was created, application version is when application was build. But it might cause confusion.

Or we take the 10 digit version, as I mentioned in the comment?

oterral:
The full timestamp seems good to me.

loicgasser:
I would stick to `date '+%Y%m%d%H%M'` or `201609061003`.

My idea was to replace
https://github.com/geoadmin/mf-geoadmin3/blob/master/scripts/deploydev.sh#L15

with `cat .build-artefacts/last-version` after the build.

gjn:
For tomorrow.

gjn:
10 digits as before.

PR #3476, opened 2016-09-08
Fix labels on x axis of the profile

oterral:
[Test](http://mf-geoadmin3.dev.bgdi.ch/teo_fixes/)

Labels on x axis and in the tooltip  were always in meters.


loicgasser:
LGTM

PR #3479, opened 2016-09-08
3d: Set inertia to the Cesium camera

fredj:
depends on #3465

todo:
 - [ ] animate the zoom (zoom buttons click)

Note there's **no** inertia in cesium with the mouse wheel: see https://groups.google.com/d/msg/cesium-dev/g4sO0-67imw/TDFQ21xbM4AJ


loicgasser:
I close as this now part of https://github.com/geoadmin/mf-geoadmin3/pull/3465

PR #3486, opened 2016-09-12
Set the minimumZoomDistance to 2m

fredj:
And move the function out of the callback function

@gjn @davidoesch @cedricmoullet 
Do you want this on master or we keep all the distance stuff modifications in 3dtiles branch ?


fredj:
ok, I'll create a new pull request (the target can't be changed)

PR #3496, opened 2016-09-20
Increase debounce on update

loicgasser:


cedricmoullet:
Could you provide a test link ?

loicgasser:
[TestLink](https://mf-geoadmin3.int.bgdi.ch/gal_profile_debounce/index.html)

loicgasser:
Debounce went from 0.133 sec to 1. Seems reasonable to me.

PR #3499, opened 2016-09-21
Make sure the layer exists

loicgasser:
This PR fixes an unreported JS error.

Step to reproduce:

Open https://map.geo.admin.ch
Click Draw tool
Draw a line
Delete all features
Click back button

Result:
Error -> getSource of null (because layer is set to null)

Expected Result:
No JS error

### DIFF
--- a/src/components/draw/DrawDirective.js
+++ b/src/components/draw/DrawDirective.js
@@ -700,7 +700,11 @@ goog.require('ga_styles_service');
         // create/update the file on s3
         ////////////////////////////////////
         var save = function(evt) {
-          if (layer && layer.getSource().getFeatures().length == 0) {
+          if (!layer {
+            // Do nothing if the layer does not exist
+            return;
+          }
+          if (layer.getSource().getFeatures().length == 0) {
             //if no features to save, delete the file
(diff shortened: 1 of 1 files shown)


src/components/draw/DrawDirective.js
```diff
@@ -700,7 +700,7 @@ goog.require('ga_styles_service');
         // create/update the file on s3
         ////////////////////////////////////
         var save = function(evt) {
-          if (layer.getSource().getFeatures().length == 0) {
+          if (layer && layer.getSource().getFeatures().length == 0) {
```
oterral: `if (!layer) return;`

This function should do nothing if the layer doesn't exist.

loicgasser:
the fix is superficial, as discussed there are other issues that need to be addressed.

oterral:
No I don't think so it fixes the `getSource` bug, and that makes sense that the save functzion does nothing if there is no layer to save.

loicgasser:
Yes indeed, shall we open anther issue concering this "timing" issue when deactivating the draw?

oterral:
Yes pls

Issue #3487, opened 2016-09-12
[catalogs/layertrees] add a layer by keyboard does not work anymore


mariokeusen:
UC/steps to reproduce: open app, open the catalog accordeon, navigate with the arrow keys, try to add a layer to the map with the space key
Result: app does not react on pushing the space tab
Expected: according to the current test book a layer should be added to the map

Question: did we get rid of this feature for some reason?

loicgasser:
Can't reproduce on Chrome, but I can on FF. What did you test with?

davidoesch:
reproduced on ff

mariokeusen:
reproduced on ie11

loicgasser:
Fixed

PR #3500, opened 2016-09-21
Put back keyboard nav on Catalog and Layertree

loicgasser:
Fix for https://github.com/geoadmin/mf-geoadmin3/issues/3487

[Test Link](https://mf-geoadmin3.int.bgdi.ch/gal_keyboardnav/index.html)

Fixes Keyboard nav on FF and IE.
Focused elements are now more obvious as well.

Keyboard nav is impossible when display property is set to None. Therefore I had to adapt a few things.
I decided to manually apply chrome default focus style to all other browsers so that we have the same look and feel across browsers.

ping @oterral for the review.

### DIFF
--- a/src/components/catalogtree/style/catalogtree.less
+++ b/src/components/catalogtree/style/catalogtree.less
@@ -2,6 +2,12 @@
 
   .ga-catalogitem-node {
 
+    button {
+      &:focus {
+        box-shadow: 0 0 3px 0px rgb(77, 144, 254);
+      }
+    }
+
     a {
       color: black;
       display: block;
@@ -60,8 +66,7 @@
       margin: 0;
     }
     input[type=checkbox] {
-      margin-left: .5em;
-      margin-right: .5em;
+      margin-top: .7em;
     }
   }
 
@@ -69,7 +74,7 @@
     margin-top: 0px;
     margin-bottom: 0px;
     padding-left: 0px;
-    
+
     &:not(.ga-catalogitem-selected) label:hover {
       color: #666;
     }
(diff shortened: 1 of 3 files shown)


davidoesch:
@mariokeusen can you test?

src/components/catalogtree/style/catalogtree.less
```diff
@@ -2,6 +2,13 @@
 
   .ga-catalogitem-node {
 
+    button {
+      &:focus {
+        outline: none;
+        box-shadow: 0 0 3px 0px rgb(77, 144, 254);
+      }
+    }
+
```
oterral: What was wrong with the default outline? I found it more visible.

src/components/layermanager/style/layermanager.less
```diff
@@ -71,16 +71,20 @@
...
     width: 100%;
     z-index: 1;
     background-color: #fff;
 
     button {
       position: absolute;
+      &:focus {
+        outline: none;
+        box-shadow: 0 0 3px 0px rgb(77, 144, 254);
+      }
```
oterral: Same here default outline is more visible.

oterral:
Minor comments on `outline` . I think we should keep the default one to be coherent with the rest of the app.

loicgasser:
sure I'll adapt

loicgasser:
and thanks for the review

oterral:
Sorry I meant also the box-shadows were not needed

loicgasser:
They are don't see anything on Firefox or IE otherwise.

oterral:
I see somthing when I focus it with the keyboard but not  with clicking.

Issue #3471, opened 2016-09-06
Marker overlays are highlightable


p1d1d1:
On Chrome:
1) Open map.geo.admin.ch
2) Click the geolocation button
3) Click withing the circle --> a tooltip is shown about the circle itself

oterral:
It's awesome you can get the coordinate of your position !!!!

gjn:
On mobile:
- we remove profile for all features
- we remove popup for gelocation circle

On desktop:
- we remove profile for geolocation circle

For all: adapt title of our own markers.

p1d1d1:
I don't see any coordinate, just perimeter and surface of the circle.
Am I missing something?
Why don't disabling everything also on desktop?

oterral:
Yes the highlight on circle will be deactivated everywhere. But if you click on the point in the middle you have the coordinate.

loicgasser:
@davidoesch https://github.com/geoadmin/mf-geoadmin3/issues/3490 is in contradiction with what has been discussed here. As the draw/measure is not accessible on mobile, the only way one could access this feature at the moment is if you click a kml feature you previously created on desktop.

davidoesch:
Will check it after my holidays

Gasser Loïc <notifications@github.com> schrieb am Do., 22. Sep. 2016, 16:03:

gjn:
will be fixed tomorrow.

PR #3502, opened 2016-09-22
Handle highlightable markers/features

loicgasser:
Fix all the issues described in https://github.com/geoadmin/mf-geoadmin3/issues/3471

From now on:
- all markers must have a proper `label` if they are selectable
- if the marker is unselectable, one must add an `unselectable: true` when creating a new feature.

As discussed:
- profile link is disabled on mobile and from WMS info tooltip (as we don't have the geometry)
- markers have now a proper title (`search`, `crosshair`, `geolocation`)
- Geolocation circle is not selectable

[Test Link](https://mf-geoadmin3.int.bgdi.ch/gal_vector_markers/index.html)

### DIFF
--- a/src/components/tooltip/TooltipDirective.js
+++ b/src/components/tooltip/TooltipDirective.js
@@ -44,7 +44,7 @@ goog.require('ga_topic_service');
             '<div ng-bind-html="html.snippet"></div>' +
             '<div ng-if="::html.showVectorInfos" class="ga-vector-tools">' +
               '<div ga-measure="::html.feature"></div>' +
-              '<div ng-if="::!html.mobile" ' +
+              '<div ng-if="::!html.showProfile" ' +
                    'ga-profile-bt="::html.feature"></div>' +
             '</div>' +
             '<div ga-shop ' +
@@ -699,7 +699,8 @@ goog.require('ga_topic_service');
                 showVectorInfos: (value instanceof ol.Feature),
                 clickGeometry: new ol.geom.Point(scope.clickCoordinate),
                 snippet: $sce.trustAsHtml(html),
-                mobile: gaBrowserSniffer.mobile
+                showProfile: !gaBrowserSniffer.mobile ||
+                    (value instanceof ol.Feature && value.getGeometry())
               });
             };
           }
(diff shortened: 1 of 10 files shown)


src/components/tooltip/TooltipDirective.js
```diff
@@ -44,7 +44,8 @@ goog.require('ga_topic_service');
             '<div ng-bind-html="html.snippet"></div>' +
             '<div ng-if="::html.showVectorInfos" class="ga-vector-tools">' +
               '<div ga-measure="::html.feature"></div>' +
-              '<div ga-profile-bt="::html.feature"></div>' +
+              '<div ng-if="::!html.showProfile" ' +
```
oterral: The display of the button is done by the `ga-profile-bt` directive [see](https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/profile/partials/profile-bt.html).

So I don't think we need this `ng-if` .

src/components/tooltip/TooltipDirective.js
```diff
@@ -540,10 +541,12 @@ goog.require('ga_topic_service');
 
             // Create the html popup for a feature then display it.
             var showVectorFeature = function(feature, layer) {
+              var label = layer.label ||
+                  $translate.instant(feature.getProperties().label);
```
oterral: Is `getProperties` never null or undefined?

src/components/tooltip/TooltipDirective.js
loicgasser: I am referring to https://github.com/geoadmin/mf-geoadmin3/issues/3471#issuecomment-244926101, but I am fine with leaving it. it's not really optimized and our profile is fragile

src/components/tooltip/TooltipDirective.js
loicgasser: no always an object, if empty returns `{}`

oterral: Nice.

loicgasser:
Translations added and rebased.

oterral:
Thx for this.

Issue #3491, opened 2016-09-19
Lubis : all years all loaded when panning


davidoesch:
All mobile OS
Select a specific year in lubis, then pan as shown in

https://goo.gl/photos/6wuFtu56bd9Di5v27

Expected result: filter of year is also applied while panning.
This is really nasty when on a weak mobile connection

PR #3506, opened 2016-09-26
Set no cache for time enabled layers WMTS (use cloudfront instead...)

loicgasser:
Fix for https://github.com/geoadmin/mf-geoadmin3/issues/3491

There is no way to tell ol3 to clear the cache manually. (which is not a huge hack)
Also I did not find a way to disable the animation.

The other work around this bug would be to remove and re-add the layer. But I think this may be really bad.

We can also probably use the default cache size again as
https://github.com/openlayers/ol3/pull/4964
is now merged.

[Test Link](http://mf-geoadmin3.int.bgdi.ch/gal_nocahe_timeenabled/index.html)

### DIFF
--- a/src/components/map/MapService.js
+++ b/src/components/map/MapService.js
@@ -718,9 +718,10 @@ goog.require('ga_urlutils_service');
                 dimensions: {
                   'Time': timestamp
                 },
-                // Temporary until https://github.com/openlayers/ol3/pull/4964
-                // is merged upstream
-                cacheSize: 2048 * 3,
+                // Workaround: Set a cache size of zero when layer is
+                // timeEnabled see:
+                // https://github.com/geoadmin/mf-geoadmin3/issues/3491
+                cacheSize: layer.timeEnabled ? 0 : 2048,
                 projection: gaGlobalOptions.defaultEpsg,
                 requestEncoding: 'REST',
                 tileGrid: gaTileGrid.get(layer.resolutions,


src/components/map/MapService.js
```diff
@@ -720,7 +720,10 @@ goog.require('ga_urlutils_service');
                 },
                 // Temporary until https://github.com/openlayers/ol3/pull/4964
                 // is merged upstream
-                cacheSize: 2048 * 3,
+                // Workaround: Set a cache size of zero when layer is
+                // timeEnabled see:
+                // https://github.com/geoadmin/mf-geoadmin3/issues/3491
+                cacheSize: layer.timeEnabled ? 0 : 2048 * 3,
```
oterral: According to the old comment this cache size is not needed anymore (ol3 update has been merged), so maybe we could set to 0 everytime. 

Anyway you can remove the old comment.

loicgasser: yes I wanted to ask you as I was unsure. We could maybe leave the 2048 default value for all others. I don't know about the side effects for sure.

oterral: I don't know either. We can try and see what's happening :)

oterral:
Remove the old comment and use the default 2048 and you can merge.

oterral:
And thanks for finding this fix

loicgasser:
I updated it, I still think it's good to have at least the default cache until ol3 gives us the possibility to flush the internal cache and benefit from it for all other layers.

oterral:
There is [`source.refresh()`](http://openlayers.org/en/master/apidoc/ol.source.WMTS.html#refresh) method but I'm not sure if this could help us in that case.

gjn:
We loose performance on time enabled layers (our most used layer is time enabled), I'm not a fan of this.

Any way to solve this differently?

loicgasser:
There is a new issue in ol3. https://github.com/openlayers/ol3/issues/5841
No real solution so far, but it may be fixed soon.

Issue #3514, opened 2016-09-29
Adding CadastralWebMap from search is extremely slow


loicgasser:
Step to reproduce:

1. Open https://map.geo.admin.ch?lang=en
2. Search for  Cad
3. Hover on some other layers (Cadastral surveying lots in fact)
3. Add cadastral web map map

Result:
A ton of tiny tiles are downloaded and rendering is extra slow

Expected Result:
Same as adding it via permalink
https://map.geo.admin.ch?layers=ch.kantone.cadastralwebmap-farbe

My take is that we're using the config of another layer..

ping @oterral @gjn

loicgasser:
I think we have in general a huge problem with preview layers. It's affecting the catalog as well. I would fix that ASAP

procrastinatio:
@loicgasser We cascade Cadastrawebmap, we do not host or manage the WMS server. If it is dead slow, there is nothing we can do.

For the rest, I don't know.

loicgasser:
It is solved see PR above

loicgasser:
So not related to preview layers directly but to the resolutions we would pass to our tiled layers. We were altering the initial resolutions definition.

PR #3515, opened 2016-09-29
Fix resolution definition

loicgasser:
FIx for https://github.com/geoadmin/mf-geoadmin3/issues/3514

Bug introduced in
https://github.com/geoadmin/mf-geoadmin3/commit/1820eb9e6febc578ec284e6bf8d2c1d4c70aaccc

I have a hard time believing we had this bug for so long. It's pretty incredible.
I think it explains a lot of bugs we never understood.

So basically we would change the resolution list forever whenever we had a layer with minResolution added. :x:

### DIFF
--- a/src/components/map/MapService.js
+++ b/src/components/map/MapService.js
@@ -19,10 +19,16 @@ goog.require('ga_urlutils_service');
 
   module.provider('gaTileGrid', function() {
     var origin = [420000, 350000];
-    var defaultResolutions = [4000, 3750, 3500, 3250, 3000, 2750, 2500, 2250,
-        2000, 1750, 1500, 1250, 1000, 750, 650, 500, 250, 100, 50, 20, 10, 5,
-        2.5, 2, 1.5, 1, 0.5];
-    var wmsResolutions = defaultResolutions.concat([0.25, 0.1]);
+
+    function getDefaultResolutions() {
+        return [4000, 3750, 3500, 3250, 3000, 2750, 2500, 2250,
+                2000, 1750, 1500, 1250, 1000, 750, 650, 500, 250,
+                100, 50, 20, 10, 5, 2.5, 2, 1.5, 1, 0.5];
+    }
+
+    function getWmsResolutions() {
+        return getDefaultResolutions().concat([0.25, 0.1]);
+    }
 
     function createTileGrid(resolutions, type) {
       if (type === 'wms') {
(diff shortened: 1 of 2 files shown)


loicgasser:
[Test Link](https://mf-geoadmin3.int.bgdi.ch/gal_preview_layers/index.html)

I am even tempted to merge it now lool.

loicgasser:
that possibly explains the terrible perfs we you had all emapis layers as well

src/components/map/MapService.js
```diff
@@ -19,10 +19,16 @@ goog.require('ga_urlutils_service');
 
   module.provider('gaTileGrid', function() {
     var origin = [420000, 350000];
-    var defaultResolutions = [4000, 3750, 3500, 3250, 3000, 2750, 2500, 2250,
-        2000, 1750, 1500, 1250, 1000, 750, 650, 500, 250, 100, 50, 20, 10, 5,
-        2.5, 2, 1.5, 1, 0.5];
-    var wmsResolutions = defaultResolutions.concat([0.25, 0.1]);
```
oterral: I'm not sure to undrstand why this doesn't work the `concat` doesn't affect the defaultResolutions array.

oterral: ah ok I see the problem is the splice further.

src/components/map/MapService.js
```diff
@@ -43,7 +49,8 @@ goog.require('ga_urlutils_service');
       return {
         get: function(resolutions, minResolution, type) {
           if (!resolutions) {
-            resolutions = (type == 'wms') ? wmsResolutions : defaultResolutions;
+            resolutions = (type == 'wms') ? getWmsResolutions() :
+                getDefaultResolutions();
```
oterral: Can you remove the  `resolutions = `    line 58 . It seems useless. Thx.

oterral:
Amazing @loicgasser,  you are the best bug hunter :)  Thx for the fix.

cedricmoullet:
For my understanding: why did it slow the viewer ?

oterral:
At the swiss zoom level, instead of requesting 6 tiles, it requested  hundred of very small tiles at the same time.

src/components/map/MapService.js
oterral: @loicgasser you're right, it works so leave it as it is.

oterral:
Thx

gjn:
`So basically we would change the resolution list forever whenever we had a layer with minResolution added.`

Could you elaborate a little bit more? I don't think I fully understand.

gjn:
I think this went unnoticed for a long time because we have few layers with minResolution that triggers the slicing.

Good find though.

gjn:
A const language feature would have prevented this. Oh wait, there's const in javascript - in ecmascript 6: 
http://es6-features.org/#Constants

PR #3529, opened 2016-10-04
Update angular libs

oterral:
Only major change is the translate-attr directive is standalone now.

[Test](https://mf-geoadmin3.int.bgdi.ch/teo_fixes/index.html)


gjn:
LGTM. Thanks alot.

PR #3535, opened 2016-10-06
Add map loading metricc in debug=true mode

gjn:
This PR adds a new service that surveys the loading of the map. It emits messages in the console that shows timings for map loading events. This can be helpful for debugging purposes and analysing map loading in a little more detail.

Currently, this only works with the `debug=true` parameter in the permalink.

The basic functionality might be used in the future. In general, we might trigger a `maploadstart` and `maploadend` event that could be used for other components.

Go to the [Testlink](https://mf-geoadmin3.int.bgdi.ch/gjn_load/index.html?debug=true&layers=ch.swisstopo.swisstlm3d-wanderwege,ch.bazl.luftfahrthindernis,ch.astra.unfaelle-personenschaeden_alle) and open the console. Then browser around the map (pan/zoom).

### DIFF
--- a/src/components/MapLoadService.js
+++ b/src/components/MapLoadService.js
@@ -1,0 +1,130 @@
+goog.provide('ga_map_load_service');
+
+goog.require('ga_map_service');
+
+(function() {
+
+  var module = angular.module('ga_map_load_service', [
+    'ga_map_service'
+  ]);
+
+ /**
+   * This service can be used to inspect loading of the map.
+   * It times the loading of the map and emits messages about
+   * it
+   */
+  module.provider('gaMapLoad', function() {
+
+    this.$get = function($window, gaLayerFilters) {
+
+      var layersPending = 0;
+      var layersDone = 0;
+      var mapStartTime;
+
+      var updateMapLoading = function() {
+        if (layersPending == layersDone) {
+          var seconds = (new Date() - mapStartTime) / 1000;
+          mapStartTime = undefined;
+          $window.console.info('Map loaded in ' + seconds + 's');
+        }
+      };
+
+      var addLayerPending = function() {
+        if (mapStartTime == undefined) {
+          $window.console.info('Start loading map.');
+          mapStartTime = new Date();
+        }
+        layersPending++;
+        updateMapLoading();
+      };
+
+      var addLayerLoaded = function() {
+        layersDone++;
+        updateMapLoading();
+      };
+
+      var LayerProgress = function(layer) {
+        var src, startTime;
+        var loading = 0;
+        var loaded = 0;
+
+        var update = function() {
+          if (loading == loaded) {
+            var seconds = (new Date() - startTime) / 1000;
+            startTime = undefined;
+            $window.console.log('Layer ' + layer.id +
+                                ' finished loading in ' + seconds + 's.');
+            addLayerLoaded();
+          }
+        };
+
+        var addLoading = function() {
+          if (startTime == undefined) {
+            startTime = new Date();
+            addLayerPending();
+          }
+          loading++;
+          update();
+        };
+        var addLoaded = function() {
+          loaded++;
+          update();
+        };
+
+        if (layer instanceof ol.layer.Tile ||
+            layer instanceof ol.layer.Image) {
+          var start, error, end;
+          src = layer.getSource();
+          if (src instanceof ol.source.WMTS ||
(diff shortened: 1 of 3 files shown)


src/components/MapLoadService.js
```diff
@@ -0,0 +1,136 @@
...
+              start = 'tileloadstart';
+              error = 'tileloaderror';
+              end = 'tileloadend';
+          } else if (src instanceof ol.source.ImageWMS) {
+              start = 'imageloadstart';
+              error = 'imageloaderror';
+              end = 'imageloadend';
+          }
+          if (start) {
+            src.on(start, function() {
```
oterral: `src.on(start, addLoading);`

src/components/MapLoadService.js
```diff
@@ -0,0 +1,136 @@
...
+                   gaLayerFilters.background(l);
+          };
+
+          //React on adding/removing layers
+          scope.$watchCollection('maploadLayers | filter:maploadFilter',
+                                 function(layers) {
+            angular.forEach(layers, function(layer) {
+              if (layer.type == 'aggregate') {
+                angular.forEach(layer.getLayers(), function(l) {
+                  if (!progs.hasOwnProperty(l.id)) {
```
oterral: `!progs[l.id]`

src/components/MapLoadService.js
```diff
@@ -0,0 +1,136 @@
...
+                    progs[l.id] = new LayerProgress(l);
+                  }
+                });
+              } else if (!progs.hasOwnProperty(layer.id)) {
+                progs[layer.id] = new LayerProgress(layer);
+              }
```
oterral: A recursive function will avoid duplicate loop.

src/components/MapLoadService.js
```diff
@@ -0,0 +1,136 @@
...
+      };
+
+      var progs = {};
+
+      var Mapload = function() {
+        this.init = function(scope) {
+          // Only active if debug=true is specified
+          if (gaPermalink.getParams().debug != 'true') {
+            return;
+          }
```
oterral: This `if` could be moved to the mainController. If we have more service like this one in the future we could add it in the same if.

src/js/MainController.js
```diff
@@ -91,6 +93,7 @@ goog.require('ga_topic_service');
     // The main controller creates the OpenLayers map object. The map object
     // is central, as most directives/components need a reference to it.
     $scope.map = createMap();
+    gaMapload.init($scope);
```
oterral: I think putting here the condition `if (gaPermalink.getParams().debug != 'true') ` makes more sense for future services.

src/components/MapLoadService.js
oterral: And it will make the service easier to test.

gjn: For now, I agree. I think this service could be used in non-debug version as well. It might be good to know for some functionality (like when trying the get an image from the map, see https://github.com/geoadmin/mf-geoadmin3/pull/3464) when the map has finished loading.

But we can then switch it back when needed.

Good review, I'll adapt.

gjn:
@oterral Adapted according to your valuable comments. Branch deployed.

gjn:
Ok. Will add broadcasts so it's testable.

gjn:
@oterral How could I mock layer loading events? They don't happen over ajax events and couldn't find something related in openlayers directly. I need to trigger e.g. 'tileloadstart' from ol without actually doing real request. Any idea?

oterral:
@gjn for testing ol events hav you try to manually dispatch one with the dispatchEvent function
https://github.com/openlayers/ol3/blob/bdced571625ef2493377ba661fc48b8c0afa0e12/src/ol/source/urltile.js#L113

gjn:
Not as easy, as I need to have a tile first. How to access tile?

gjn:
I tried to find unit tests in ol3 for this and didn't find any. There's only a test-app that checks if events are triggered, no unit tests.

oterral:
@gjn I've added an example how to test the service.

[Here is the spy api if you want to test more complex thing.](http://sinonjs.org/docs/#spies)

gjn:
Perfect. That's exactly what I needed. Thanks alot.

I'll add tests.

gjn:
Ok, this has been rebased and tests have been added.

@oterral last review and merge?

oterral:
perfect thx for the tests

PR #3538, opened 2016-10-06
WMS: Don't proxy admin urls

loicgasser:
small optimization, I pass it over to the CORS Guru

### DIFF
--- a/src/components/importwms/ImportWmsDirective.js
+++ b/src/components/importwms/ImportWmsDirective.js
@@ -43,14 +43,15 @@ goog.require('ga_wms_service');
               // Kill the current uploading
               $scope.cancel();
 
-              var proxyUrl = gaUrlUtils.proxifyUrl(url);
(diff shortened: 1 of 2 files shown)


src/components/importwms/ImportWmsDirective.js
```diff
@@ -43,14 +43,15 @@ goog.require('ga_wms_service');
               // Kill the current uploading
               $scope.cancel();
 
-              var proxyUrl = gaUrlUtils.proxifyUrl(url);
+              url = gaUrlUtils.isAdminValid(url) && gaUrlUtils.isHttps(url) ?
+                  url : gaUrlUtils.proxifyUrl(url);
```
gjn: I don't see any difference. In proxifyUrl, we already have these tests. What urls gets a proxy that shouldn't?

loicgasser:
Did not check it well enough. Thanks for the fix @gjn

Issue #3536, opened 2016-10-06
KML icons not completely shown


davidoesch:
All Browser all OS

open

https://map.geo.admin.ch/?topic=ech&lang=fr&bgLayer=ch.swisstopo.swissimage&layers=KML%7C%7Chttps:%2F%2Fpublic.geo.admin.ch%2FHfMpusq3QK2fBE72T6VxUw&X=258499.00&Y=577832.38&zoom=7 
result
![image](https://cloud.githubusercontent.com/assets/4577727/19157550/f51b1af8-8be5-11e6-8ffb-ba8d0a05a0a3.png)

not text and icon not shown, might be related to https://github.com/geoadmin/mf-geoadmin3/issues/3531

loicgasser:
clearly a prio one as it affects all icons of all kmls when loaded via permalink

loicgasser:
For some reasons it is due to this commit
https://github.com/geoadmin/mf-geoadmin3/commit/dc76b58b9d768707b0d563dd6b4bf319eb7567d0

Why did we merge that guys with the emergency deploy.

loicgasser:
https://mf-geoadmin3.int.bgdi.ch/teo_up/index.html?%3Ftopic=ech&lang=fr&bgLayer=ch.swisstopo.swissimage&X=258499.00&Y=577832.38&zoom=7&topic=ech&layers=KML%7C%7Chttps:%2F%2Fpublic.geo.admin.ch%2FHfMpusq3QK2fBE72T6VxUw

That's the deployed branch of this pull request
https://github.com/geoadmin/mf-geoadmin3/pull/3528

gjn:
I merged the angulare update this morning.
Emergency deploy should have ben based on r_161005 tag. I thought you did that? @oterral?
Generally, emergency deploys  should always come from branches based on release tag.
Not good.
Am 06.10.2016 8:31 nachm. schrieb Gasser Loïc <notifications@github.com>:For some reasons it is due to this commit
dc76b58

Why did we merge that guys with the emergency deploy.

loicgasser:
Don't take into account my previous comments

well it seems to be the case in fact:
https://github.com/geoadmin/mf-geoadmin3/compare/r_161005...r_161005_hotfixes

I don't know what's going on, you can confirm but: (`make cleanall user` each time)

[...]

PR #3542, opened 2016-10-07
Disable kml extent to display icons

gjn:
Fix for #3536 on hotfix branch


gjn:
This is the PR based on the fix by @loicgasser . I will apply it to the current hotfix branch and re-deploy this branch.

PR #3552, opened 2016-10-19
Fix #3548: Always returns a good extent

### LINKED ISSUE #3548: Print -->  vector layer with a map rotation doesn't work
Solution : 
replace: 
    `geometry.intersectsExtent(getPrintRectangleCoords())`

by: 
    `ol.extent.getIntersection(getPrintRectangleCoords(), geometry.getExtent())`

oterral:


oterral:
Tested

PR #3562, opened 2016-10-21
Fixes

### LINKED ISSUE #3471: Marker overlays are highlightable

### LINKED ISSUE #3482: Draw line - add profile  on finish
User feedback
" why do I not get a profile when I draw a line?"
-> obviously users don't get the difference between line and measure

Proposal
A) we add the profile as for measure upon finishing digitizing
or
B) we add a "profile" link 
image

### LINKED ISSUE #3539: Mobile: Infobox is not in front
All devices All Os
open map.geo.admin.ch in mobile mode
open an infobox

result:
infobox layer is in back
image

expected result:
infobox layer is in front

oterral:
[Test](https://mf-geoadmin3.int.bgdi.ch/teo_fixes/index.html)

Fix #3539, test popups:

On desktop : All popups are under modals and the popup is put on front when it has the focus.

On mobile: All popups are on top of the menu except the tooltip popup which is under.

Fix #3482, test draw:

Draw a Line or a polygon then the profile is displayed automatically at the end of the draw.

Fix #3471, test draw:

Open this [link](https://mf-geoadmin3.int.bgdi.ch/teo_fixes/mobile.html?topic=ech&lang=fr&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege,KML%7C%7Chttps:%2F%2Fpublic.geo.admin.ch%2FZve1GElEQImi17-pls3W5w&layers_visibility=false,false,false,false,true&layers_timestamp=18641231,,,,&mobile=true&X=200760.00&Y=704525.74&zoom=5), click on the line and you should see the 'Show profile' link in the tooltip.

### DIFF
--- a/src/components/popup/PopupDirective.js
+++ b/src/components/popup/PopupDirective.js
@@ -11,12 +11,22 @@ goog.require('ga_print_service');
   ]);
 
   module.directive('gaPopup',
-    function($rootScope, $translate, $window, gaBrowserSniffer,
-        gaPrint) {
-      var zIndex = 2000;
+    function($rootScope, $translate, $window, gaBrowserSniffer, gaPrint) {
+      var zIndex;
+      var screenLimit = 0;
+      var screenTabletLimit = 768;
+      var screenMobileLimit = 480;
+
       var bringUpFront = function(el) {
-        zIndex += 1;
-        el.css('z-index', zIndex);
+        // We get the initial z-index css.
+        var currZIndex = parseInt(el.css('z-index'));
+        if (!zIndex) {
+          zIndex = 2500; // same as in popup.less
+          zIndex++;
+        }
+        if (!currZIndex || zIndex > currZIndex) {
+          el.css('z-index', ++zIndex);
+        }
         $rootScope.$emit('gaPopupFocused', el);
       };
       var updatePosition = function(scope, element, pixel) {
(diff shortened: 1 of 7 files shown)


src/components/popup/PopupDirective.js
```diff
@@ -11,12 +11,22 @@ goog.require('ga_print_service');
...
   module.directive('gaPopup',
-    function($rootScope, $translate, $window, gaBrowserSniffer,
-        gaPrint) {
-      var zIndex = 2000;
+    function($rootScope, $translate, $window, gaBrowserSniffer, gaPrint) {
+      var zIndex;
+      var screenLimit = 0;
+      var screenTabletLimit = 768;
+      var screenMobileLimit = 480;
+
```
gjn: hmm. same values as in css. when changed there, we have to change here as well.

No other way that is more maintainable?

oterral: I know, it's not optimal but I haven't found  another solution.

oterral: :+1:

loicgasser: Let's merge it and change it everywhere in a separate PR. (Popup and so on)

PR #3572, opened 2016-10-26
Add warning in dev/int/ci/infra environments

gjn:
This adds a clearly visible warning below the search box when the application does not run under the official host. This is the sensibilize people not to share a technical link.

[Testklink](https://mf-geoadmin3.int.bgdi.ch/gjn_devint/index.html)


src/index.mako.html
```diff
@@ -154,6 +154,9 @@
   % if device == 'desktop':
         <span ga-help="24,31,25" ga-help-options="{showOnHover:true}"></span>
   % endif
+        <div class="warning-not-prod" ng-class="{'host-not-prod': !globals.hostIsProd}">
```
oterral: Can we use ng-if instead : `ng-if=::!globals.hostIsProd`  ?

src/style/app.less
```diff
@@ -1930,7 +1930,35 @@ body:not(.embed) {
...
 
+  &:hover, &:focus, &:active {
+    z-index: 3000;
+    height: auto;
+    white-space: normal;
+  }
+
+  &.host-not-prod {
+    display: block;
+  };
```
oterral: I think you can optimize the css.

oterral:
I add minor comments otherwise I think it's a good idea.

procrastinatio:
Yes, but too small. Should be in BIG LETTERS accross the map.

And also add something like "Draft" on the print, in case people are printing not yet published data.

src/index.mako.html
gjn: yes, will adapt

src/style/app.less
```diff
@@ -1930,7 +1942,35 @@ body:not(.embed) {
...
+    z-index: 3000;
+    height: auto;
+    white-space: normal;
+  }
 
+  &.host-not-prod {
+    display: block;
+  };
+
+}
```
gjn: css is taken from 3d warning that exists already.

gjn:
I will add a big text in the logo. I thought about text across the map, but I think it's to invasive. We work with this daily and it'll be used by externals for abnahme. Have to find balance between clear signal and still having workable stuff.

For print: in general yes, can be added later on.

oterral:
Just to mention writing something on the official logo is not CD bund compliant I think :).

gjn:
You are right. It's also not an officiel site :)

PR #3573, opened 2016-10-26
Fixes

oterral:
[Test](https://mf-geoadmin3.int.bgdi.ch/teo_fixes/index.html?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&X=166000.00&Y=637000.00&dev3d=true&zoom=1&debug&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege&layers_visibility=false,false,false,false&layers_timestamp=18641231,,,)

Testing:

Draw a line -> the profile and the styles popups should opens.
Deselect the line then re-select it  -> the 2 popups should opens.

Go out the draw mode click on the line -> the shop doesn't appear.

### DIFF
--- a/src/components/draw/DrawDirective.js
+++ b/src/components/draw/DrawDirective.js
@@ -686,7 +686,8 @@ goog.require('ga_styles_service');
                   onClosePopup);
               scope.$applyAsync();
             }
-          } else {
+          }
+          if (!gaMapUtils.isMeasureFeature(scope.feature)) {
             // Move the popup on the closest coordinate of the click event
             var coord = clickCoord ?
                 geometry.getClosestPoint(clickCoord) :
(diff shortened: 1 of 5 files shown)


AFoletti:
Following your instructions I only have 1 popup on the line after deselecting-reselecting it (info tooltip). What should the second be?

Another issue:
- draw a line
- add a point marker
- exit edit mode
- select the line --> all is OK
- select the point --> info on the point is visualized in the popup, but the line length and option to open profile remains

loicgasser:
Good catch @AFoletti . Note that it is already the case on prod.
I think we should go with the present fixes and fix that for next time.

oterral:
bad deploybranch

oterral:
Nice bug found @AFoletti  I cretae an issue

PR #3580, opened 2016-10-27
Remove unsupported platform

loicgasser:
It seems that the combination of IE9 and Windows 7 are not supported by sauclabs anymore.

```Warning: you have no plugins providing access control for builds, so falling back to legacy behavior of permitting any downstream builds to be triggered.```

See also:
https://jenkins.prod.bgdi.ch/job/geoadmin3/727/console

It works now as expected without this platform.

### DIFF
--- a/test/saucelabs/test.py
+++ b/test/saucelabs/test.py
@@ -66,8 +66,6 @@ if __name__ == '__main__':
             'version': "47.0", 'screenResolution': "1280x1024"},
         # Internet Exeplorer
         {'platform': "Windows 7", 'browserName': "internet explorer",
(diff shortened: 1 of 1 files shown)


gjn:
ping @davidoesch . Can we drop IE9 support soon?

davidoesch:
We need This warning since a lot of customers still work with ie9. For re3 we keep support. Re4 tbd

gjn:
Saucelabs doesn't support IE9 anymore. So our auto-tests for IE9 are not available anymore...

davidoesch:
so we have to live with that-- "IE9 support is best effort"

ok?

gjn:
I'm testing it by hand from time to time...

PR #3581, opened 2016-10-27
Add layers_params PL parameter

gjn:
For emapis functionality - but could be used for other use cases as well.

This PR adds a new permalink parameters. We already have [layers_opacity, layers_visibility and layers_timestamp](http://help.geo.admin.ch/?id=54&lang=de) to configure layers specifically. The new `layers_params` parameter allows to add request parameters to any image layer requests.

In the emapis case, it allows to specifcy an equery parameter with `layer_params=equery...`.

It's open if we want to document this. I'm also open to other name suggestions. @loicgasser I don't think ESRI conventions makes sense here, as we already have other params with same pattern.

[Testlink](https://mf-geoadmin3.int.bgdi.ch/gjn_layerparams/index.html?topic=emapis&layers=ch.blw.emapis-hochbau&lang=de&bgLayer=voidLayer&X=166308.27&Y=610841.50&zoom=0&catalogNodes=1916&layers_params=equery%3D(typ_code%3D41))

@davidoesch @oterral Thoughts?

### DIFF
--- a/src/components/UrlUtilsService.js
+++ b/src/components/UrlUtilsService.js
@@ -111,8 +111,12 @@ goog.provide('ga_urlutils_service');
               kv = keyValue.split('=');
               key = this_.tryDecodeURIComponent(kv[0]);
               if (angular.isDefined(key)) {
-                obj[key] = angular.isDefined(kv[1]) ?
-                    this_.tryDecodeURIComponent(kv[1]) : true;
+                if (kv.length > 1) {
+                  obj[key] = this_.tryDecodeURIComponent(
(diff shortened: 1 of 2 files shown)


loicgasser:
+1 for your solution, it's even more generic

davidoesch:
the more paremeters we have to more we have to migrate

my feedback
- +1 for @gjn proposal
- don't document it officially ( os it is not an official function described in the API)

gjn:
ok. Thanks for the feedback. @p1d1d1 on next deploy, this can be used. Let me know if you have questions.

p1d1d1:
Ok for me. thanks

PR #3583, opened 2016-10-28
Add some KML tests and fix 2 more bugs

oterral:
Bugs  fixed are :  
- Some url of maki icons which was not transformed automatically with a `-` in the name like square-stroked-24@2x.png
-  When a KML is parsed and a point is displayed using the default style. The scale was two high.

### DIFF
--- a/src/components/map/KmlService.js
+++ b/src/components/map/KmlService.js
@@ -73,7 +73,7 @@ goog.require('ga_urlutils_service');
         // Replace all old maki urls image by the color service url
         // Test regex here: https://regex101.com/r/rF2tA1/3
         kml = kml.replace(
-          /<href>https?:\/\/[a-z\d\.\-]*(bgdi|geo.admin)\.ch[a-zA-Z\d\-_\/]*img\/maki\/([a-z]*-24@2x\.png)/g,
+          /<href>https?:\/\/[a-z\d\.\-]*(bgdi|geo.admin)\.ch[a-zA-Z\d\-_\/]*img\/maki\/([a-z\-]*-24@2x\.png)/g,
           '<href>' + gaGlobalOptions.apiUrl + '/color/255,0,0/$2'
         );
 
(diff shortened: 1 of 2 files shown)


src/components/map/KmlService.js
```diff
@@ -153,7 +153,7 @@ goog.require('ga_urlutils_service');
           // OL applies a default scale multiplier to 0.5
           // https://github.com/openlayers/ol3/blob/master/src/ol/format/kml.js#L622
           // https://github.com/openlayers/ol3/pull/5745
-          if (image && image.getScale()) {
+          if (image && image.getScale() && image.getScale() != 1) {
```
loicgasser: what if it is 2? Can it be greate than 1?

loicgasser: I think it's better to first:
```
Math.sqrt(image.getScale())
```
if 1 it remains 1
and then multiply by 2?

That's what ol used to do before:
https://github.com/openlayers/ol3/pull/5745

loicgasser: no?

loicgasser: Are you sure...
Here they used to define the setter:

https://github.com/openlayers/ol3/blob/v3.18.0/src/ol/format/kml.js#L1451 (where they took the square root if scale was defined)

Here the read the node (so they use take the square root):
https://github.com/openlayers/ol3/blob/v3.18.0/src/ol/format/kml.js#L611

then

https://github.com/openlayers/ol3/blob/v3.18.0/src/ol/format/kml.js#L214

I don't think they ever multiplied the scale by 2.
Can you find me the reference?

I would really just do:
```
if (image && image.getScale()) {
    image.setScale(Math.sqrt(image.getScale());
}
```

That is if you want to keep the old behaviour. Maybe I don't understand....

oterral: In the last version they don't take the square root and divide by 2 by default. 

so we need to multiply by 2 to have the scale in the KML file and then take the square root like before.

loicgasser: Yes ok, thank you. This scale stuff really makes no sense at the moment but it's the best fix we could do.

PR #3590, opened 2016-11-02
Update geojson source file on lang change

loicgasser:
When changing languages one should download the new geojson URL from the layersConfig file.
This was not the case for mainly 2 reasons.

1. We didn't access the new geojsonUrl in the layersConfig
2. We used to listen to `$translateChangeEnd` which happens before the new layersConfig file was loaded

[Test Link](https://mf-geoadmin3.int.bgdi.ch/gal_geosjon_lang/index.html?topic=ech&lang=fr&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.bafu.hydroweb-messstationen_gefahren)

### DIFF
--- a/src/components/map/RealtimeLayersService.js
+++ b/src/components/map/RealtimeLayersService.js
@@ -10,15 +10,15 @@ goog.require('ga_map_service');
 
   module.provider('gaRealtimeLayersManager', function() {
     this.$get = function($rootScope, $http, $timeout, gaLayerFilters,
-        gaMapUtils, gaUrlUtils) {
+        gaMapUtils, gaUrlUtils, gaLayers) {
 
       var timers = [];
       var realTimeLayersId = [];
       var geojsonFormat = new ol.format.GeoJSON();
 
       function setLayerSource(layer) {
-        var fullUrl = gaUrlUtils.proxifyUrl(layer.geojsonUrl);
         var olSource = layer.getSource();
+        var fullUrl = gaUrlUtils.proxifyUrl(layer.geojsonUrl);
         $http.get(fullUrl).then(function(response) {
           var data = response.data;
           olSource.clear();
@@ -78,10 +78,13 @@ goog.require('ga_map_service');
         });
 
         // Update geojson source on language change
-        $rootScope.$on('$translateChangeEnd', function(evt) {
+        // This event is triggered after the layersConfig is loaded
+        $rootScope.$on('gaLayersTranslationChange', function(evt) {
 
           realTimeLayersId.forEach(function(bodId, i) {
             var olLayer = gaMapUtils.getMapLayerForBodId(map, bodId);
+            olLayer.geojsonUrl =
+                gaLayers.getLayerProperty(olLayer.bodId, 'geojsonUrl');
             $timeout.cancel(timers[i]);
             setLayerSource(olLayer);
           });
--- a/test/specs/map/RealtimeLayersService.spec.js
+++ b/test/specs/map/RealtimeLayersService.spec.js
@@ -74,7 +74,10 @@ describe('ga_realtimelayers_service', function() {
 
       module(function($provide) {
         $provide.value('gaLayers', {
-          loadConfig: function() {}
+          loadConfig: function() {},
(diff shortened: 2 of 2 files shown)


loicgasser:
jenkins test

loicgasser:
some selenium failures on the search totally unrelated.

test/specs/map/RealtimeLayersService.spec.js
```diff
@@ -195,8 +198,8 @@ describe('ga_realtimelayers_service', function() {
       var layer = map.getLayers().item(0);
       expect(layer.getSource().getFeatures().length).to.be(1);
 
-      // Here we don't test the change of the url because it's done in another
-      // service, we only test if the data are requetsed on language change.
+      var newLang = 'custom2';
+      dataUrl = dataUrl.replace('_custom', '_' + newLang);
```
oterral: The meaning of the comment was : 'The RealtimeLayersService is not responsible of changing the url' so here you test something unrelated to the RealtimeLayersService . 

On language change the  RealtimeLayersService  reload  the geojsonUrl property it doesn't care of what's in it so the test was good no need to complexify more.

src/components/map/RealtimeLayersService.js
```diff
@@ -10,14 +10,16 @@ goog.require('ga_map_service');
...
       var timers = [];
       var realTimeLayersId = [];
       var geojsonFormat = new ol.format.GeoJSON();
 
-      function setLayerSource(layer) {
-        var fullUrl = gaUrlUtils.proxifyUrl(layer.geojsonUrl);
+      function setLayerSource(layer, isNewSource) {
+        var fullUrl = isNewSource ?
+            gaLayers.getLayerProperty(layer.bodId, 'geojsonUrl') :
+            gaUrlUtils.proxifyUrl(layer.geojsonUrl);
```
oterral: To be more simple, we have 2 solutions:

-  Add a getter to the geojsonUrl property in gaDefinePropertiesForLayer

or

-  Remove the  `geojsonUrl` property from gaDefinePropertiesForLayer and access to the property through `gaLayers.getLayerProperty` everywhere.

With one of these solution no need of `isNewSource` parameter.

src/components/map/RealtimeLayersService.js
```diff
@@ -10,14 +10,16 @@ goog.require('ga_map_service');
...
 
       var timers = [];
       var realTimeLayersId = [];
       var geojsonFormat = new ol.format.GeoJSON();
 
-      function setLayerSource(layer) {
-        var fullUrl = gaUrlUtils.proxifyUrl(layer.geojsonUrl);
+      function setLayerSource(layer, isNewSource) {
+        var fullUrl = isNewSource ?
+            gaLayers.getLayerProperty(layer.bodId, 'geojsonUrl') :
```
oterral: `proxifyUrl`  was used because there was no CORS headers so probably you still  need it, unless somthing change server side.

oterral:
Good fix some minor comments.

src/components/map/RealtimeLayersService.js
loicgasser: https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/map/MapService.js#L253
We could edit this property instead

src/components/map/RealtimeLayersService.js
loicgasser: Let's leave it as it is not proxified anyway if we're using `admin` URL I am not sure this will be and is always the case. It doesn't hurt.

src/components/map/RealtimeLayersService.js
loicgasser: yes got it now

oterral: Consensus found doing like in LayerManager : https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/layermanager/LayermanagerDirective.js#L338

Not optimal but does the job.

loicgasser:
updated

loicgasser:
I'll put it in LayerManager.

oterral:
No let it there it's the good place.

PR #3611, opened 2016-11-15
Add tests for gaMeasure service

oterral:
No functionnal changes.

Review without spaces : https://github.com/geoadmin/mf-geoadmin3/pull/3611/files?w=0

### DIFF
--- a/src/components/measure/MeasureService.js
+++ b/src/components/measure/MeasureService.js
@@ -107,84 +107,85 @@ goog.require('ga_measure_filter');
         this.updateOverlays = function(layer, feature) {
           var overlays = feature.get('overlays');
 
-          if (overlays) {
-            var currIdx = 0, geomLine;
-            var geom = geomLine = feature.getGeometry();
+          if (!overlays) {
+            return;
+          }
+          var currIdx = 0, geomLine;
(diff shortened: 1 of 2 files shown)


pfanguin:
It's ok for me.

src/components/measure/MeasureService.js
```diff
@@ -239,48 +240,52 @@ goog.require('ga_measure_filter');
...
+            return;
+          }
           map.getLayers().on('remove', function(evt) {
             if (evt.element === layer) {
-              var features = evt.element.getSource().getFeatures();
-              for (var i = 0; i < features.length; i++) {
-                this.removeOverlays(features[i]);
-                features[i].set('overlays', undefined);
+              var i, features = evt.element.getSource().getFeatures();
+              while (i = features.pop()) {
```
loicgasser: last feature poped is undefined so you actually do `evt.element.getSource().removeFeature(undefined)`
I guess this is ok on ol side, but maybe `if (i){}` would be safer.

oterral: Basically it's what this loop does. If `i` is `undefined ` it doesn't enter in the loop.

PR #3592, opened 2016-11-03
[WIP] Using topic groups and horizontal scrolling bar

loicgasser:
Fix for https://github.com/geoadmin/mf-geoadmin3/issues/3157

Not being able to use flex because of IE9 is not fun. @oterral knows what I am talking about.

Here is a [Test Link](https://mf-geoadmin3.int.bgdi.ch/gal_topic_groups/index.html?api_url=%2F%2Fmf-chsdi3.dev.bgdi.ch%2Fgal_topic_groups&lang=en)

This PR introduces the concepts of topics group, e.g. groups appear on the same line.
It is generic enough that we can create as many topic group as we want without changing the code.

This PR works with
https://github.com/geoadmin/mf-chsdi3/pull/2311

Before merging:

1. Add all missing single topics
2. Add all images and translations
3. Update `bod_master` with migration script
4.  Remove https://github.com/geoadmin/mf-chsdi3/commit/9db0dd58da8bca36ddfbb0ba436a9f540c25af35
5. Merge https://github.com/geoadmin/mf-chsdi3/pull/2311
6. Merge this PR

Issue #3157, opened 2016-02-19
Topic and single Topic


davidoesch:
Situation
- topic: current topic selector is growing - on desktop a scrolling is now needed
- Single topics as shown in www.geo.admin.ch on footer based https://docs.google.com/spreadsheets/d/11zjWdLgaCZWiFhB8kP2AWhOb9vKBqLuXeXDPTSd4rDk/edit?pli=1#gid=0  should be intgerated as well

Task:
-merge single topics  and topics to have them in map.geo.admin.ch and www.geo.admin.ch

Action: (tbd)

map.geo.admin.ch
- topics selector
- [ ] make three lines, scrollable horizontally
- [ ] first line BLUE topics and the end the RED ones
- [ ] Second line GREEN topics
- [ ] Third line Single topic maps as in https://docs.google.com/spreadsheets/d/11zjWdLgaCZWiFhB8kP2AWhOb9vKBqLuXeXDPTSd4rDk/edit?pli=1#gid=0 

www.geo.admin.ch
add the GREEN, RED BLUE as well in a divider (see example fanguin)

Result: tipic selection in map and www are synced

@gjn @cedricmoullet  feedback?

gjn:
+1.

cedricmoullet:
I like the idea to synch both.
But I don't get how the single topics will be integrated in map.geo.admin.ch
(from design perspective). Do you have a mockup for desktop and mobile ?


2016-02-19 8:51 GMT+01:00 Gilbert Jeiziner <notifications@github.com>:

> +1.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/geoadmin/mf-geoadmin3/issues/3157#issuecomment-186107206>
> .
>



-- 
Twitter: http://twitter.com/cedricmoullet
Linked In: http://www.linkedin.com/in/cedricmoullet

<http://twitter.com/cedricmoullet>

davidoesch:
they are just images with a permalink as they do currently exits at the bottom of www.geo.admin.ch.

cedricmoullet:
yes, ok. But how do you integrate this in the UI of map.geo.admin.ch ?

2016-02-19 11:05 GMT+01:00 David Oesch <notifications@github.com>:

> they are just images with a permalink as they do currently exits at the
> bottom of www.geo.admin.ch.
>
> —
> Reply to this email directly or view it on GitHub
[...]

PR #3612, opened 2016-11-15
Change topic selector

loicgasser:
[Test Link](https://mf-geoadmin3.int.bgdi.ch/gal_topic_selector/index.html)

Relates to discussions in:
https://github.com/geoadmin/mf-geoadmin3/issues/3157
https://github.com/geoadmin/mf-geoadmin3/pull/3592

It would be good to test on all platform before merging this.


davidoesch:
@mariokeusen can you test on dekstop plattforms? I'll test mobile

feedback so far

- [ ] "thema wechseln" as button is ugly, I would rather stick to text and icon as it is now
- [ ] scroll the modal window rather than in the app

image

loicgasser:
This is not the latest version something went wrong with the deploybranch sorry

loicgasser:
image

Ok so leave button as it is now in geoadmin.
Tune modal max max-height a bit.

src/index.mako.html
```diff
@@ -420,17 +420,9 @@
...
-            <a class="dropdown-toggle" data-toggle="dropdown" href="#">
-              <span translate>choose_theme</span>
-              <i class="fa fa-chevron-sign-right"></i>
-            </a>
-            <div class="dropdown-menu">
-              <div ga-topic ga-topic-ui="list"></div>
-            </div>
-          </div>
-  % endif
+          <button type="button" class="btn btn-default pull-right topic-btn" data-toggle="modal" data-target=".ga-topic-modal">
```
oterral: can you use an identifier for the data-target like this `data-target="#topicSelector .modal"` ?   A class could potentially be not unique.

src/components/topic/partials/topic.html
```diff
@@ -1,8 +1,22 @@
-<div class="ga-topic-item"
-     ng-repeat="topic in ::topics"
-     ng-click="setActiveTopic(topic)"
-     ng-class="{'ga-topic-active': topic == activeTopic}"
-     data-original-title="{{topic.tooltip | translate}}">
-  <div translate>{{::topic.id}}</div>
-  <div class="ga-topics-sprite ga-topics-sprite-{{::topic.id}}"></div>
+<div class="modal fade ga-topic-modal">
```
oterral: no need to use a `ga-topic-modal` class .

src/components/topic/TopicDirective.js
```diff
@@ -11,12 +11,11 @@ goog.require('ga_topic_service');
...
           restrict: 'A',
-          templateUrl: function(element, attrs) {
-            return 'components/topic/partials/topic.' +
-              ((attrs.gaTopicUi == 'select') ? 'select.html' : 'html');
-          },
+          templateUrl: 'components/topic/partials/topic.html',
           scope: {},
           link: function(scope, element, attrs) {
+            var modalClsName = 'ga-topic-modal';
+            var modal = element.find('div.' + modalClsName);
```
oterral: Use the `.modal` class instead`.   More generic and we don't need a constant to get the correct element.

src/components/topic/TopicDirective.js
```diff
@@ -37,6 +37,7 @@ goog.require('ga_topic_service');
               scope.activeTopic = gaTopic.get();
               scope.$applyAsync(function() {
                 element.find('.ga-topic-item').tooltip({
+                  container: 'div.' + modalClsName,
```
oterral: Maybe we could remove all these toolitps and use the tilte attribute ? what do you think?

oterral:
I've tested a bit and it seems on tablet there is css adjusments to make.

davidoesch:
On mobile, modal can be have a larger height
screenshot_20161115-160703

loicgasser:
Updated css were stuck in varnish cache.
Please test review again

mariokeusen:
Made a few tests on desktop (FF, IE, Chrome): it works and I like it!

src/components/topic/partials/topic.html
```diff
@@ -1,8 +1,22 @@
-<div class="ga-topic-item"
-     ng-repeat="topic in ::topics"
-     ng-click="setActiveTopic(topic)"
-     ng-class="{'ga-topic-active': topic == activeTopic}"
-     data-original-title="{{topic.tooltip | translate}}">
-  <div translate>{{::topic.id}}</div>
-  <div class="ga-topics-sprite ga-topics-sprite-{{::topic.id}}"></div>
+<div id="topicSelector" class="modal fade">
```
oterral: no id in hte partial . I meant putting it in the index.mako.html
```
<div id="topicSelector" ga-topic>
```

src/components/topic/TopicDirective.js
```diff
@@ -37,6 +37,7 @@ goog.require('ga_topic_service');
               scope.activeTopic = gaTopic.get();
               scope.$applyAsync(function() {
                 element.find('.ga-topic-item').tooltip({
+                  container: modalId,
```
oterral: can't we avoid the use of container?

davidoesch:
Nice. tested on different Brwosers and devices. especially on mobile tablets it way more useful than the wheel

one thing to fix: no button only text (as the delete and share button in draw) and remove the icon (we don't use icons in the UI--KISS)

image

loicgasser:
KISS: a button is simple, a hidden anchor is not

Let me try to convice you a bit here. **I can ensure you nobody will see it**, even less than with the current design. Maybe you don't like the design in this PR, I guess it's a matter of taste, but we should really try to make an effort to show this feature to our users. Make sure people understand they can click it and change topic. We alwas say that people don't change topic, maybe there is a reason. During usability tests people never realized they were able to change the topic. Don't you have another idea?

davidoesch:
A button won't solve the issue that ppl don't find the topics. this has to be solved with more than a button
it is about consistency in the UI: we simply don't have buttons in the accordion top level. We don't use icons in the UI of map.geo.admin.ch. WHY? because we introduced removed and reintroduced and removed icons  and the background is quite well described here http://mapbrief.com/2013/02/19/an-iconography-of-confusion-why-map-portals-dont-work-part-iv/

we will change the access to  topics, completely 
there appraoches like this
image
or even in an ideal world will have SEO whcih will guide users directly to topics or have topics in the search as well... so with UI now we just do patchwork

Conclusio: we stick to our CI / CD . and when we fix it, we fix it completely. we have a link currently and use now a link as well. this  issue was about fixing the height of the topic selector. we fix that first and then we move/improve topics.

loicgasser:
ok no problem

loicgasser:
You notice the use of the same icon though. ;)

src/components/topic/TopicDirective.js
loicgasser: If not defined tooltip on topics pushes the modal content aside.

loicgasser:
comments addressed

oterral:
@davidoesch are you happy with the design

davidoesch:
+1 clear go

PR #3637, opened 2016-11-30
[WIP] Remove view's extent

oterral:
Some layer have an extent bigger than the default one.

This PR will work when chsdi will return an extent property in the layersConfig

[Test](https://mf-geoadmin3.int.bgdi.ch/teo_fixes/index.html?topic=ech&lang=fr&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.swiss-map-vector1000.metadata&X=171889.81&Y=622188.44&zoom=0)


pfanguin:
LGTM

PR #3652, opened 2016-12-09
Clip digital products by rectangle

loicgasser:


loicgasser:
We checked with @pauloamado and everthing works as expected. I am merging that to prepare Monday's shop demo.

PR #3660, opened 2016-12-15
Fix keep version

loicgasser:
A nasty one...

```bash
ltgal@ip-10-220-4-246 (custom):~/mf-geoadmin3 **(gal_keep_version)**
$ export KEEP_VERSION='false'
ltgal@ip-10-220-4-246 (custom):~/mf-geoadmin3 **(gal_keep_version)**
$ make .build-artefacts/last-version
mkdir -p .build-artefacts/
test "1612151424" != "1612151423" && echo 1612151424 > .build-artefacts/last-version || :
ltgal@ip-10-220-4-246 (custom):~/mf-geoadmin3 **(gal_keep_version)**
$ export KEEP_VERSION='true'
ltgal@ip-10-220-4-246 (custom):~/mf-geoadmin3 **(gal_keep_version)**
$ make .build-artefacts/last-version
mkdir -p .build-artefacts/
test "1612151424" != "1612151424" && echo 1612151424 > .build-artefacts/last-version || :
```

### DIFF
--- a/scripts/deploysnapshot.sh
+++ b/scripts/deploysnapshot.sh
@@ -14,14 +14,14 @@ fi
 
 pwd=$(pwd)
 DEPLOY_TARGET=$2
-NEW_VERSION=$3
 RC_FILE=rc_$DEPLOY_TARGET
 SNAPSHOTDIR=/var/www/vhosts/mf-geoadmin3/private/snapshots/$1
 
 # Prepare snapshot
 cd $SNAPSHOTDIR/geoadmin/code/geoadmin
-if [ "$NEW_VERSION" = "true" ]; then
+if [ "$KEEP_VERSION" = "false" ]; then
   make .build-artefacts/last-version
+  export KEEP_VERSION='true'
 fi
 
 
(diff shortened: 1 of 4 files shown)


scripts/deploysnapshot.sh
```diff
@@ -21,6 +21,7 @@ SNAPSHOTDIR=/var/www/vhosts/mf-geoadmin3/private/snapshots/$1
 # Prepare snapshot
 cd $SNAPSHOTDIR/geoadmin/code/geoadmin
 if [ "$NEW_VERSION" = "true" ]; then
+  export KEEP_VERSION='false'
```
procrastinatio: or `unset NEW_VERSION` ?

loicgasser: No per default KEEP_VERSION is 'true' in Makefile for rc_(dev|int|prod) it will be re-applied later when sourcing those files.

oterral: I'm not sure to understand what is the difference between NEW_VERSION and KEEP_VERSION, why don't we put KEEP_VERSION =false in rc_(dev|int|prod) directly and remove NEW_VERSION ?

loicgasser: We actually always want keep the version when deploying the first time so `KEEP_VERSION='true'`. Optionally, to avoid overriding what we have in production we want to generate a new version from a snapshot for instance.
So behaviour is correct. Now we could remove the NEW_VERSION parameter that is correct. And simply do
`make s3deployint KEEP_VERSION='false'.

loicgasser:
adapted according to comments.

PR #3661, opened 2016-12-15
More robust variable def

loicgasser:
When deploying if you source another file before hand, you're in trouble.
With this PR we are explicit.


procrastinatio:
Yes, some varaibles are defined in source files, while other default are in the `Makefile`. We may also `unset` all variables we use ?

loicgasser:
All are defined in source files. We can't use unset or we will loose the options we use to parameter our scripts. We could remove the ones in the makefile now. But better safe than sorry.

oterral:
To be consstent with orthers urls I've added the `.build-artefacts/last-print-url` target.

oterral:
THx for this @loicgasser

PR #3664, opened 2017-01-03
Change caching header and only use no-cache

gjn:
I was reading up on http caching over Christmas :=). This should reduce the transferred data of our non cached resources, mainly the html pages of map.geo.admin.ch.

Before, we used the no-store header, which forced the browsers to alwas download the complete html content from the server. With only the no-cache header, we only download the html when either last-modified date changed or the etag changed. Otherwise, we simply respond with a 304. The difference is 8KB less data transfered per request. Also, page loads faster. Note that 304 still registers as a hit/visit, so it won't impact statistics.

Note that the client still checks the server (in our case S3) for new content on every new attempt - so it should never use old pages.

[Testlink](https://mf-geoadmin3.int.bgdi.ch/gjn_caching/index.html)

We save around 4GB of daily data transfer and we safe 20$ per month.


oterral:
It's good for me. Let's try in real life.:)

PR #3665, opened 2017-01-03
Fix #3576: Use the wms urls define in the layer's source

### LINKED ISSUE #3576: Remove the use of config.wmsUrl in PrintDirective
We should use `gaGlobalOptions.wmsUrl` instead or set the url property on the WMS bod layers.

https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/print/PrintDirective.js#L255

oterral:
Fix #3576 

In fact,  the use  of the `globalOptions.wmsUrl` property was not relevant. It makes more sense to use the real url defined in the source object.

### DIFF
--- a/src/components/print/PrintDirective.js
+++ b/src/components/print/PrintDirective.js
@@ -243,12 +243,14 @@ goog.require('ga_time_service');
         },
         'WMS': function(layer, config) {
           var enc = $scope.encoders.layers['Layer'].call(this, layer);
-          var params = layer.getSource().getParams();
+          var source = layer.getSource();
+          var params = source.getParams();
           var layers = params.LAYERS.split(',') || [];
           var styles = (params.STYLES !== undefined) ?
               params.STYLES.split(',') :
               new Array(layers.length).join(',').split(',');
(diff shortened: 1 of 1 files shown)


gjn:
We can still overload it with wms_url parameter?

oterral:
The same protocol error as before.

oterral:
I fix

oterral:
Fixed [see](https://mf-geoadmin3.int.bgdi.ch/teo_fixes/index.html?lang=fr&topic=luftbilder&wms_url=%2F%2Fwms.geo.admin.ch&bgLayer=ch.swisstopo.pixelkarte-grau&X=190000.00&Y=660000.00&dev3d=true&zoom=1&debug&layers=ch.swisstopo.lubis-luftbilder_schwarzweiss,ch.swisstopo.lubis-luftbilder_farbe,ch.swisstopo.lubis-bildstreifen&layers_timestamp=99991231,99991231,&catalogNodes=1179,1180,1184,1186&layers_visibility=false,false,true)

gjn:
Thanks alot!

PR #3666, opened 2017-01-04
Show more precision on tooltip popup markers

gjn:
When clicking on a marker (either one places via pl parameter, search or kml), we only see meters. This PR will show millimeter precision in the tooltips for markers.

People complained about it, because when you enter a coordinate in the search with mm precision, the accosiated tooltip does not contain mm...and sharing is harder that way.

[Testlink](https://mf-geoadmin3.int.bgdi.ch/gjn_prec/index.html)

### DIFF
--- a/src/components/measure/MeasureService.js
+++ b/src/components/measure/MeasureService.js
@@ -13,8 +13,8 @@ goog.require('ga_measure_filter');
       var Measure = function() {
 
         // Transform 2111333 in 2'111'333
-        this.formatCoordinates = function(coordinates) {
-          var raw = ol.coordinate.toStringXY(coordinates, 0);
+        this.formatCoordinates = function(coordinates, prec) {
+          var raw = ol.coordinate.toStringXY(coordinates, prec || 0);
           if (coordinates && coordinates.length === 3) {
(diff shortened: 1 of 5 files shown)


src/components/measure/MeasureService.js
```diff
@@ -13,8 +13,9 @@ goog.require('ga_measure_filter');
       var Measure = function() {
 
         // Transform 2111333 in 2'111'333
-        this.formatCoordinates = function(coordinates) {
-          var raw = ol.coordinate.toStringXY(coordinates, 0);
+        this.formatCoordinates = function(coordinates, prec) {
+          var raw = ol.coordinate.toStringXY(coordinates,
+                                             prec == undefined ? 0 : prec);
```
oterral: `prec || 0 `

oterral:
can you add a test please.

oterral:
at least for MeasureService , the other directives haven't a test class.

src/components/measure/MeasureService.js
gjn: man, I don't code enough anymore.

gjn:
Adapted and added tests. Thanks alot for the review.

oterral:
lgtm thx

Issue #3600, opened 2016-11-09
drawing tool: replace EDIT and SHARE link with shortened URLS


davidoesch:
edit and share link are long and not URLencoded. Users save them on their system. I suspect that saving in word etc introduces CR's and empty spaces---which then result in faulty links

proposal: create  shortened URLS uopn opening SHARE in drawing

oterral:
Yes we can

PR #3668, opened 2017-01-04
Teo fixes

oterral:
Fix https://github.com/geoadmin/mf-geoadmin3/issues/3600

Use short url for draw shar links

[Test](https://mf-geoadmin3.int.bgdi.ch/teo_fixes/index.html?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.swisstlm3d-wanderwege&layers_visibility=false,true&X=202228.40&Y=645507.67&zoom=1&adminId=YoTW95iMQfuKdVjui2i_SA)

### DIFF
--- a/src/components/NetworkStatusService.js
+++ b/src/components/NetworkStatusService.js
@@ -1,11 +1,9 @@
 goog.provide('ga_networkstatus_service');
 
-goog.require('ga_browsersniffer_service');
 goog.require('ga_waitcursor_service');
 (function() {
 
   var module = angular.module('ga_networkstatus_service', [
-    'ga_browsersniffer_service',
     'ga_waitcursor_service'
   ]);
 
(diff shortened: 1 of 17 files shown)


src/components/NetworkStatusService.js
```diff
@@ -60,13 +58,8 @@ goog.require('ga_waitcursor_service');
...
     this.$get = function($document, $rootScope, $timeout, $window,
-        gaBrowserSniffer, gaGlobalOptions) {
-      if (!gaBrowserSniffer.mobile) {
-        return {
-          offline: false,
-          check: function() {}
-        };
-      }
+        gaGlobalOptions) {
+
```
gjn: why was this removed?

oterral: There was a circular dependency injection with gaUrlUtils using $http and since the iPad, for example, uses the desktop version I think it makes sense to remove this conditionnal use of the geNetworkService.

oterral: Not $http sorry. with gaBrowserSniffer.

gjn: What are the consequences? Do desktop apps now poll the checker all the time?

Or only when offline is detected first?

oterral: On app launch: 
    If offline  -> launch checker requests every 2sec until it gets online again 
    If online  -> If an http request failed -> launch checker to see if really offline -> If offline -> launch checker every 2 sec ....

gjn: Fine with me then.

Please squash a little, then merge.

PR #3669, opened 2017-01-05
Fix #3663: Doesn't delete the kml file when we push the delete button

### LINKED ISSUE #3663: KML Edit links disappear
Reproduced on all system: both screens are emtpy

"Bei unten stehendem Link verschwanden jeweils alle Eintragungen wenn man etwas neues zeichnen wollte. Bei erneutem aufrufen des Links sind die Eintragungen jedoch wieder erschienen. Dies ist mir bereits letzte Woche mit diesem Link passiert. Nun wollte ich erneut etwas zeichnen und wieder verschwand der Inhalt, auch wenn ich den Link neu aufrufe, erscheint die Strecke nicht mehr. "

Link zum teilen:
https://map.geo.admin.ch/?topic=ech&lang=de&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege,KML%7C%7Chttps:%2F%2Fpublic.geo.admin.ch%2FRgl5sNpaRlyofR-RWjkJFg&layers_visibility=false,false,false,false,true&layers_timestamp=18641231,,,,&X=245231.10&Y=699439.64&zoom=6

Link zum editieren:
https://map.geo.admin.ch/?topic=ech&lang=de&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege&layers_visibility=false,false,false,false,true&layers_timestamp=18641231,,,,&X=245231.10&Y=699439.64&zoom=6&adminId=nBYFnFAbTeSKmg8FSrvTtg

both links are empty

image

other customer with similar issues:

Customer feedback ID : 2017010807353602:

oterral:
[Test](https://mf-geoadmin3.dev.bgdi.ch/ltteo/index.html?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege&layers_visibility=false,false,false,false&layers_timestamp=18641231,,,&X=199964.47&Y=650035.53&zoom=1)

### DIFF
--- a/src/components/draw/DrawDirective.js
+++ b/src/components/draw/DrawDirective.js
@@ -335,7 +335,7 @@ goog.require('ga_styles_service');
             unWatch.pop()();
           }
 
-          wasNotOnMap = (map.getLayers().getArray().indexOf(layer) == -1);
+          wasNotOnMap = map.getLayers().getArray().indexOf(layer) == -1;
           // Re-add the layer if it's not already on map.
           if (wasNotOnMap) {
             map.addLayer(layer);
--- a/src/components/draw/DrawDirective.js
+++ b/src/components/draw/DrawDirective.js
@@ -274,16 +274,18 @@ goog.require('ga_styles_service');
 
           // Set the layer to modify if exist, otherwise we use an empty
           // layer
-          layer = createDefaultLayer(map, useTemporaryLayer);
+          layer = layer || createDefaultLayer(map, useTemporaryLayer);
           if (!useTemporaryLayer) {
 
-            // If there is a layer loaded from public.admin.ch, we use it for
-            // modification.
-            map.getLayers().forEach(function(item) {
-              if (gaMapUtils.isStoredKmlLayer(item)) {
-                layer = item;
-              }
-            });
+            if (!layer.adminId) {
+              // If there is a layer loaded from public.admin.ch, we use it for
+              // modification.
+              map.getLayers().forEach(function(item) {
+                if (gaMapUtils.isStoredKmlLayer(item)) {
+                  layer = item;
+                }
+              });
+            }
 
             // Attach events to the source to update the KML file stored
             // and to remove overlays when necessary
@@ -315,6 +317,9 @@ goog.require('ga_styles_service');
           }
         };
 
+        // This variable keeps track of the existence or not of the layer before
+        // activation of the draw tool (kml loaded from permalink)
+        var wasNotOnMap;
         // Activate the component: active a tool if one was active when draw
         // has been deactivated.
         var activate = function() {
@@ -329,6 +334,13 @@ goog.require('ga_styles_service');
           while (unWatch.length) {
             unWatch.pop()();
           }
+
+          wasNotOnMap = (map.getLayers().getArray().indexOf(layer) == -1);
+          // Re-add the layer if it's not already on map.
+          if (wasNotOnMap) {
+            map.addLayer(layer);
+          }
+
           // if a layer is added from other component (Import KML, Permalink,
           // DnD ...) and the currentlayer has no features, we define a
           // new layer.
@@ -379,11 +391,11 @@ goog.require('ga_styles_service');
             ol.Observable.unByKey(unSourceEvents.pop());
           }
 
-          // Remove the layer if no features added
-          if (layer && (useTemporaryLayer ||
-              layer.getSource().getFeatures().length == 0)) {
+          // Remove the layer if no features added and if the layer was not on
+          // map before activation of the draw tool.
+          if (wasNotOnMap && (layer && (useTemporaryLayer ||
+              layer.getSource().getFeatures().length == 0))) {
             map.removeLayer(layer);
-            layer = null;
           }
           map.removeInteraction(select);
           map.removeInteraction(modify);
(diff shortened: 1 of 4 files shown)


davidoesch:
Well this won't cover the use case if a user wants to delete the KML file so it will not be accessible.
using the testlink above, if I created now a drawing, sharing it and then deleting the drawing completely ...the shared link still contains the drawing. eg here the kml
https://public.int.bgdi.ch/iHCEi1hKRVCsPX9cFX1crQ 
So : this fix will actually only disable the remove button.
-> or do I understand it wrong? 

for me a no go

gjn:
we can't have 2 different behaviours on the same button. So either we add a button or decide what we want.

davidoesch:
I don't understand the function of the button: it says: delete. As a user I expect it is deleted, also from the permalink, so no one can access the data (this was the reason we decided to have an admin and a view link, this saves us the hassel of running a a complete user mmgt..) With this pr a "delete" does not delete the kml content. So collaborating users will work with wrong information...
If a user hits the delete button he expects the drawing itself and the kml (data) gone... if he hits it accidentally , we already do show a warning. 
 

I'm still not sure if I understand now the behaviour / intention of this pr.

oterral:
arf yes you're right I need to save the empty file.

oterral:
@davidoesch it's fixed now the kml file is never delete and an empty kml is saved instead

gjn:
Hmm. Testlink doesn't seem to work for me. Share link doesn't show anything...hmm.

davidoesch:
Same here...

oterral:
Works for me. Disable cache ?

procrastinatio:
To me, looks like the `adminId` given throught the _shortlink_ is not the correct one.

src/components/draw/DrawDirective.js
```diff
@@ -274,16 +274,18 @@ goog.require('ga_styles_service');
...
           if (!useTemporaryLayer) {
 
-            // If there is a layer loaded from public.admin.ch, we use it for
-            // modification.
-            map.getLayers().forEach(function(item) {
-              if (gaMapUtils.isStoredKmlLayer(item)) {
-                layer = item;
-              }
-            });
+            if (!layer.adminId) {
```
oterral: In case another public kml file is added after the layer property  is already set, we don't want this new layer  replace the first one.

oterral:
int is dead. @davidoesch cna you test on my home -> [here](https://mf-geoadmin3.dev.bgdi.ch/ltteo/index.html?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege&layers_visibility=false,false,false,false&layers_timestamp=18641231,,,&X=199964.47&Y=650035.53&zoom=1)

davidoesch:
tested according to line 24-27 in https://docs.google.com/spreadsheets/d/14n3JYNL-4ACiqExUeoqqbHmWmqtfq2BB-qC21Beb6Dw/edit#gid=0 no different behaviour as on prod. win10 chtome

src/components/draw/DrawDirective.js
```diff
@@ -329,6 +334,13 @@ goog.require('ga_styles_service');
           while (unWatch.length) {
             unWatch.pop()();
           }
+
+          wasNotOnMap = (map.getLayers().getArray().indexOf(layer) == -1);
```
loicgasser: () no need

src/components/draw/DrawDirective.js
```diff
@@ -716,24 +728,12 @@ goog.require('ga_styles_service');
...
-            map.removeLayer(layer);
-            return;
-          }
           scope.statusMsgId = 'draw_file_saving';
           var kmlString = gaExportKml.create(layer,
               map.getView().getProjection());
           var id = layer.adminId ||
               gaFileStorage.getFileIdFromFileUrl(layer.url);
-          gaFileStorage.save(id, kmlString,
+          gaFileStorage.save(id, kmlString || '<kml></kml>',
```
loicgasser: why simply not saving when kml string is empty?

src/components/draw/DrawDirective.js
```diff
@@ -315,6 +317,9 @@ goog.require('ga_styles_service');
           }
         };
 
+        // This variable keeps track of the existence or not of the layer before
+        // activation of the draw tool (kml loaded from permalink)
+        var wasNotOnMap;
```
loicgasser: loadedViaPermalink ? would make more sense to me (of course you need to reverse the if conditions). this is a bit cryptic.

src/components/draw/DrawDirective.js
loicgasser: > In case another public kml file is added after the current draw layer

How is this possible in mapgeoadmin?

src/components/draw/DrawDirective.js
oterral: if kmlString is empty the service save nothing if I remember correctly. And we must have KML valid file to get it back.

loicgasser: ok in general if kml is empty I feel like we shouldn't save it, and not add it to the permalink until it has at least one feature and has been saved successuflly.

oterral: No because people wants to keep the same permalink and changing prmalink when it's empty the user could potentially save a link without his KML file

src/components/draw/DrawDirective.js
oterral: loaded via permalink is an example it could be another component. I think the comment is enough to understand what I meant creating this variable.

src/components/draw/DrawDirective.js
oterral: yes

PR #3682, opened 2017-01-19
Add reframe service

loicgasser:
[test link](https://mf-geoadmin3.int.bgdi.ch/gal_reframe/index.html)

When looking for a coordinate in LV95 we then transform it to LV03 using proj4js.
Neverthless, this transformation is rather inaccurate and we had costumers complain about it via webgis feedback.

This PR introduces a reframe service for transformations from

- lv03 to lv95
- lv95 to lv03

using geodesy service.

If the service doesn't work for any reason, we simply fallback to the transformation using proj4js.

Before (no reframe via search):

no_reframe

After (with reframe via search):

with_reframe

### DIFF
--- a/src/components/search/SearchService.js
+++ b/src/components/search/SearchService.js
@@ -1,7 +1,10 @@
 goog.provide('ga_search_service');
+
+goog.require('ga_reframe_service');
 (function() {
 
   var module = angular.module('ga_search_service', [
+    'ga_reframe_service'
   ]);
 
   var DMSDegree = '[0-9]{1,2}[°|º]\\s*';
@@ -28,12 +31,18 @@ goog.provide('ga_search_service');
   // It's a limitiation of proj4 and a sensible default (precision is 10km)
   var MGRSMinimalPrecision = 7;
 
+  var roundCoordinates = function(coords) {
+    return [Math.round(coords[0] * 1000) / 1000,
+      Math.round(coords[1] * 1000) / 1000];
+  };
+
   module.provider('gaSearchGetCoordinate', function() {
-    this.$get = function($window) {
+    this.$get = function($window, $q, gaReframe) {
+
       return function(extent, query) {
         var position;
-        var valid = false;
 
+        // Parse MGRS notation
         var matchMGRS = query.match(regexMGRS);
         if (matchMGRS && matchMGRS.length == 1) {
           var mgrsStr = matchMGRS[0].split(' ').join('');
@@ -41,11 +50,12 @@ goog.provide('ga_search_service');
             var wgs84 = $window.proj4.mgrs.toPoint(matchMGRS[0]);
             position = ol.proj.transform(wgs84, 'EPSG:4326', 'EPSG:21781');
             if (ol.extent.containsCoordinate(extent, position)) {
-              valid = true;
+              return $q.when(roundCoordinates(position));
             }
           }
         }
 
+        // Parse degree EPSG:4326 notation
         var matchDMSN = query.match(regexpDMSN);
         var matchDMSE = query.match(regexpDMSE);
         if (matchDMSN && matchDMSN.length == 1 &&
@@ -81,12 +91,12 @@ goog.provide('ga_search_service');
                 'EPSG:4326', 'EPSG:21781');
           if (ol.extent.containsCoordinate(
             extent, position)) {
-              valid = true;
+            return $q.when(roundCoordinates(position));
           }
         }
 
         var match = query.match(regexpCoordinate);
-        if (match && !valid) {
+        if (match) {
           var left = parseFloat(match[1].replace(/\'/g, ''));
           var right = parseFloat(match[2].replace(/\'/g, ''));
           //Old school entries like '600 000 200 000'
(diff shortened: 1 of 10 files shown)


src/components/search/SearchService.js
```diff
@@ -100,31 +110,37 @@ goog.provide('ga_search_service');
...
+                if (ol.extent.containsCoordinate(extent, position)) {
+                  defer.resolve(roundCoordinates(position));
+                } else {
+                  defer.resolve(undefined);
+                }
               }
-            }
+            }, function() {
+              // On error use proj4js
+              position = ol.proj.transform(position, 'EPSG:4326', 'EPSG:21781');
```
oterral: Could we move the proj4 transformation in case of error  in the reframe service ?
with that we could get rid of the `else` line 120 .

src/components/search/SearchService.js
```diff
@@ -100,31 +110,37 @@ goog.provide('ga_search_service');
...
-            } else {
-              position =
-                [left < right ? left : right,
-                  right > left ? right : left];
-              position = ol.proj.transform(position,
-                'EPSG:4326', 'EPSG:21781');
-              if (ol.extent.containsCoordinate(
-                extent, position)) {
-                valid = true;
+            gaReframe.get95To03(position).then(function(response) {
```
oterral: you could return the promise directly.

src/components/search/SearchService.js
```diff
@@ -100,31 +110,37 @@ goog.provide('ga_search_service');
...
+              position = response.data.coordinates;
+              if (ol.extent.containsCoordinate(extent, position)) {
+                defer.resolve(roundCoordinates(position));
```
oterral: if you return the promise directly line 116 just do `return  roundCoordinates(position)`

src/components/search/SearchService.js
```diff
@@ -100,31 +110,37 @@ goog.provide('ga_search_service');
...
+                if (ol.extent.containsCoordinate(extent, position)) {
+                  defer.resolve(roundCoordinates(position));
+                } else {
+                  defer.resolve(undefined);
+                }
               }
-            }
+            }, function() {
+              // On error use proj4js
+              position = ol.proj.transform(position, 'EPSG:2056', 'EPSG:21781');
```
oterral: Could we move the proj4 transformation in case of error in the reframe service ?
with that we could get rid of the else line 120 .

src/components/search/SearchService.js
loicgasser: But then it is not self contained anymore? I need to put the code here-under in searchdirective. not sure I understand. the idea is to always return a promise with the final coordinates. so that you simply need to do
` gaSearchGetCoordinate(extent, query).then(function(coords) {});` and you're ensured you can use the `coords directly`

src/components/search/SearchService.js
loicgasser: By using a deferred promise in gaReframe maybe?

oterral: yes

src/components/search/SearchService.js
oterral: yes that's my point :) 

doing `return gaReframe.get95To03(position).then()`  and makes the reframe service uses  a deffered promise so it always returns a coords or undefined .

loicgasser: ok gotcha

loicgasser:
adapted according to your comments thanks.

src/components/search/SearchService.js
```diff
@@ -100,31 +110,21 @@ goog.provide('ga_search_service');
...
+              if (ol.extent.containsCoordinate(extent, position)) {
+                return roundCoordinates(position);
               }
-            }
+            });
           }
+        } else if (valid) {
+          defer.resolve(roundCoordinates(position));
+        } else {
+          defer.resolve(undefined);
```
oterral: Just thinking that  for the sake of clarity we coulduse  `$q.when(the value)` every where instead of the `defer.resolve` variable.

loicgasser: ok

loicgasser: ok

loicgasser:
done

oterral:
Can you add test for the reframe service ?  with tests in case of request errors

loicgasser:
sure, note that I already check that it is called in https://github.com/geoadmin/mf-geoadmin3/pull/3682/commits/12cc2fa3856b6eeb27cb7f9f56a0c5c43fe44786

test/specs/search/SearchService.spec.js
```diff
@@ -2,68 +2,130 @@ describe('ga_search_service', function() {
...
+        gaReframe = $injector.get('gaReframe');
       });
     });
 
     it('supports CH1903 coordinate', function() {
-      expect(getCoordinate(extent, '600000 200000')).to.eql([600000, 200000]);
+      var spy = sinon.spy(gaReframe, 'get95To03')
+      getCoordinate(extent, '600000 200000').then(function(position) {
+        expect(position).to.eql([600000, 200000]);
+        expect(spy.callCount).to.eql(0);
```
loicgasser: expect test should be outside..

loicgasser: or i should  use the done function in it('a', function(done){} right?

oterral: Yes exactly otherwise the test would be always true.

oterral: and launch a `rootScope.$digest()` after the call to `getCoordinate`

loicgasser: yep

test/specs/search/SearchService.spec.js
```diff
@@ -1,69 +1,195 @@
...
+      });
+      $rootScope.$digest();
+      getCoordinate(extent, '7° 1\' 25.0\'\' E 46° 1\' 25.0\'\' N').then(function(position) {
+        expect(position).to.eql([567859.21, 96981.625]);
+        done();
+      });
+      $rootScope.$digest();
+      getCoordinate(extent, '46° 1\' 25.0\'\' N 7° 1\' 25.0\'\' E').then(function(position) {
+        expect(position).to.eql([567859.21, 96981.625]);
+        done();
```
oterral: You can't use multiple time done functions in a `it()`  . Because  it consider the test good only if one `done()` is executed not all of them.

You need to use : 
```
$q.all([promise1, promise2, promise3]).then(function() {
  done()
}); 
```

test/specs/ReframeService.spec.js
```diff
@@ -0,0 +1,46 @@
...
+    beforeEach(function() {
+      inject(function($injector, gaGlobalOptions) {
+        $rootScope = $injector.get('$rootScope');
+        $httpBackend = $injector.get('$httpBackend');
+        gaReframe = $injector.get('gaReframe');
+        lv03tolv95Url = gaGlobalOptions.lv03tolv95Url;
+        lv95tolv03Url = gaGlobalOptions.lv95tolv03Url;
+      });
+    });
+
```
oterral: You can also add a test if the request is successful to be sure the godd value is returned.

loicgasser:
I rebased fixed the tests

test/specs/search/SearchService.spec.js
loicgasser: Ok I just gave a name to each test, it's clearer when debugging.

test/specs/ReframeService.spec.js
loicgasser: tests added

oterral:
Perfect thanks !!!

Issue #3714, opened 2017-02-07
Impossible to print vector polygons with 3 points


oterral:
[Test](https://map.geo.admin.ch/?X=214089.51&Y=692860.69&zoom=6&topic=ech&lang=de&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege,KML%7C%7Chttps:%2F%2Fpublic.geo.admin.ch%2FyK1QZOcrSeiPFdjmnZJqhg&layers_visibility=false,false,false,true,true&layers_timestamp=18641231)

Java error:
```
Error while generating PDF:
DocumentException: java.lang.IllegalArgumentException: Invalid number of points in LinearRing (found 3 - must be 0 or >= 4)
	at com.vividsolutions.jts.geom.LinearRing.validateConstruction(LinearRing.java:114)
	at com.vividsolutions.jts.geom.LinearRing.<init>(LinearRing.java:106)
.
.
.
```

gjn:
We try to fix the line sent to the print.

procrastinatio:
How do we validate geometries in OL3 ? For instance, this [KML](https://s.geo.admin.ch/71b617fba5) has 7 features, of which 4 are non valid and hence not displayed ?

PR #3728, opened 2017-03-03
Removing features with invalid geometries before printing

procrastinatio:
Trying to fix https://github.com/geoadmin/mf-geoadmin3/issues/3714

The application is sending invalid GeoJSON geometries to the print server. For the record, the [GeoJSON format](https://tools.ietf.org/html/rfc7946)

The following [KML](https://www.procrastinatio.org/kml/2points-multipolygons.kml) has invalid (multi)polygons

On prod, this KML makes print crash:
https://s.geo.admin.ch/71bfa62ee8

With this PR, invalid geometries are removed:

https://mf-geoadmin3.int.bgdi.ch/mom_geom_valid

https://mf-chsdi3.int.bgdi.ch/shorten/71bfa870ae

### DIFF
--- a/src/components/GeomUtilsService.js
+++ b/src/components/GeomUtilsService.js
@@ -21,6 +21,15 @@ goog.provide('ga_geomutils_service');
       }
     };
 
+    // Ensure valid linear ring
+    var isValidLinearRing = function(linearRing) {
+      var coords = linearRing.getCoordinates();
+      if (coords.length <= 3) {
+         return false;
+      }
+      return true;
+    };
+
     // Ensure polygon is closed
     var closePolygon = function(polygon) {
       var coords = [];
@@ -100,8 +109,10 @@ goog.provide('ga_geomutils_service');
             } else if (geometry instanceof ol.geom.Polygon) {
                isValid = isValidMultiGeom(this, geometry,
                    geometry.getLinearRings());
-            } else if (geometry instanceof ol.geom.LinearRing ||
-                geometry instanceof ol.geom.LineString) {
+            } else if (geometry instanceof ol.geom.LinearRing) {
+              isValid = isValidLinearRing(geometry) &&
+                !this.hasUniqueCoord(geometry.getCoordinates());
+            } else if (geometry instanceof ol.geom.LineString) {
               isValid = !this.hasUniqueCoord(geometry.getCoordinates());
             }
             if (!isValid) {
--- a/src/components/print/PrintDirective.js
+++ b/src/components/print/PrintDirective.js
@@ -331,8 +331,28 @@ goog.require('ga_urlutils_service');
               styles = ol.style.Style.defaultFunction(feature);
             }
           }
-
+          // Removing feature with invalid geometries
           var geometry = feature.getGeometry();
+          var polygons = [];
+          if (geometry.getType() == 'MultiPolygon') {
+            polygons = geometry.getPolygons();
+          }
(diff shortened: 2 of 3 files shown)


loicgasser:
Nice in general for these kind of things a service would be nice, so that you can test, re-use it.

also for inspiration:
https://github.com/bjornharrtell/jsts/blob/c742f471bf3357cc42a8f7d2bb229ed71f7df03a/src/org/locationtech/jts/operation/polygonize/EdgeRing.js#L50

loicgasser:
https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/GeomUtilsService.js#L47

also

gjn:
This is good for me. Thanks alot!

gjn:
Could you put it in service as @loicgasser suggested?

procrastinatio:
OL3 has no problem displaying invalid polygon (not closed, only three coordinates, or with a null area), which produces invalid GeoJSON. All tests are failing.

But in the first place, the KML are bogus, they should be tested first.

loicgasser:
yes, we should correct the kml before adding it to the map I agree.

loicgasser:
but instead of removing them you should fix them.

src/components/GeomUtilsService.js
```diff
@@ -102,7 +111,8 @@ goog.provide('ga_geomutils_service');
                    geometry.getLinearRings());
             } else if (geometry instanceof ol.geom.LinearRing ||
                 geometry instanceof ol.geom.LineString) {
-              isValid = !this.hasUniqueCoord(geometry.getCoordinates());
+              isValid = isValidLinearRing(geometry) &&
```
oterral: geometry can be a LineString and a Line String with 3 coordinates is valid.

gjn:
So this won't fix the original issue anymore?

(I know that the demand to fix this bug with removing the bad stuff is a hack and proper kml validation and handling should be implemented, but we can't do that now).

procrastinatio:
The only goal is not to make the print server crash. If a polygon has less than 4 points, it is discard.

gjn:
Print doesn't crash, but kml is also not shown as before:

https://mf-geoadmin3.int.bgdi.ch/mom_geom_valid/index.html?topic=ech&lang=de&bgLayer=ch.swisstopo.pixelkarte-grau&layers=KML%7C%7Chttps:%2F%2Fpublic.geo.admin.ch%2FyK1QZOcrSeiPFdjmnZJqhg&X=213797.61&Y=691638.32&zoom=7

https://map.geo.admin.ch/?X=214089.51&Y=692860.69&zoom=6&topic=ech&lang=de&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=KML%7C%7Chttps:%2F%2Fpublic.geo.admin.ch%2FyK1QZOcrSeiPFdjmnZJqhg

@davidoesch How tolerant should we be?

procrastinatio:
At least it is coherent between view and print. Otherwise, people will start to complain about features not being printed, and I will have to solve these case.

gjn:
I'm convinced. Merge if @davidoesch agrees.

Effect of this PR: We stop showing invalid geometries that can't be printed.

davidoesch:
MUST:
- Print and View: WYSIWYG 
- What has been drawn with our tools has to be printable

if those two are given:
+1 for merge

gjn:
Thanks alot @procrastinatio !

PR #3596, opened 2016-11-04
Ngeo import

oterral:
[test](https://mf-geoadmin3.int.bgdi.ch/teo_ngeo2/index.html?topic=ech&lang=fr&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege&layers_visibility=false,false,false,false&layers_timestamp=18641231,,,)

To do:

- [x] try to use css file from ngeo
- [x] try to use the ngeo.File service in all the application if possible
- [x] fix translations of ngeo component -> wait for https://github.com/camptocamp/ngeo/pull/2185
- [x] find why the coverage.txt file is disappeared
- [x] fix jenkins
- [x] verfiy that all Makefile targets work
- [x] apply th patch for #3547 in ngeo if necessary
- [x] fix button to open suggestions in ngeo-> wait for https://github.com/camptocamp/ngeo/pull/2186
- [x] fix e2e tests
- [x]  ~~fix width if possible see [comment](https://github.com/geoadmin/mf-geoadmin3/pull/3596#issuecomment-258599059)~~ -> impossible without creating bugs
- [x] finish translation see [comment](https://github.com/geoadmin/mf-geoadmin3/pull/3596#issuecomment-258599059)
- [x]  fix dev mode on Firefox and IE=11 see [comment](https://github.com/camptocamp/ngeo/pull/2158#discussion_r94959491)
- [x]  avoid copy of all ngeo files when we deploy 

IE <=10 in dev mode doesn't work

PR #3731, opened 2017-03-08
Test CORS

loicgasser:
With this PR it is now possible to load external resources in geoadmin without proxy, provided that:

1. It supports https
2. It supports CORS

[Test Link](https://mf-geoadmin3.int.bgdi.ch/gal_proxify/index.html)

### DIFF
--- a/src/components/UrlUtilsService.js
+++ b/src/components/UrlUtilsService.js
@@ -5,7 +5,7 @@ goog.provide('ga_urlutils_service');
 
   module.provider('gaUrlUtils', function() {
 
-    this.$get = function(gaGlobalOptions, $http, $window) {
+    this.$get = function(gaGlobalOptions, $http, $q, $window) {
 
       var UrlUtils = function(shortenUrl) {
 
@@ -38,7 +38,7 @@ goog.provide('ga_urlutils_service');
           return (this.isValid(url) && /^https/.test(url));
         };
 
-        // Test if URL represents resource that needs to pass via ogcProxy
+        // Test if URL represents resource that needs to pass via proxy
         this.needsProxy = function(url) {
           if (this.isBlob(url)) {
             return false;
@@ -48,11 +48,56 @@ goog.provide('ga_urlutils_service');
                   /.*kmz$/.test(url));
         };
 
-        this.proxifyUrl = function(url) {
+        // Test using a head request if the remote resource enables CORS
+        this.isCorsEnabled = function(url) {
+          return $http.head(url, { timeout: 1500 });
+        };
+
+        this.buildProxyUrl = function(url) {
+          if (!this.isValid(url)) {
+            return url;
+          }
+          var parts = /^(http|https)(:\/\/)(.+)/.exec(url);
+          var protocol = parts[1];
+          var resource = parts[3];
+          // proxy is RESTFful, <service>/<protocol>/<resource>
+          return gaGlobalOptions.proxyUrl + protocol + '/' +
+              encodeURIComponent(resource);
+        };
+
+        this.proxifyUrlInstant = function(url) {
           if (this.needsProxy(url)) {
-            return gaGlobalOptions.ogcproxyUrl + encodeURIComponent(url);
+            return this.buildProxyUrl(url);
           }
           return url;
+        };
+
+        this.getCesiumProxy = function() {
+          var that = this;
+          return {
+            getURL: function(resource) {
+              if (that.needsProxy(resource)) {
+                return that.buildProxyUrl(resource);
+              }
+              return resource;
+            }
+          };
+        };
+
+        this.proxifyUrl = function(url) {
+          var that = this;
+          var deferred = $q.defer();
+          if (!this.isBlob(url) && this.isHttps(url) &&
+              !this.isAdminValid(url) && !/.*kmz$/.test(url)) {
+            this.isCorsEnabled(url).then(function(enabled) {
+              deferred.resolve(url);
+            }, function() {
+              deferred.resolve(that.buildProxyUrl(url));
+            });
+          } else {
+            deferred.resolve(this.proxifyUrlInstant(url));
+          }
+          return deferred.promise;
         };
 
         this.shorten = function(url, timeout) {
--- a/src/components/map/KmlService.js
+++ b/src/components/map/KmlService.js
@@ -63,8 +63,16 @@ goog.require('ngeo.fileService');
         // List of google icons: http://www.lass.it/Web/viewer.aspx?id=4
         kml = kml.replace(
           /<href>http(?!(s:\/\/maps\.(google|gstatic)\.com[a-zA-Z\d\.\-\/_]*\.png|s?:\/\/[a-z\d\.\-]*(bgdi|geo.admin)\.ch))/g,
-          '<href>' + gaGlobalOptions.ogcproxyUrl + 'http'
+          '<href>' + gaGlobalOptions.proxyUrl + 'http'
         );
+        // We still need to convert <href>https://proxy.admin.ch/https:// to
+        // <href>https://proxy.admin.ch/https/
+        kml = kml.replace(
+          new RegExp('<href>' + gaGlobalOptions.proxyUrl + 'http://', 'g'),
+          '<href>' + gaGlobalOptions.proxyUrl + 'http/')
+            .replace(
+          new RegExp('<href>' + gaGlobalOptions.proxyUrl + 'https://', 'g'),
+          '<href>' + gaGlobalOptions.proxyUrl + 'https/');
 
         // Replace all http hrefs from *.geo.admin.ch or *.bgdi.ch by https
         // Test regex here: http://regex101.com/r/fY7wB3/5
@@ -352,19 +360,21 @@ goog.require('ngeo.fileService');
           if (gaNetworkStatus.offline) {
             return this.addKmlToMap(map, null, layerOptions, index);
           } else {
-            return $http.get(gaUrlUtils.proxifyUrl(url), {
-              cache: true
-            }).then(function(response) {
-              var data = response.data;
-              var fileSize = response.headers('content-length');
-
-              if (ngeoFile.isKml(data) && ngeoFile.isValidFileSize(fileSize)) {
-                layerOptions.useImageVector = that.useImageVector(fileSize);
-                return that.addKmlToMap(map, data, layerOptions, index);
-              }
-            }, function() {
-              // Try to get offline data if exist
-              return that.addKmlToMap(map, null, layerOptions, index);
+            return gaUrlUtils.proxifyUrl(url).then(function(proxyUrl) {
+              return $http.get(proxyUrl, {
+                cache: true
+              }).then(function(response) {
+                var data = response.data;
+                var fileSize = response.headers('content-length');
+                if (ngeoFile.isKml(data) &&
+                    ngeoFile.isValidFileSize(fileSize)) {
+                  layerOptions.useImageVector = that.useImageVector(fileSize);
+                  return that.addKmlToMap(map, data, layerOptions, index);
+                }
+              }, function() {
+                // Try to get offline data if exist
+                return that.addKmlToMap(map, null, layerOptions, index);
+              });
             });
           }
         };
--- a/src/components/map/MapService.js
+++ b/src/components/map/MapService.js
@@ -779,7 +779,7 @@ goog.require('ga_urlutils_service');
           var dsP = Cesium.KmlDataSource.load(config3d.url, {
             camera: scene.camera,
             canvas: scene.canvas,
-            proxy: new Cesium.DefaultProxy(gaGlobalOptions.ogcproxyUrl)
+            proxy: gaUrlUtils.getCesiumProxy()
           });
           return dsP;
         };
@@ -915,13 +915,14 @@ goog.require('ga_urlutils_service');
               extent: extent
             });
             var setLayerSource = function() {
-              var fullUrl = gaUrlUtils.proxifyUrl(layer.geojsonUrl);
               var geojsonFormat = new ol.format.GeoJSON();
-              $http.get(fullUrl).then(function(response) {
-                olSource.clear();
-                olSource.addFeatures(
-                  geojsonFormat.readFeatures(response.data)
-                );
+              gaUrlUtils.proxifyUrl(layer.geojsonUrl).then(function(proxyUrl) {
+                $http.get(proxyUrl).then(function(response) {
+                  olSource.clear();
+                  olSource.addFeatures(
+                    geojsonFormat.readFeatures(response.data)
+                  );
+                });
               });
             };
 
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,8 @@ LAST_PUBLIC_URL := $(shell if [ -f .build-artefacts/last-public-url ];  then cat
 PRINT_URL ?= //print.geo.admin.ch
 PRINT_TECH_URL ?= //service-print.
 LAST_PRINT_URL := $(shell if [ -f .build-artefacts/last-print-url ]; then cat .build-artefacts/last-print-url 2> /dev/null; else echo '-none-'; fi)
+PROXY_URL ?= //service-proxy.prod.bgdi.ch
+LAST_PROXY_URL ?= $(shell if [ -f .build-artefacts/last-proxy-ulr ]; then cat .build-artefacts/last-proxy-url 2> /dev/null; else echo '-none-'; fi)
 
 PUBLIC_URL_REGEXP ?= ^https?:\/\/public\..*\.(bgdi|admin)\.ch\/.*
 ADMIN_URL_REGEXP ?= ^(ftp|http|https):\/\/(.*(\.bgdi|\.geo\.admin)\.ch)
@@ -52,10 +54,10 @@ GIT_BRANCH := $(shell if [ -f .build-artefacts/deployed-git-branch ]; then cat .
 GIT_LAST_BRANCH := $(shell if [ -f .build-artefacts/last-git-branch ]; then cat .build-artefacts/last-git-branch 2> /dev/null; else echo 'dummy'; fi)
 BRANCH_TO_DELETE ?=
 DEPLOY_ROOT_DIR := /var/www/vhosts/mf-geoadmin3/private/branch
-OL_VERSION ?= v4.1.0 # v4.1.0, 14 april 2017
-OL_CESIUM_VERSION ?= v1.26 # v1.26, 4 april 2017
-CESIUM_VERSION ?= a2c421d39ac3f43b047681049dd517f489747a72 # camptocamp/c2c_patches_vector_tiles_labels, 20 april 2017
-NGEO_VERSION ?= 3ecccd0c805ccf29dc8fe406dd4309bd292d50c8 # master, 16 january 2017
+OL_VERSION ?= v4.1.0 # v4.1.0, April 14 2017
+OL_CESIUM_VERSION ?= v1.26 # # v1.26, April 4 2017
+CESIUM_VERSION ?= a2c421d39ac3f43b047681049dd517f489747a72 # camptocamp/c2c_patches_vector_tiles_labels, April 20 2017
+NGEO_VERSION ?= 4f3430b4d27bb0a7e2f0c7356fbcca8f134991e5 # master, April 27 2017
 DEFAULT_TOPIC_ID ?= ech
 TRANSLATION_FALLBACK_CODE ?= de
 LANGUAGES ?= '[\"de\", \"fr\", \"it\", \"en\", \"rm\"]'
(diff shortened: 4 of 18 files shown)


procrastinatio:
189/78 KB
https://procrastinatio.github.io/kml/bln-style.kmz
https://procrastinatio.github.io/kml/bln-style.kml


1.6/0.2 MB
https://procrastinatio.github.io/kml/simple_005.kml
https://procrastinatio.github.io/kml/simple_005.kmz

4.3/1.4 MB
https://procrastinatio.github.io/kml/simple_00005a.kml
https://procrastinatio.github.io/kml/simple_00005a.kmz


12/4.0 MB
https://procrastinatio.github.io/kml/simple_000005a.kml
https://procrastinatio.github.io/kml/simple_000005a.kmz

loicgasser:
Great ! Thanks a lot !!!

src/components/UrlUtilsService.js
```diff
@@ -48,13 +48,44 @@ goog.provide('ga_urlutils_service');
                   /.*kmz$/.test(url));
         };
 
-        this.proxifyUrl = function(url) {
+        // Test using a head request if the remote resource enables CORS
+        this.isCorsEnabled = function(url) {
+          var deferred = $q.defer();
+          $http.head(url, { timeout: 2500 }).then(function(response) {
```
procrastinatio: Maybe a bit too long 2.5 sec

src/components/UrlUtilsService.js
```diff
@@ -48,13 +48,44 @@ goog.provide('ga_urlutils_service');
...
           if (this.needsProxy(url)) {
             return gaGlobalOptions.ogcproxyUrl + encodeURIComponent(url);
           }
           return url;
         };
 
+        this.proxifyUrl = function(url) {
+          var deferred = $q.defer();
+          if (!this.isBlob(url) && this.isHttps(url) &&
+              !this.isAdminValid(url) && !/.*kmz$/.test(url)) {
```
procrastinatio: We may also consider that all domains under our control ('.bgdi.ch' and '.geo.admin.ch') do not need to be checked and support HTTPS and CORS by default.

src/components/importkml/ImportKmlDirective.js
```diff
@@ -51,24 +51,26 @@ goog.require('ga_urlutils_service');
...
-                $scope.fileSize = fileSize;
-              } else {
+            gaUrlUtils.proxifyUrl($scope.fileUrl).then(function(url) {
+              $http.get(url, {
+                cache: true,
+                timeout: $scope.canceler.promise
+              }).then(function(response) {
+                var data = response.data;
+                var fileSize = response.headers('content-length');
+                if (gaKml.isValidFileContent(data) &&
```
procrastinatio: If filesize is bigger than 4 MB, we cannot use the proxy and we are doomed. Maybe issue a warning.

src/components/UrlUtilsService.js
loicgasser: yes but you will break the certificate. but technically yes

src/components/importkml/ImportKmlDirective.js
loicgasser: I am waiting on a final decision and service deployment before changing the limit. But yes we may change it.

src/components/UrlUtilsService.js
loicgasser: yes, I reduced it to 1.5 sec, but if network is bad on mobile, we have a small buffer.

src/components/UrlUtilsService.js
gjn: That's what the isAdminValid does. So all *.geo.admin.ch addresses are bypassed anyway.

src/components/UrlUtilsService.js
```diff
@@ -48,13 +48,44 @@ goog.provide('ga_urlutils_service');
                   /.*kmz$/.test(url));
         };
 
-        this.proxifyUrl = function(url) {
+        // Test using a head request if the remote resource enables CORS
+        this.isCorsEnabled = function(url) {
+          var deferred = $q.defer();
+          $http.head(url, { timeout: 1500 }).then(function(response) {
+            // If the head request responded successfully it means that CORS
+            // is enabled
```
gjn: Are you sure about that assumption? Don't we need to check the response headers?

I always assumed a correctly formed pre-flight OPTIONS request is needed. See http://stackoverflow.com/questions/12173990/how-can-you-debug-a-cors-request-with-curl (among other resources, CORS is a beast)

gjn:
Please test with the big 3. (send in email this morning).

Can you create testlinks for them?

src/components/UrlUtilsService.js
loicgasser: yes using curl and being in the browser is 2 different things. The browser will block the resource and as a result the request will be blocked.
I'll provide more test cases.

gjn:
Ok, I tested with the big 3. All 3 will still pass by the ogcproxy. So at least 93% of traffic will still pass via ogxproxy.

Anyhow, this is great for other resources that properly support CORS/https.

gjn:
@oterral I'd like your opinion on this.

src/components/UrlUtilsService.js
loicgasser: Now you re right we may have false positive in case the server doesn t allow head requests. Head requests do pre-flight, like get. http://stackoverflow.com/questions/22107694/why-does-a-cross-origin-head-request-need-a-preflight-check
Best would be to simulate the option request and check the content or simply do a first get and use the proxy on error.

loicgasser:
```
$ curl -I https://www.wandern.ch/ewalc/images/ewalk2/696/geodata/geodata_de.xml
HTTP/1.1 200 OK
Date: Fri, 10 Mar 2017 09:46:32 GMT
Server: Apache
Last-Modified: Mon, 30 Jan 2017 20:13:09 GMT
ETag: "4d8d-5475571c34058"
Accept-Ranges: bytes
Content-Length: 19853
Vary: Accept-Encoding
NZSpeedy: OFF
NZX: OK
Content-Type: application/xml
```

They don't support CORS. It's blocked by the browser as expected and we have to use ogcproxy.


https://www.gipfelbuch.ch/global/kml/gipfel/1198/show

Now we have an issue when requesting a resource that has been moved, like the one above (redirection over http).
So we still need to handle this case.

Again this needs to go through ogcproxy as expected. I don

loicgasser: nop not an issue :D

procrastinatio: HEAD on https://www.wandern.ch/ewalc/images/ewalk2/696/geodata/geodata_de.xml gives a 200, but the requests fails  in an ajax context (nor CORS). So the test is good, it has to go though the proxy

procrastinatio: The alternative would be trying to load the resource, without any testing, and use the proxy only if it fails, effectively discouraging the use of the proxy.

src/components/UrlUtilsService.js
```diff
@@ -48,13 +48,44 @@ goog.provide('ga_urlutils_service');
                   /.*kmz$/.test(url));
         };
 
-        this.proxifyUrl = function(url) {
+        // Test using a head request if the remote resource enables CORS
+        this.isCorsEnabled = function(url) {
+          var deferred = $q.defer();
+          $http.head(url, { timeout: 1500 }).then(function(response) {
```
oterral: you can return directly the promise. No need to handle results.

```
return $http.head(url, { timeout: 1500 });
```

src/components/UrlUtilsService.js
```diff
@@ -48,13 +48,44 @@ goog.provide('ga_urlutils_service');
...
+        };
+
+        this.proxifyUrlInstant = function(url) {
           if (this.needsProxy(url)) {
             return gaGlobalOptions.ogcproxyUrl + encodeURIComponent(url);
           }
           return url;
         };
 
+        this.proxifyUrl = function(url) {
```
oterral: I think the content of this function is a duplication of `needsProxy`, you could move the `isCorsEnabled` stuff in the `needsProxy` function then, promisifying :) the `needsProxy` function you keep the `proxifyUrl` function more understandable, and you get rid of `proxifyUrlInstant`:

```
this.proxifyUrl = function(url) {
  var defer = $q.defer();
  this.needsProxy(url).then(function() {
      defer.resolve(gaGlobalOptions.ogcproxyUrl + encodeURIComponent(url));
  }, function() {
      defer.resolve(url);
  })
  return defer.promise;
};
```

loicgasser: yes, though proxify instant is used to determine the url for Cesium WMS sources. I don't think we can set the sources of WMS Cesium layers in an asynchronous way.

oterral: You're right it'sa problem. May be we could add a `sync` parameter to `proxifyUrl` and `needsProxy` functions which skip the HEAD request if true. I think it worth a try.

loicgasser:
FYI I am waiting for https://github.com/geoadmin/mf-geoadmin3/pull/3596 to be merged before adressing any comment since these PR change the code in the same locations.

gjn:
Something I don't understand. The gipfelbuch kml returns a 200 on the HEAD request in the browser, but it's still passed to the ogcproxy:

https://mf-geoadmin3.int.bgdi.ch/gal_proxy_integration/index.html?swisssearch=https://www.skitourenguru.ch/calc_data/kml/Switzerland_RiskSymbol.kml

Shouldn't it try to load it directyl if HEAD returns 200?

Also, the proxy doesn't seem to work currently.

src/components/UrlUtilsService.js
gjn: What about the sync here?

loicgasser:
Head returns a 200 but fails in an ajax context so it s correct because they don t have cors.
Function has been deleted and changed name because marc is working on creating the minimal rights.

loicgasser:
No matter what is the status code if they don t have cors it will fail.

procrastinatio:
The file https://www.skitourenguru.ch/calc_data/kml/Switzerland_RiskSymbol.kml is loaded without the proxy, but lacking CORS headers, it won't display.

If you use a serious site with correct headers (https://www.procrastinatio.org/kml/Les_Bayards-02-02-2015.kml), it works flawlessly.

gjn:
Ok, I think I understand now. Confusing that HEAD request shows ok in the networking tab. Error can be seen in the console tab.

@procrastinatio https://www.skitourenguru.ch/calc_data/kml/Switzerland_RiskSymbol.kml  is loaded via proxy (at least that's what I see in the console)

src/components/UrlUtilsService.js
loicgasser: yep of course..

src/components/UrlUtilsService.js
loicgasser: This is not a duplicate we have `&&` operator instead of `||` in order to test CORS, we must statisfy all these conditions.

I can make one function with a sync param but honeslty, I don't see the point.

loicgasser:
Updated
@oterral do you see any significant limitations to this approach with ngeo import?

oterral:
On the approach no. On the integration maybe, because the  [transformUrl](https://github.com/geoadmin/mf-geoadmin3/pull/3596/files#diff-2037d4809c5b41498dbbfd34d25e52ecR185) function doesn't accept promise, I guess. Probably something we need to modify in ngeo.

loicgasser:
We need to change:

https://github.com/camptocamp/ngeo/blob/master/src/modules/import/importonline.js#L134

procrastinatio:
Buggy KML http://www.nabodat.ch/images/docs/kml/BOKA_1_5000.kml

src/components/importkml/ImportKmlDirective.js
```diff
@@ -51,11 +51,9 @@ goog.require('ga_urlutils_service');
             $scope.canceler = $q.defer();
 
             // Angularjs doesn't handle onprogress event
-            gaUrlUtils.proxifyUrl($scope.fileUrl).then(function(url) {
+            gaUrlUtils.proxifyUrl($scope.fileUrl).then(function(url, headers) {
```
procrastinatio: where it is defined. Now the `x-api_key`header is not sent

procrastinatio:
`zappa` is slowly killing me (https://github.com/geoadmin/service-proxy/issues/25)...

src/components/importkml/ImportKmlDirective.js
```diff
@@ -51,24 +51,27 @@ goog.require('ga_urlutils_service');
...
-              var fileSize = response.headers('content-length');
-              if (gaKml.isValidFileContent(data) &&
-                  gaKml.isValidFileSize(fileSize)) {
-                $scope.userMessage = $translate.instant('upload_succeeded');
-                $scope.fileContent = data;
-                $scope.fileSize = fileSize;
-              } else {
+            gaUrlUtils.proxifyUrl($scope.fileUrl).then(function(opts) {
+              $http.get(opts.url, {
+                headers: opts.headers,
```
oterral: it seems the headers are always the same here. Why not putting it directly here instead of opts obj.

src/components/UrlUtilsService.js
oterral: The point is to have one function for all which returns a url or a promise which returns a url.

src/components/UrlUtilsService.js
```diff
@@ -48,13 +48,54 @@ goog.provide('ga_urlutils_service');
...
+              deferred.resolve({
+                url: that.buildProxyUrl(url),
+                headers: { 'x-api-key': gaGlobalOptions.proxyApiKey }
+              });
+            });
+          } else {
+            deferred.resolve({
+              url: this.proxifyUrlInstant(url),
+              headers: { 'x-api-key': gaGlobalOptions.proxyApiKey }
+            });
```
oterral: the deferred should returns only the url

src/components/importkml/ImportKmlDirective.js
oterral: Forget the comment. It makes sense.

src/components/UrlUtilsService.js
oterral: Forget too. i Understand why.

src/components/UrlUtilsService.js
```diff
@@ -48,13 +48,54 @@ goog.provide('ga_urlutils_service');
...
+
+        this.buildProxyUrl = function(url) {
+          if (!this.isValid(url)) {
+            return url;
+          }
+          var parts = /^(http|https)(:\/\/)(.+)/.exec(url);
+          var protocol = parts[1];
+          var resource = parts[3];
+          return gaGlobalOptions.proxyUrl + protocol + '/' +
+              encodeURIComponent(resource);
```
oterral: Add a comment to explain why we are doing this.

oterral:
Something still bother me with this implementation. We should create a service `gaProxy` 

Instead of doing:

```
gaUrlUtils.proxifyUrl(layer.url).then
```

we should use something : ```  gaProxy.get(layer.url) .then```

but I can live without it.

procrastinatio:
This

    $ curl  -H "Referer: https://tuttu" -o /dev/null -s -w "%{http_code}\n" https://service-proxy.prod.bgdi.ch/https/www.procrastinatio.org/kml/bln-style.kml
    403

And 

    $ curl  -H "Referer: https://map.geo.admin.ch" -o /dev/null -s -w "%{http_code}\n" https://service-proxy.prod.bgdi.ch/https/www.procrastinatio.org/kml/bln-style.kml
    200

gjn:
This is a go for next weeks deploy. The big 3 have tested and are ok with the changes.

Is ogcproxy in production ready? export PROXY_URL=//service-proxy.prod.bgdi.ch

I think this branch needs a rebase.

loicgasser:
waiting for https://github.com/camptocamp/ngeo/pull/2532 and then I'll rebase. yes code is production ready.

loicgasser:
jenkins test

loicgasser:
rebased

loicgasser:
jenkins test

src/components/map/MapService.js
```diff
@@ -779,7 +779,7 @@ goog.require('ga_urlutils_service');
           var dsP = Cesium.KmlDataSource.load(config3d.url, {
             camera: scene.camera,
             canvas: scene.canvas,
-            proxy: new Cesium.DefaultProxy(gaGlobalOptions.ogcproxyUrl)
+            proxy: new Cesium.DefaultProxy(gaGlobalOptions.proxyUrl + 'http/')
```
oterral: why this `+ 'http'` ?

src/components/UrlUtilsService.js
oterral: Please add a comment

src/components/map/KmlService.js
```diff
@@ -63,8 +63,16 @@ goog.require('ngeo.fileService');
...
+          '<href>' + gaGlobalOptions.proxyUrl + 'http'
         );
+        // We still need to convert <href>https://proxy.admin.ch/https:// to
+        // <href>https://proxy.admin.ch/https/
+        kml = kml.replace(new RegExp('<href>' + gaGlobalOptions.proxyUrl +
+                                     'http://', 'g'),
+                          '<href>' + gaGlobalOptions.proxyUrl + 'http/')
+                 .replace(new RegExp('<href>' + gaGlobalOptions.proxyUrl +
+                                     'https://', 'g'),
+                          '<href>' + gaGlobalOptions.proxyUrl + 'https/');
```
oterral: Use a classic indentation it's not understandable like this.

src/components/map/MapService.js
oterral: why not https?

loicgasser:
it works the same way as before (instead of <service>?url=<url>, we have <service>/<protocol>/<resource>, I'll add the comments.
I have some issues updating the submodule with the approach you took, I'll propose something else.

oterral:
so it's not the same :)

src/components/map/MapService.js
loicgasser: That might be an issue in fact.
DefaultProxy expects a full url? Here we use http per default in the background.

loicgasser:
In the current Jenkins build we:
```
error: Your local changes to the following files would be overwritten by checkout:
	src/modules/import/importlocal.js
	src/modules/import/importonline.js
	src/modules/import/wmsgetcap.js
	src/modules/import/wmsgetcapitem.js
Please, commit your changes or stash them before you can switch branches.
```
when updating a module.
simply removing src/ngeo on cleanall is enough to ensure a proper reset after a cleanall.
I also removeed the `git submodule update --init` in Jenkins build script, I guess this is an old artefact.

src/components/UrlUtilsService.js
loicgasser: it's either we have one function and duplicated code, see: `deferred.resolve(this.proxifyUrlInstant(url));` line 86, either we keep it as is.

src/components/map/MapService.js
loicgasser: I am a little puzzled with this part of the code. this function is only used in getOlLayerById but we never have kml data sources in our layersConfig. Can you explain when this is used?

src/components/UrlUtilsService.js
oterral: it's ok keep it as it is

loicgasser:
This will also fix:
https://github.com/geoadmin/mf-geoadmin3/issues/3783
as I updated ngeo.
I re-rebased and added the getCesiumProxy as discussed, I think it is more correct with Cesium as well.

oterral:
There is an issue with osm wms:

[Before](https://map.geo.admin.ch/?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege,WMS%7C%7CLANDSAT%201975%7C%7Chttp:%2F%2Frips-gdi.lubw.baden-wuerttemberg.de%2Farcgis%2Fservices%2Fwms%2FUIS_0100000017900001%2FMapServer%2FWMSServer%7C%7C0%7C%7C1.3.0%7C%7Ctrue&layers_visibility=false,false,false,false,true&layers_timestamp=18641231,,,,&lon=8.36178&lat=47.41519&elevation=10314&heading=360.000&pitch=-40.090)

[With new proxy](https://mf-geoadmin3.int.bgdi.ch/gal_proxify/index.html?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege,WMS%7C%7CLANDSAT%201975%7C%7Chttp:%2F%2Frips-gdi.lubw.baden-wuerttemberg.de%2Farcgis%2Fservices%2Fwms%2FUIS_0100000017900001%2FMapServer%2FWMSServer%7C%7C0%7C%7C1.3.0%7C%7Ctrue&layers_visibility=false,false,false,false,true&layers_timestamp=18641231,,,,&lon=8.36178&lat=47.41519&elevation=10314&heading=360.000&pitch=-40.090)

loicgasser:
@oterral it's fixed thanks for pointing that out!! This WMS is serving gzip images, which does not make much sense, but we have to take into account this exception.

oterral:
perfect for me this PR is good to go

gjn:
A milestone!

Thanks everyone involved, @loicgasser @danduk82 @procrastinatio @oterral .

PR #3738, opened 2017-03-16
Fix #3722: Update size of embed map on iframe load event

### LINKED ISSUE #3722: Bad aspect ratio with embed maps on mobile
This is a bug report. I'm experimenting with your free iFrame integration service. I'm embedding a KML file with some placemarks. On Desktop, everything is running fine. On mobile devices, the aspect ratio of the map is not ok after loading. It's fine again when I rotate my device into portrait mode and back again. But before, the vertical axis is too short, the map looks squeezed. The icon of the KML are as well. Also, the placemark icons are hard to select. To me it appears the sensitive spot is below the icon itself. I have to tap below the icon to select an icon. Again, after rotating my device into portrait mode, everything's working fine.

I checked on iOS and Android, with different browsers (Safari, Firefox, Chrome). Same behaviour everywhere.

I attached 2 screenshots, one with the bad aspect ratio (just after loading the website), another I took rotating my device into portrait mode and back again.

For a limited time, you find an example of this behaviour here: http://wis.slf.ch/versuchsfeld/poc2.html

On iPhone it can be reproduced on Android 7.1.1 not see as well discussion in forum

oterral:
Fix #3722
@marrcl can you test with this url : https://mf-geoadmin3.int.bgdi.ch/teo_fixes/embed.html

### DIFF
--- a/test/specs/map/MapDirective.spec.js
+++ b/test/specs/map/MapDirective.spec.js
@@ -1,21 +1,148 @@
 describe('ga_map_directive', function() {
+  var map, elt, scope, parentScope;
+  var $httpBackend, $compile, $window, $q, $rootScope, gaBrowserSniffer, gaLayers, gaOffline;
 
-  var map, element, scope;
+  var loadDirective = function() {
+    parentScope = $rootScope.$new();
+    var tpl = '<div ga-map ga-map-map="map" ga-map-options="options"></div>';
+    elt = $compile(tpl)(parentScope);
+    $rootScope.$digest();
+    scope = elt.isolateScope();
+  };
 
-  beforeEach(inject(function($rootScope, $compile) {
-    map = new ol.Map({
-      layers: []
+  describe('in all pages', function() {
+    var layer;
+    var expectedUrl = 'https://example.com/all?lang=en';
+    var dfltLayersConfig = {
+      foo: {
+        type: 'wmts',
+        timeEnabled: true,
+        timestamps: ['t0', 't1', 't2']
+      }
+    };
+
+    beforeEach(function() {
+
+      module(function($provide) {
+        $provide.value('gaTopic', {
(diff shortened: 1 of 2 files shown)


marrcl:
@oterral, I added your URL. It's still not working correctly, actually it's worse. Check http://wis.slf.ch/versuchsfeld/poc4.html. But probably only cause the header with search field and menu ist visible....
img_4276

oterral:
@marrcl sorry bad url use embed.html instead of index.html
 https://mf-geoadmin3.int.bgdi.ch/teo_fixes/embed.html

marrcl:
Hint: More narrow iFrame makes it worse, see screenshot.

img_4278

oterral:
@marrcl Seems to work now.

I've teste dwith this codepen on iPhone 4s: http://codepen.io/oterral/pen/ryJVPz

marrcl:
@oterral, confirming, it works. Hope to see this live soon. Thanks you!

Screenshot:
img_4276 2

oterral:
Thx for testing @marrcl . It should be on production on thursday.

test/specs/map/MapDirective.spec.js
```diff
@@ -1,21 +1,156 @@
...
+          }
+        });
+        /*$provide.value('gaLayers', {
+          loadConfig: function(){return $q.when({
+            'foo': {
+              timeEnabled: true,
+              timestamps: ['20171231', '20141231']
+            }
+          })}
+        });*/
```
loicgasser: unused comment?

loicgasser:
LGTM please merge, just a question about a comment you left.

test/specs/map/MapDirective.spec.js
oterral: removed thx

PR #3746, opened 2017-03-23
Cloudfront headers and infra configuration

gjn:
This PR adds an infra configuration to be able to deploy the application to the infra bucket in the same ways as int/prod.

It also changes the headers for the no-cache files (html, appcache, etc) to better work with CloudFront.

Old:
`Cache-Control: no-cache`
New:
`Cache-Control: max-age=0, must-revalidate, s-maxage=60`

The desired effect is that
a) browser never cache it (max-age=0 and must-revalidate)
b) Cloudfront (or other shared caches) can cache it for 1 minute.

Note that a ctrl-F5 should even prevent getting the data from cloudfront.

[Testlink](https://mf-geoadmin3.int.bgdi.ch/gjn_cf/index.html)

### DIFF
--- a/rc_infra
+++ b/rc_infra
@@ -1,0 +1,11 @@
+export KEEP_VERSION=true
+export DEPLOY_TARGET=int
+export E2E_TARGETURL=https://mf-geoadmin3.infra.bgdi.ch
+export APACHE_BASE_PATH=
+export VARNISH_HOSTS=(ip-10-220-6-234.eu-west-1.compute.internal ip-10-220-4-93.eu-west-1.compute.internal)
+export API_URL=//mf-chsdi3.int.bgdi.ch
+export MAPPROXY_URL=//wmts{s}.int.bgdi.ch
+export WMS_URL=//wms-bgdi.int.bgdi.ch
+export SHOP_URL=//shop-bgdi.int.bgdi.ch
+export PUBLIC_URL=//public.int.bgdi.ch
+export PRINT_URL=//service-print.int.ch
--- a/Makefile
+++ b/Makefile
@@ -238,6 +238,10 @@ deploydev:
 		./scripts/deploydev.sh; \
 	fi
 
+.PHONY: s3deployinfra
+s3deployinfra: guard-SNAPSHOT guard-S3_MF_GEOADMIN3_INFRA .build-artefacts/requirements.timestamp
+	./scripts/deploysnapshot.sh $(SNAPSHOT) infra;
+
 .PHONY: s3deployint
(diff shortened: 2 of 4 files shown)


gjn:
I'd like to have this in for next weeks deploy. So we can activate CloudFront afterwards.

loicgasser:
LGTM and I tested seems good to me.
 Now, I want to make sure you also know that
layersConfig and topics services will be cached also for ever. (`cache-control:max-age=31536000, public`)

So you won't be able to update a current deployed version without re-creating a new version.
Did you plan to put Cloudfront in front of the integration bucket as well?

Otherwise we should probably add at least the `must-revalidate` header for chsdi cached services when deploying a branch.

gjn:
Yes, aware of that. It's already the case now. And it's impossible to invalidate browser caches from users anyway.

The point of having versioned versions of json, js, css etc. is to have immutable objects. Immutable objects are an important concept in caching logics. It's far easier to create new versions than it is to manage caching well. 

So creating new versions on each build should actually be the default. I remember we talked about it. We are on S3, so storage should not be an issue. Just create new versions whenever we deploy.

loicgasser:
Yes when we deploy to prod it's not an issue.

When we deploy our branches to int multiple times, it is an issue. Remember you didn't want any version in the deployed branch, just the name.

loicgasser:
yes

gjn:
@oterral @loicgasser Ok to merge this for tomorrow?

gjn:
I'll rebase.

oterral:
+1

rc_infra
```diff
@@ -0,0 +1,11 @@
...
+export DEPLOY_TARGET=int
+export E2E_TARGETURL=https://mf-geoadmin3.infra.bgdi.ch
+export APACHE_BASE_PATH=
+export VARNISH_HOSTS=(ip-10-220-6-234.eu-west-1.compute.internal ip-10-220-4-93.eu-west-1.compute.internal)
+export API_URL=//mf-chsdi3.int.bgdi.ch
+export MAPPROXY_URL=//wmts{s}.int.bgdi.ch
+export WMS_URL=//wms-bgdi.int.bgdi.ch
+export SHOP_URL=//shop-bgdi.int.bgdi.ch
+export PUBLIC_URL=//public.int.bgdi.ch
+export PRINT_URL=//service-print.int.ch
```
loicgasser: You need to remove this file

loicgasser:
`src/` files are not versionned by the way. And using `DEEP_CLEAN` option won't prevent them from being in the cache until the end of times.
The deploybranch command is not handled properly handled. 

Also, I have now idea what cloudfront dist to flush at the moment.

Issue #3155, opened 2016-02-18
[topic selector] order of topic types


cedricmoullet:
![image](https://cloud.githubusercontent.com/assets/319676/13143940/822a5f6c-d647-11e5-8077-61060af86bde.png)

we will solve this as well midterm by: #3157

cedricmoullet:
@mariokeusen ?

mariokeusen:
@AFoletti ?

mariokeusen:
Please reorder the topics according to the following order also because we'll publish the new topic cadastre with the next deploy:
1 blue topics (federal offices)
2 green topics (specialized topics)
3 red topics (general catalogs)

davidoesch:
Now cadastre is wrong as well

thereofore
Please reorder the topics according to the following order also because we'll publish the new topic cadastre with the next deploy:
1 blue topics (federal offices) and ALPHABETICALLY
2 green topics (specialized topics) and ALPHABETICALLY
3 red topics (general catalogs) and ALPHABETICALLY

davidoesch:
Green topics are not alphabetically ordered on mobile and desktop

loicgasser:
On language change reorder topics? Would it not confuse people if we start moving the topics around?

davidoesch:
The standard user will use only one language... So the topic is always at the same position

oterral:
We have a problem here. Because we don't know which topics are blue, red or green in frontend. Any chance to add the info in the [topic](https://mf-geoadmin3.dev.bgdi.ch/ltteo/src/services) results.

gjn:
We'll add needed infomration in topic service (topic group). Then we'll do ordering on client side.

loicgasser:
fixed via https://github.com/geoadmin/mf-geoadmin3/issues/3155

PR #3748, opened 2017-03-23
Sort topic alphabetically and by group id, langs from global options

loicgasser:
[Test](https://mf-geoadmin3.int.bgdi.ch/gal_groupid/index.html?api_url=//mf-chsdi3.dev.bgdi.ch/gal_topic_group)

Fix for: https://github.com/geoadmin/mf-geoadmin3/issues/3155

To merge first 
https://github.com/geoadmin/tool-wms/pull/68
https://github.com/geoadmin/mf-chsdi3/pull/2504

ping @davidoesch for testing


davidoesch:
DE is ok  but FR IT EN is not  ok

loicgasser:
Can you be more specific ? To everything looks ok.

davidoesch:
FR IT EN  topic selector is not translated

currently it does not work at all
image

loicgasser:
Note that some topics are not translated in all languages, but that s already the case in prod.

davidoesch:
image

loicgasser:
Ok, I don t have that but I see.

loicgasser:
what about now?

gjn:
Works for me.

gjn:
hmm. Was to fast. RM seems to be probelmatic:

image

loicgasser:
same in dev see:
https://mf-geoadmin3.dev.bgdi.ch/?topic=ech&lang=rm&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege&layers_visibility=false,false,false,false&layers_timestamp=18641231,,,

Problem is with the name that is too long.

gjn:
So it's a topic not yet in production? Because in production, there's not such a problem.

Which topic? Could we adapt the text?

loicgasser:
image

same issue it's just that it's on the last line so we don't see it as much,

loicgasser:
Infos bains immobigliars

gjn:
@davidoesch @mariokeusen Chance to change that text to something shorter?

loicgasser:
Infos bains immo. would be fine

loicgasser:
selenium is complaining because it's not pointing to the correct backend.

davidoesch:
Inf. bains immobigliars
image

src/components/topic/TopicService.js
```diff
@@ -119,6 +118,28 @@ goog.require('ga_permalink');
...
+          }
+          for (var j in groups) {
+            sortedTopics = sortedTopics.concat(
+              groups[j].sort(function(a, b) {
+                return a.name.localeCompare(b.name);
+              })
+            );
+          }
+          return sortedTopics;
+        };
```
oterral: can't you replace this stuff with an exsiting angular `groupBy` filter ?

src/components/topic/TopicDirective.js
```diff
@@ -31,19 +31,25 @@ goog.require('ga_topic_service');
...
-              scope.$applyAsync(function() {
-                element.find('.ga-topic-item').tooltip({
-                  container: modal,
-                  placement: 'bottom'
+            $rootScope.$on('$translateChangeEnd', function() {
+              scope.topics = gaTopic.sort(gaTopic.getTopics());
+            });
+
+            $translate.onReady(function() {
+              gaTopic.loadConfig().then(function() {
```
oterral: I'm  against  the use of  $translate.onReady because that means we need to wait for translations to load topics and we don't have to wait, that slow down the loading of the page. And if the translations are broken the topic will not load.

I suggest to remove the $translate.onReady and use the translate in the HTML partials as before.

src/components/topic/partials/topic.html
```diff
@@ -8,12 +8,12 @@ <h4 class="modal-title" translate>choose_theme</h4>
       <div class="modal-body">
         <div class="input-modal-container">
           <div class="ga-topic-item"
-               ng-repeat="topic in ::topics"
+               ng-repeat="topic in topics"
```
oterral: the use of `groupBy` filter could avoid to add code

src/components/topic/partials/topic.html
```diff
@@ -8,12 +8,12 @@ <h4 class="modal-title" translate>choose_theme</h4>
...
                ng-click="setActiveTopic(topic)"
                ng-class="{'ga-topic-active': topic == activeTopic}"
                data-original-title="{{topic.tooltip | translate}}">
-            <div translate>{{::topic.id}}</div>
-            <div class="ga-topics-sprite ga-topics-sprite-{{::topic.id}}"></div>
+            <div>{{topic.name}}</div>
```
oterral: as I said in a previous comment why not letting the `translate` attribute ?

src/components/topic/TopicDirective.js
loicgasser: can't sort before you have the translations...

src/components/topic/partials/topic.html
loicgasser: Theoretically, groupBy + orderBy, but good luck

oterral: Have you try `orderBy:['group', 'name']` ?

loicgasser: this will do the trick probably. I will try it out.

src/components/topic/partials/topic.html
loicgasser: once they are translated no need to add useless watchers.

loicgasser:
fixed according to your comments.

loicgasser:
@oterral To keep the async loading of the topics and not wait for translations to be loaded we could :
1. Create a promise using the onReady callback of $translate
2. Use $q.all for both promises (topic and langs)

This way both json are

gjn:
@loicgasser yes, this sound good to me.

loicgasser:
ok, it's done, anything else I am missing? If you're ok give me a +1 and I'll merge/deploy the other PRs.

gjn:
I'd rather have https://github.com/geoadmin/mf-geoadmin3/pull/3751 tested, merged and deployed today. This PR (and the dependants) can wait for next weeks deploy.

That doesn't mean not to merge them. I'll test in afternoon.

Please test https://github.com/geoadmin/mf-geoadmin3/pull/3751 in detail. Quite a few changes and changing from sync to async might have sideeffect.

@loicgasser @oterral Agree with the priorities?

loicgasser:
+1

loicgasser:
I tested quite a bit, I was just a little suspicious about the failing tests.

loicgasser:
fix for https://github.com/geoadmin/mf-geoadmin3/issues/3753 as well

gjn:
This is fine with me. Please proceed.

Note: the mf-chsdi3 PR contains hard-coded stuff that needs to be reverted first.

Thanks.

src/components/topic/TopicDirective.js
```diff
@@ -8,14 +8,30 @@ goog.require('ga_topic_service');
...
         return {
           restrict: 'A',
           templateUrl: 'components/topic/partials/topic.html',
           scope: {},
           link: function(scope, element, attrs) {
             var modal = element.find('.modal');
 
+            var translationsReady = function() {
+              var deferred = $q.defer();
+              $translate.onReady(function() {
```
oterral: `$translate.onReady` already returns a promise, you can use it directly in the `$q.all`

src/components/topic/TopicDirective.js
```diff
@@ -31,7 +47,12 @@ goog.require('ga_topic_service');
...
             });
 
-            gaTopic.loadConfig().then(function() {
+            $rootScope.$on('$translateChangeEnd', function() {
+              scope.topics = gaTopic.getTopics();
+            });
+
+            $q.all([translationsReady(), gaTopic.loadConfig()]).then(
+                function() {
               scope.topics = gaTopic.getTopics();
```
oterral: same comment as before, using translateTopics here makes more sense

src/components/topic/TopicDirective.js
```diff
@@ -31,7 +47,12 @@ goog.require('ga_topic_service');
               }
             });
 
-            gaTopic.loadConfig().then(function() {
+            $rootScope.$on('$translateChangeEnd', function() {
+              scope.topics = gaTopic.getTopics();
```
oterral: I think it makes more sense to use to `translateTopics` here and line 56, to avoid this translation loop is executed on every digest cycle.

src/components/topic/partials/topic.html
```diff
@@ -8,12 +8,12 @@ <h4 class="modal-title" translate>choose_theme</h4>
       <div class="modal-body">
         <div class="input-modal-container">
           <div class="ga-topic-item"
-               ng-repeat="topic in ::topics"
+               ng-repeat="topic in translateTopics(topics) | orderBy:['groupId', 'name']"
```
oterral: Same comment. Using `translateTopics` here you will translate the topics on every digest cycle and not only when the language change.

loicgasser:
jenkins test

src/components/topic/partials/topic.html
loicgasser: what I did intially.

src/components/topic/TopicDirective.js
loicgasser: yep

src/components/topic/TopicDirective.js
loicgasser: you're right

loicgasser:
comments addressed

PR #3752, opened 2017-03-28
Removing invalid WMS urls

procrastinatio:
[Demo](https://mf-geoadmin3.int.bgdi.ch/mom_valid_wms/index.html)


davidoesch:
Thanks... Perfect

PR #3757, opened 2017-03-31
Handle contextmenu and touchevents correctly

### LINKED ISSUE #3749: Firefox: can't open contextpopup
The touch events are activated for contextopup and tooltip on firefox desktop.

oterral:
Fix #3749 

[test](https://mf-geoadmin3.int.bgdi.ch/teo_fixes/index.html?topic=ech&lang=de&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege&layers_visibility=false,false,false,false&layers_timestamp=18641231,,,&X=179307.26&Y=632153.45&zoom=3)

This PR fixes the tooltip and the context popup on hybrid devices (touch and mouse).

### DIFF
--- a/src/components/contextpopup/ContextPopupDirective.js
+++ b/src/components/contextpopup/ContextPopupDirective.js
@@ -1,12 +1,15 @@
 goog.provide('ga_contextpopup_directive');
 
+goog.require('ga_event_service');
 goog.require('ga_networkstatus_service');
 goog.require('ga_permalink');
 goog.require('ga_reframe_service');
 goog.require('ga_what3words_service');
+
 (function() {
 
   var module = angular.module('ga_contextpopup_directive', [
+    'ga_event_service',
     'ga_networkstatus_service',
     'ga_permalink',
     'ga_reframe_service',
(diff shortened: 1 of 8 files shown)


src/components/contextpopup/ContextPopupDirective.js
```diff
@@ -162,32 +180,52 @@ goog.require('ga_what3words_service');
               }
 
               overlay.setPosition(coord21781);
-              showPopover();
+              element.show();
+              // We use a boolean instead of  jquery .is(':visible') selector
+              // because that doesn't work with phantomJS.
+              isPopoverShown = true;
```
gjn: I prefer to initialize such variables that are used in different places (even though they are not needed). It better establishes their scope.

gjn: There's maybe also a way to get rid of it alltogehter and use the state of the element directly.

src/components/map/MapService.js
```diff
@@ -323,65 +327,55 @@ goog.require('ga_urlutils_service');
...
-                  }, 350, false);
-                }
-              }
-              moving = false;
+            // We authorize a move of 5 pixels max for touch device.
+            if (down &&
+                (evt.clientX < down.clientX - 5 ||
+                evt.clientX > down.clientX + 5 ||
+                evt.clientY < down.clientY - 5 ||
+                evt.clientY > down.clientY + 5)) {
```
gjn: Interesting approach!

src/components/contextpopup/ContextPopupDirective.js
oterral: Yes the easiest way was the jquery stuff but phantomjs doesn't like that.

gjn:
Thank you for the adaptions.

PR #3776, opened 2017-04-19
Do not cache named branches

loicgasser:
Because src resources don't hold the version but only the name of the branch.


oterral:
+1

PR #3790, opened 2017-04-28
Increase s-maxage to 15 minutes for critical resources

gjn:
1 minute s-maxage results in some peaks that appear to often. With increasing it to 15 minutes, we hope to address this a little bit better.

Note: you have to wait 15 minutes after a deploy before the new index.html files are visible (that's the downside).


loicgasser:
What do you mean about the peaks? Cloudfront invalidation?

Can we tune it using 5 min first?

gjn:
I've send you an email about the analysis.

PR #3791, opened 2017-04-28
Localcomparator for topic ordering

loicgasser:
Have a look at topic énergie here:

https://map.geo.admin.ch?lang=fr

collection, expression, reverse, comparator

We can't combine multiple comparators for a collection, so I finally had to create my own.

Proof in angular tests:
https://github.com/gkalpak/angular.js/blob/master/test/ng/filter/orderBySpec.js

[Test Link](https://mf-geoadmin3.int.bgdi.ch/gal_localcomparator/index.html?lang=fr)

### DIFF
--- a/src/components/topic/TopicDirective.js
+++ b/src/components/topic/TopicDirective.js
@@ -31,6 +31,21 @@ goog.require('ga_topic_service');
               scope.activeTopic = newTopic;
             };
 
+            scope.localeSensitiveComparator = function(o1, o2) {
+              var v1 = o1.value;
+              var v2 = o2.value;
+              var groupId1 = v1.groupId;
+              var groupId2 = v2.groupId;
+              if (groupId1 != groupId2 && typeof(groupId1) == 'number' &&
+                  typeof(groupId2) == 'number') {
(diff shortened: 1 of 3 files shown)


gjn:
Nice!

src/components/topic/TopicDirective.js
```diff
@@ -31,6 +31,21 @@ goog.require('ga_topic_service');
               scope.activeTopic = newTopic;
             };
 
+            scope.localeSensitiveComparator = function(o1, o2) {
+              var v1 = o1.value;
+              var v2 = o2.value;
+              var groupId1 = v1.groupId;
+              var groupId2 = v2.groupId;
+              if (groupId1 != groupId2 && typeof(groupId1) == 'number' &&
+                  typeof(groupId2) == 'number') {
```
oterral: groupId is not always a number ?

loicgasser: Yes it is, I was thinking that it was better if optional, that was the behaviour before and wanted to make it compapatible geoadmin forks.

PR #3795, opened 2017-05-04
Update ol and ol-cesium

oterral:


oterral:
Trivial update no major changes

PR #3796, opened 2017-05-04
Fix #3054: Use autoprefixer to generate prefixed css

### LINKED ISSUE #3054: Use Autoprefixer
Since [Bootstrap mixins](http://getbootstrap.com/css/#less-mixins-vendor) are deprecated in bootstrap v4, Bootstrap suggest using [Autoprefixer](https://css-tricks.com/autoprefixer/) instead.

oterral:
Before we used [vendor mixins of bootstrap](http://getbootstrap.com/css/#less-mixins-vendor) to generat css for specifix old browsers.

With this PR we remove the use of bs mixins and let [autoprefixer](https://github.com/postcss/autoprefixer) decides which css must be prefixed depending on the [browserslist](https://github.com/ai/browserslist) specified in the package.json file.

For example you can test on old browsers:
    - mobile menu opening
    - background selector list opening

[Test](https://mf-geoadmin3.int.bgdi.ch/fix_3054/index.html?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&X=190000.00&Y=660000.00&zoom=1&dev3d=true&debug)

### DIFF
--- a/src/components/offline/style/offline.less
+++ b/src/components/offline/style/offline.less
@@ -21,11 +21,9 @@
   button {
     transition: all ease 1s;
     transform: rotateY(360deg);
-    -webkit-transform: rotateY(-360deg);
 
     &.ga-btn-active {
       transform: rotateY(360deg);
-      -webkit-transform: rotateY(360deg);
 
       .fa-ga-offline {
         color: #ed1c24;
@@ -115,7 +113,7 @@
     border: @circle-half-px solid #666;
     border-radius: 50%;
     position: absolute;
-    .transition-transform(1s ease);
+    transition: transform 1s ease;
   }
 
 }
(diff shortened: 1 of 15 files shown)


gjn:
What does it do exactly? What components to test?

oterral:
description updated.

src/components/offline/style/offline.less
```diff
@@ -21,11 +21,9 @@
   button {
     transition: all ease 1s;
     transform: rotateY(360deg);
-    -webkit-transform: rotateY(-360deg);
```
gjn: So this line will be added by the postfix routine?

oterral: Exactly

gjn:
Tested on my mobile. Looks good. Thanks alot.

Issue #3794, opened 2017-05-04
Print PDF: marker not shown


davidoesch:
ALL OS / ALL browsers

since last deploy
map.geoa.dmin.ch
Enter "raron"
Click print
result: pdf does not contain marker

expected result:
PDf has a marker

Observation:
IF marker is removed before printing by clicking in search. on chrome PDF generation is faster.

oterral:
The print server can't access this image via cloudfront.

Without cloud front it works.

davidoesch:
I move this to showstopeper since we have a lot of complaints

gjn:
What is the url that fails from the print server?

gjn:
https://map.geo.admin.ch/master/688362f/1704261351/1704261351/img/marker.png

@procrastinatio Any idea why print would have problems with this?

gjn:
I think it's missing  `Access-Control-Allow-Origin: *` header. I try to adapt in cloudfront settings.

gjn:
What I did didn't help.

gjn:
A quick fix would be to use the technical url for the marker URL when passing it to the print. We could do this for next wee.

I tried to find out why print can't load resources from cloudfront, but can't find the logs - or didn't find anything in the logs on the production machines.

procrastinatio:
Please use https://github.com/geoadmin/service-print/issues/29, this is not a mf-geoadmin3 issue.

procrastinatio:
@gjn, it is not a web request, `Access-Control-Allow-Origin: *` should not change anything.

procrastinatio:
@gjn the logs are pretty lame, with only a:
`[Wed May 03 17:03:13 2017] [error] (70007)The timeout specified has expired: proxy: dialog to (null) (*) failed`

gjn:
Might be related to http2. I test.

gjn:
It seems like the Jakarta Commons-HttpClient/3.1 client used by the print can't get anything from cloudfront. It's old client (5 years old) and probably doesn't understand some modern http protocol negotions.

Shortterm solution: use technical address for this resource
Midterm solution: move those resource out of CloudFront
Longterm solution: update print

gjn:
https://github.com/geoadmin/service-print/issues/29

gjn:
[...]

PR #3799, opened 2017-05-05
Avoid using CF addresses for resources in print

gjn:
Here's a fix for https://github.com/geoadmin/mf-geoadmin3/issues/3794

It assures that we don't use CloudFront addresses for our markes in the print. It uses the direct S3 bucket URLS. Not that CloudFront might not be the issue, see https://github.com/geoadmin/service-print/issues/29#issuecomment-299209328

[Testlink](https://mf-geoadmin3.int.bgdi.ch/gjn_marker/index.html?topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers_visibility=false&layers_timestamp=18641231&lang=de&X=137981&Y=631628&zoom=5&crosshair=marker)

### DIFF
--- a/src/components/UrlUtilsService.js
+++ b/src/components/UrlUtilsService.js
@@ -9,6 +9,15 @@ goog.provide('ga_urlutils_service');
 
       var UrlUtils = function(shortenUrl) {
 
+        // Ugly hack because print can't work with most CF distributions. So
+        // we tap s3 backend directly
+        var CF_TO_NONCF = [
+         ['map.geo.admin.ch',
+          's3-eu-west-1.amazonaws.com/mf-geoadmin3-prod-dublin'],
+         ['mf-geoadmin3.int.bgdi.ch',
+          's3-eu-west-1.amazonaws.com/mf-geoadmin3-int-dublin']
+        ];
+
         // from Angular
         // https://github.com/angular/angular.js/blob/v1.4.8/src/ng/directive/input.js#L15
         var URL_REGEXP = new RegExp('^(blob:)?(ftp|http|https):\\/\\/' +
@@ -133,6 +142,19 @@ goog.provide('ga_urlutils_service');
 
         this.getHostname = function(str) {
           return decodeURIComponent(str).match(/:\/\/(.[^/]+)/)[1].toString();
+        };
+
+        // If it's cloudfron URL, replace it with non-CF one
+        this.nonCloudFrontUrl = function(url) {
+          if (this.isValid(url)) {
+            var hostName = this.getHostname(url);
+            for (var i = 0; i < CF_TO_NONCF.length; i++) {
+              if (hostName == CF_TO_NONCF[i][0]) {
+                return url.replace(CF_TO_NONCF[i][0], CF_TO_NONCF[i][1]);
+              }
+            }
+          }
+          return url;
         };
 
         this.append = function(url, paramString) {
--- a/src/components/print/PrintDirective.js
+++ b/src/components/print/PrintDirective.js
@@ -630,7 +630,8 @@ goog.require('ga_urlutils_service');
             'type': 'Vector',
             'styles': {
               '1': { // Style for marker position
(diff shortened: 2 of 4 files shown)


danduk82:
Wouldn't it be better to simply fix the print service instead? We might test if a compiled version with java 7 works, and if yes use this new one.
If there is an urgency for the markers to re-appear in the print, we can temporarly revert to varnish frontends (as easy as a DNS change, TTL = 10m).

gjn:
Yes, this is short term hack.

gjn:
We'll address it for good when we migrate the print. And I don't want to switch back to varnish backeds just because of this.

danduk82:
there is nothing that lasts long as "temporary" solutions...

gjn:
switching back to old varnish backend would also be temporary. And I don't want that.

We have to migrate print this year - and we address it then. I'll make sure that this patch will be removed then.

procrastinatio:
@danduk82 I did compile the printer-server with Java7 and test all sort of combinations from the CLI, which are consistant with what we observe on tomcat. The issue is not with a single certificat or CF: the mix we have is deadly.

We also have strange things in the navigator when mixing CF:

This KML is supposed to have two markers (one is missing)
https://mf-chsdi3.int.bgdi.ch/shorten/7303e1439b

And here are the two markers:
https://s.geo.admin.ch/7303e1b15d

gjn:
The certificate from api3.geo.admin.ch contains map.geo.admin.ch as alternative name.
And the map.geo.admin.ch certificate is a wildcard certificate (*.geo.admin.ch).

Some clients, and as Marc has shown, even modern browsers, seems to be confused by that.

oterral:
Let's do this if you think there is no ohers ways :see_no_evil:

gjn:
Funny thing: you you take the kml of Marc and display it on integration, the marker from map.geo.admin.ch doesn'th show:
[Doesn't work on in](https://mf-geoadmin3.int.bgdi.ch/?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-grau&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege,KML%7C%7Chttps:%2F%2Fwww.procrastinatio.org%2Fkml%2Fprint-different-origin.kml&layers_visibility=false,false,false,false,true&layers_timestamp=18641231,,,,&X=208650.17&Y=638883.43&zoom=3)

[But same file on map.geo does work](https://map.geo.admin.ch/?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-grau&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege,KML%7C%7Chttps:%2F%2Fwww.procrastinatio.org%2Fkml%2Fprint-different-origin.kml&layers_visibility=false,false,false,false,true&layers_timestamp=18641231,,,,&X=208650.17&Y=638883.43&zoom=3)

gjn:
btw: open one of the links above in IE11 of your workstation and zoom/pan around....

oterral:
>Funny thing: you you take the kml of Marc and display it on integration, the marker from >map.geo.admin.ch doesn'th show:
>Doesn't work on in

>But same file on map.geo does work

It's normal when it comes from the same host CORS is not needed.

oterral:
Do you have this kind of things on ie 11?

oterral:
omg the good old error from Surface is now on IE11:
SCRIPT5022: IndexSizeError
Fichier : build.js, ligne : 686, colonne : 367

gjn:
Yes, but only on links with the kml. Without the kml, I don't have it (at least was not able to get it).

Strange Friday.

gjn:
I'll merge this for today. I make sure to remove this ugly fix once we have a solution for the print.

PR #3803, opened 2017-05-10
Don't use AWS based proxy

gjn:
For today's deploy, we don't want to use the AWS backed ogcproxy.

This PR keeps the CORS testing in place, but will use the existing OGC proxy service of the API

[Testlink](https://mf-geoadmin3.int.bgdi.ch/gjn_tempproxy/index.html)


gjn:
Another hack for today - but here we go.

src/components/map/KmlService.js
```diff
@@ -65,14 +65,6 @@ goog.require('ngeo.fileService');
           /<href>http(?!(s:\/\/maps\.(google|gstatic)\.com[a-zA-Z\d\.\-\/_]*\.png|s?:\/\/[a-z\d\.\-]*(bgdi|geo.admin)\.ch))/g,
           '<href>' + gaGlobalOptions.proxyUrl + 'http'
```
oterral: No it's ok it was like this before.

PR #3810, opened 2017-05-18
Auto Update KML at given interval

gjn:
With integrated GeoJSON layers, we can specify an updateDelay to reload the data from source and update the display accordingly. This allows for 'real time' data to be displayed.

This PR adds the same capability to any KML integrated via the PL. It allows to specify a layers_params updateDelay parameter, which, in seconds, specifies the update rate. No lower than 3 seconds is currently allowed. It can be specified as follows:

`layers_params=updateDelay=3`

Note that it follows the same logic as all other layer related parameters. If multiple layers are added, the layers_params must be separated by commas. `layers_params=,updateDelay=3`. Also, layers_params is already used in other contexts, as emapis.

The new `updateDelay` parameter, for now, if not officially communicated. It might be used for some external projects during the summer.

To test, open 2 separate tabs, side-by-side. On the left, past the viewer:
https://mf-chsdi3.int.bgdi.ch/shorten/732851ac87

On the right, paste the editor:
https://mf-chsdi3.int.bgdi.ch/shorten/7327dd87f5

Edit on the right and see your changes appear on the left after a certain delay.

### DIFF
--- a/src/components/map/KmlService.js
+++ b/src/components/map/KmlService.js
@@ -93,10 +93,13 @@ goog.require('ngeo.fileService');
       };
 
       // Read a kml string then return a list of features.
-      var readFeatures = function(kml) {
+      var readFeatures = function(kml, projection) {
 
         // Create the parser
         setKmlFormat();
+
+        // Sanitize KML
+        kml = sanitizeKml(kml);
 
         // Manage networkLink tags
         var all = [];
@@ -106,7 +109,8 @@ goog.require('ngeo.fileService');
           angular.forEach(networkLinks, function(networkLink) {
             if (gaUrlUtils.isValid(networkLink.href)) {
               all.push($http.get(networkLink.href).then(function(response) {
-                return readFeatures(response.data).then(function(newFeatures) {
+                return readFeatures(response.data,
+                                    projection).then(function(newFeatures) {
                   features = features.concat(newFeatures);
                 });
               }));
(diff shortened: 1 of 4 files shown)


loicgasser:
updateDelay is in milliseconds for the geojsons, I would use the same

loicgasser:
I would try to use something else than = within the permalink parameters, because it seems confusing with long permalinks. Maybe : ?

gjn:
As the parameter is communicated, I think it's easier to use seconds.

The = in layers_params is unfortunate and confusing, but I think we already use it in emapis context as well. What could an alternative look like?

gjn:
We could use quotation marks. They work as well:

`layers_params='updateDelay=3'`

For emapis, we have things like:
`layers_params='equery=(typ_code=D41)'`

All not ideal...

loicgasser:
Ok, if already the case with emapis, makes sense to do the same here. Maybe we should change the updateDelay to seconds all together instead. I know people get confused.

gjn:
@oterral Care for a preview?

oterral:
yes

davidoesch:
Hi folks
Since there is a chnage of plans (surprise) theis feature has to be online with the next deploy. The function has to be in production latest by 3rd of july 2017

oterral:
Ok i make the changes

procrastinatio:
Small note, but [NetworkLink](https://developers.google.com/kml/documentation/kmlreference#networklink) has an `refreshInterval`

```
<?xml version="1.0" encoding="UTF-8"?>
<kml xmlns="http://earth.google.com/kml/2.1">
<Document>
  <NetworkLink>
    <Link>
      <href>http://code.google.com/apis/kml/documentation/Point.kml</href>
      <refreshMode>onInterval</refreshMode>
      <refreshInterval>10</refreshInterval>
    </Link>
  </NetworkLink>
</Document>
</kml>



```

PR #3816, opened 2017-05-23
Fix #3808: use direct url when sent to print server

loicgasser:
[TestLink](https://mf-geoadmin3.int.bgdi.ch/gal_exterternal_graphics/index.html?selectedNode=LT1_1&Y=741652.45&X=190634.02&zoom=5&bgLayer=ch.swisstopo.pixelkarte-farbe&lang=fr&topic=ech&layers=KML%7C%7Chttp:%2F%2Fwww.mapplus.ch%2Fmapimage%2Fkml_mapplus_c4a9rmf3a8v77kvs2bd13erpr1.kml)

PR #3817, opened 2017-05-23
[WIP] Revert "Fix #3808: use direct url when sent to print server"

### LINKED ISSUE #3808: Measure labels are not printed
1. Open map.geo.admin.ch
2. Dessiner > Mesurer 
3. Measure something, either radius, line or polygon
4. Print what you just draw

The labels background [1] are not printed.

Related to https://github.com/geoadmin/mf-geoadmin3/issues/3794 and https://github.com/geoadmin/service-print/issues/29

[1] For instance https://map.geo.admin.ch/master/10a257a/1705101128/1705101128/img/bubble.png

oterral:
Reverts geoadmin/mf-geoadmin3#3816

To merge when print is fixed

### DIFF
--- a/src/components/UrlUtilsService.js
+++ b/src/components/UrlUtilsService.js
@@ -151,13 +151,6 @@ goog.provide('ga_urlutils_service');
             for (var i = 0; i < CF_TO_NONCF.length; i++) {
               if (hostName == CF_TO_NONCF[i][0]) {
                 return url.replace(CF_TO_NONCF[i][0], CF_TO_NONCF[i][1]);
-              } else {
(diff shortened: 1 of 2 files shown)


gjn:
I think this does not need reverting. It's reasonable to strip the proxy address for the printing of these resources. So while the other 2 fixes are reasonable to revert, I would keep this one for proxied resources in the print.

I'll prepare a PR that contains a consolidated version.

gjn:
Not needed anymore because https://github.com/geoadmin/mf-geoadmin3/pull/3874 was merged.

Issue #3387, opened 2016-07-05
In searchbox, support search for MGRS coordinates


gjn:
We currently support LV95, WGS84 and LV03 coordinates. We should also support MGRS coordinates.

mariokeusen:
Confirmed, this is a concrete wish from MGDI.

procrastinatio:
Hem, `concrete wish` ?

davidoesch:
very concrete wish

davidoesch:
This is a need from MGDI Customer feedback ID : 2016070514101587

Enter 	"32TMT 47128 15483" as defined in http://earth-info.nga.mil/GandG/publications/tm8358.1/tr83581b.html#ZZ26 in search box and center on coordinate

procrastinatio:
Two remarks:
* The mentionned point is in France, so no map at some scale
* MGRS specification has no space between coordinates. (_3-4.2 A reference is written as an entity without spaces, parentheses, dashes, or decimal points._)

For instance:
**Münster Bern**

```
Points fixes planimétriques (PFP2, PFP3) (mensuration officielle) (Office fédéral de 
topographie swisstopo, canton)
Numéro	CH030000116611667499
Nom officiel (si existant)	verifiziert
Coordonnées MN03 (Y / X)	600955.716 / 199571.734
Coordonnées MN95 (E / N)	2600955.778 / 1199571.797
```

Try with 
```
32TLT80
32TLT8200
32TLT821004
32TLT82130046
32TLT8213500464
```

PR #3821, opened 2017-05-30
MGRS should also coordinates with space lesen

procrastinatio:
MGRS parser should supports coordinates with spaces

[Demo](https://mf-geoadmin3.int.bgdi.ch/mom_fix_mgrs/index.html)

Try with, e.g..

`32TLT 82135  00464` and `32TLT8213500464` (Münster Bern)

Fixes https://github.com/geoadmin/mf-geoadmin3/issues/3387


davidoesch:
tested. ok

PR #3835, opened 2017-06-02
Import wmts

tigerfoot:
With last patches fix some non-sense-if included


oterral:
It's not what I meant.
@Jenselme mentions he adds [this](https://github.com/ioda-net/geo-front3/commit/a898e90186d37db774b62979214ab179cba81b1e)  to fix some bug but I'm not sure it's still relevant. The timeslider is correctly fullfilled from what I see on production.

Issue #3843, opened 2017-06-08
KML Print rotate text 


davidoesch:
Text in rotation mode can not be printed

reproduce
https://s.geo.admin.ch/73ae9a713c 

rotate:
print pdf

result: not all texts are printed
expected result: all texts with annotation are printed

work around:
none

procrastinatio:
Features are not inlcuded in print specs file when the map is rotated

davidoesch:
Below some observation of user

-> will #3846 fix this



I played around a bit with different map scales but same print frames positions and same orientation.
Scaling had directly an influence on the graphic items.
But also moving the map some centimeters in the print frame let change the situation and items becomes visible or fails.
Last but not least, rotating the orientation for some degrees lets the whole thing change and other graphics are concerned.



 
![image](https://user-images.githubusercontent.com/4577727/27022552-d4c3da56-4f4e-11e7-9b3e-9aed13fb0718.png)

Snapshot from print menu in map.geo
All items are correctly shown:
-	A blue track arrow heading north
-	A blue and a red referenced cross
-	The blue and red text annotation of those crosses
-	The magenta routing trackline


After creating PDF exports, several items becomes to be missing.
This depends on the selected map scale (no other changes are made between the exports except scaling the export size (Kartenmassstab / echelle):


 
![image](https://user-images.githubusercontent.com/4577727/27022578-e78633a0-4f4e-11e7-9aa6-02ee13950e33.png)

1:25‘000:
-	Blue textline is missing
-	Red textline is missing
-	Red cross is missing (blue cross is visible)
-	Top of the blue arrow is missing (only a blue line is shown)


 
![image](https://user-images.githubusercontent.com/4577727/27022588-f1341c28-4f4e-11e7-94aa-09d385eb954b.png)

1:50‘000:
-	Blue textline is missing
-	Red textline is missing

But:
-	Blue AND Red crosses are visible
-	Top of the blue arrow is visible

 
![image](https://user-images.githubusercontent.com/4577727/27022597-fc5485d4-4f4e-11e7-94c8-c967736dcffd.png)

1:100‘000:
-	Red textline is missing

But all other items ARE visible:
-	Blue textline
-	Blue AND Red crosses
-	Blue arrow incl. top of arrow
-	Magenta line

 
![image](https://user-images.githubusercontent.com/4577727/27022609-098e8af6-4f4f-11e7-9862-678c6c55cc7b.png)

1:200‘000:
All items ARE visible:
[...]

PR #3846, opened 2017-06-09
Includes all features if map is rotated

procrastinatio:
See https://github.com/geoadmin/mf-geoadmin3/issues/3843

**To test**

Open this [Demo (avec KMLs)](https://mf-chsdi3.int.bgdi.ch/shorten/73b8447c67)
* Rotate the map
* Print

In the PDF you should have:
* Small cross
* Labels
* Arrow

KML to tests:
https://www.procrastinatio.org/kml/666routing.kml,https://www.procrastinatio.org/kml/666tracks.kml,https://www.procrastinatio.org/kml/666ziele.kml

*On screen*
image

*On PDF*
image

*LUBIRE*
Pour générer de nouveau KML

* Open http://lubire.donatellis.ch/desktop.html
* Click on “Zielkarte erfassen”
* Ecrit 141411414 dans l’input Eff# 
* Click sur Darstellen
* Print

### DIFF
--- a/src/components/print/PrintDirective.js
+++ b/src/components/print/PrintDirective.js
@@ -368,8 +368,12 @@ goog.require('ga_urlutils_service');
             feature = new ol.Feature(geometry);
           }
 
-          // Encode a feature if it intersects with the extent
-          if (geometry.intersectsExtent(getPrintRectangleCoords())) {
+          // Encode a feature if it intersects with the extent and
+          // if the map is not rotated
+          var rotation = $scope.map.getView().getRotation();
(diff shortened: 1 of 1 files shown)


procrastinatio:
The `getPrintRectangleCoords` is rotated when the map is rotated, but its coordinates are wrong

Print extent is materialised as a brown rectangle:

map-not-rotated1

with

map-rotated

oterral:
That does the trick but we should get the four coordinates of the print rectangle instead of the bbox.

PR #3849, opened 2017-06-12
Print extent for rotated map

procrastinatio:
As long as we use an `ol.extent` to define the print extent, this is the best approximation we can do. When rotated it is a bit larger than need, but at least correct (brown rectangle with the same orientation as the map and the whole print extent is colored brown, so it is ok)

map-rotated-correct

compare with this:

map-rotated

Try this [demo with many points](https://mf-chsdi3.int.bgdi.ch/shorten/73c741a86e).
- Rotate the map (45° is the worst angle)
- Print
- All features, especially in the top left and bottom right corner should be present.


oterral:
yes sounds good

PR #3866, opened 2017-06-21
Discard layer which have an extent outside the requested print extent

procrastinatio:
With the new WMS and WMTS import, we may now import layers which does not cover the whole Switzerland (e.g. Canton). This PR filters out these layer, which cannot be printed, _i.e._ have an extend not in the requested print extent.

[Demo](https://mf-geoadmin3.int.bgdi.ch/mom_print_layer_extent/index.html)


src/components/print/PrintDirective.js
```diff
@@ -541,6 +541,19 @@ goog.require('ga_urlutils_service');
         if (!layer.visible || layer.opacity == 0) {
           return;
         }
+        // Only print layer which have an extent intersecting the print extent
+        var layerExtent = layer.getExtent();
+        if (layerExtent) {
+          var intersection = ol.extent.getIntersection(layerExtent,
```
oterral: I guess you don't care to get the intersection extent, you can use this instead:

ol.extent.intersects(extent1, extent2)

oterral: all in one  
```
ol.extent.intersects(layer.getExtent() || [], getPrintRectangleCoords())
```

procrastinatio: No `ol.extent.intersects(extent1, extent2)`works only if `extents2` is completely enclosed in `extent1`, which is a bit too restrictive. If only a part of `extent2` is enclosed, I want it to be printed.

src/components/print/PrintDirective.js
```diff
@@ -541,6 +541,17 @@ goog.require('ga_urlutils_service');
         if (!layer.visible || layer.opacity == 0) {
           return;
         }
+        // Only print layer which have an extent intersecting the print extent
+        if (!ol.extent.intersects(layer.getExtent() || [],
+                    getPrintRectangleCoords())) {
```
oterral: indentation

procrastinatio:
jenkins test

PR #3871, opened 2017-06-22
Added arcgisonline WMTS Services

davidoesch:
selected WMTS services  from https://services.arcgisonline.com/arcgis/rest/services


gjn:
All but the last need commas in the end.

procrastinatio:
What is the point to add a meaningless list of service? Without any description it is difficult to figure out what the service is for.

davidoesch:
Those are alternative backgrounds (canvas roadnames) plus higher res sat image layers for border regions (better than spot imagery)

PR #3815, opened 2017-05-23
[WIP] Revert "Fix #3808: use technical url hack"

oterral:
Reverts geoadmin/mf-geoadmin3#3813

PR #3804, opened 2017-05-10
[WIP] Revert "Merge pull request #3799 from geoadmin/gjn_marker"

gjn:
This reverts the ugly hack to fix the print marker problem https://github.com/geoadmin/mf-geoadmin3/issues/3794

Only merge this once we have another solution.

PR #3874, opened 2017-06-23
Revert hack caused by print bug

gjn:
This is a consolidation of https://github.com/geoadmin/mf-geoadmin3/pull/3804/, https://github.com/geoadmin/mf-geoadmin3/pull/3815 and https://github.com/geoadmin/mf-geoadmin3/pull/3817

As said in  https://github.com/geoadmin/mf-geoadmin3/pull/3817, I think it makes sense to keep the unproxify function and send the non-proxy URLs to the print server. They are not impacted with CORS limitations and it decreases our load on the proxy service.

[Testlink](https://mf-geoadmin3.int.bgdi.ch/gjn_revertprint/index.html)

@oterral @procrastinatio @loicgasser If you agree, we can close the other three pull requests and merge this once the print is ready.

### DIFF
--- a/src/components/print/PrintDirective.js
+++ b/src/components/print/PrintDirective.js
@@ -644,7 +644,7 @@ goog.require('ga_urlutils_service');
             'styles': {
               '1': { // Style for marker position
                 'externalGraphic':
-                  gaUrlUtils.nonCloudFrontUrl($scope.options.markerUrl),
+                  gaUrlUtils.unProxifyUrl($scope.options.markerUrl),
                 'graphicWidth': 20,
                 'graphicHeight': 30,
                 // the icon is not perfectly centered in the image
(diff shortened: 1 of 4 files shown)


loicgasser:
+1

src/components/print/PrintDirective.js
```diff
@@ -653,7 +653,7 @@ goog.require('ga_urlutils_service');
                 'graphicYOffset': -30
               }, '2': { // Style for measure tooltip
                 'externalGraphic':
-                  gaUrlUtils.nonCloudFrontUrl($scope.options.bubbleUrl),
+                  gaUrlUtils.unProxifyUrl($scope.options.bubbleUrl),
```
oterral: `bubbleUrl` and `markerUrl` doesn't use proxy I think, since they are geoadmin images

gjn:
@procrastinatio Is your fix deployed on current print? Can we merge this?

loicgasser:
yes it's been deployed when you were gone.
See:
https://github.com/geoadmin/service-print/releases/tag/r_170607

gjn:
Good, let's merge it then.

Thanks for the reviews.

PR #3882, opened 2017-06-29
ltoed_import IGNWMTS

davidoesch:
Added WTMS of IGN france. has a key which works only with our specific referer


cedricmoullet:
ok for me

procrastinatio:
Is there any usage limitation on the key?

cedricmoullet:
No, we have an agreement

PR #3896, opened 2017-07-05
Add 404 page as part of the project

gjn:
This adds a nice 404 page to the project, which can be used in the CF configuration to be displayed in case of a 404 error.

It's all-in-one page.

[Can be seen here](https://mf-geoadmin3.int.bgdi.ch/gjn_404/404.html)

@pauloamado will reduce it's size even more.


oterral:
Nice

gjn:
ok, page is now smallest possible. Please merge when good.

Issue #3820, opened 2017-05-29
Mobile enforcement on Desktop


davidoesch:
lately  we get more often reports that desktop is forwarded to mobile version
". Mir ist dann aufgefallen, dass die Seite mich immer automatisch auf map.geo.admin.ch/mobile.html weiterleitet (Firefox und Internet Explorer) und nicht auf map.geo.admin.ch/index.html bleibt."
seems to be related to  high res retina screens

-> can be reproduced on our win7 buraut small laptops withe IE

proposal to tweak detection (if it makes sense)

[ ] FF and IE agent are always desktop
[ ] allow desktop mode for smaller screens
[ ] other ides?

oterral:
We could removed  the mobile file and only use responsive stuff.

gjn:
Yes, I would prefer this solution as well. But this is more mid-term goal.

Right now, I think we should update our device detection mechanisms. I consider this high priority (we already discussed it last week, I think).

davidoesch:
Can you test this with browser shots? So we can estimate the impact?

davidoesch:
while add it: enable 
- [ ] print
Expert functions
- [ ] draw/measure
- [ ] import

in mobile as well. they work on mobile (at least in landscape mode)
->  most user will be happy

gjn:
We should have a fully responsive page. Meaning, no more redirection and we decide
a) what to show
b) how to show it

based on
I) screen size
II) browser capabilities

@oterral Thoughts?

oterral:
totally agree.

Only problem i see it's AppCache.

gjn:
1) First step is to update the browsersniffer component to contain latest browsers/detections

2) To have real responsive page. The only difference between mobile/desktop is
I) offline button
II) appcache

All the rest to be responsive.

PR #3898, opened 2017-07-06
Remove use of mobile variable when not needed

oterral:
Part of #3820 

[Test](https://mf-geoadmin3.int.bgdi.ch/fix_3820_/index.html?topic=ech&lang=fr&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege&layers_visibility=false,false,false,false&layers_timestamp=18641231,,,)

### DIFF
--- a/src/components/EventService.js
+++ b/src/components/EventService.js
@@ -14,6 +14,25 @@ goog.provide('ga_event_service');
           var evt = event.originalEvent || event;
           return MOUSE_REGEX.test(evt.pointerType || evt.type);
         };
+
+        // Ensure actions on mouseover/out are only triggered by a mouse
+        this.onMouseOverOut = function(elt, onMouseOver, onMouseOut, selector) {
+          var that = this;
+          var cancelMouseEvts = false;
+          elt.on('touchstart mouseover', selector, function(evt) {
+            if (!that.isMouse(evt) || cancelMouseEvts) {
+              cancelMouseEvts = true;
+              return;
+            }
+            onMouseOver(evt);
+          }).on('mouseout', selector, function(evt) {
+            if (!that.isMouse(evt)) {
+              return;
+            }
+            onMouseOut(evt);
+            cancelMouseEvts = false;
+          });
+        };
       };
 
       return new EventManager();
--- a/src/components/EventService.js
+++ b/src/components/EventService.js
@@ -8,6 +8,25 @@ goog.provide('ga_event_service');
     this.$get = function() {
       var MOUSE_REGEX = /mouse/;
 
+      // Ensure actions on mouseover/out are only triggered by a mouse
+      var addMouseOnlyEvents = function(mngr, elt, eventsIn, eventsOut,
+          callbackIn, callbackOut, selector) {
+        var cancelMouseEvts = false;
+        elt.on(eventsIn, selector, function(evt) {
+          if (!mngr.isMouse(evt) || cancelMouseEvts) {
+            cancelMouseEvts = true;
+            return;
+          }
+          callbackIn(evt);
+        }).on(eventsOut, selector, function(evt) {
+          if (!mngr.isMouse(evt)) {
+            return;
+          }
+          callbackOut(evt);
+          cancelMouseEvts = false;
+        });
+      };
+
       var EventManager = function() {
 
         this.isMouse = function(event) {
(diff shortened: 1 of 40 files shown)


src/components/backgroundselector/style/backgroundselector.less
```diff
@@ -55,7 +55,11 @@
     background-color: rgba(0, 0, 0, 0.7);
   }
 
-  &.ga-bg-mobile {
+  @media (max-width: @screen-tablet) {
+
+    .ga-bg-layer-text, .ga-bg-layer-bt-text, .fa-chevron-sign-right {
+      display: none;
+    }
```
gjn: mui bien. using css for display rules instead of javascript.

src/components/catalogtree/CatalogitemDirective.js
```diff
@@ -110,11 +110,20 @@ goog.require('ga_previewlayers_service');
...
+                    return;
+                  }
                   addPreviewLayer(scope.map, scope.item);
-                }).on('mouseleave', function(evt) {
+                }).on('mouseout', function(evt) {
+                  if (!gaEvent.isMouse(evt)) {
+                    return;
+                  }
                   removePreviewLayer(scope.map);
+                  cancelMouseEvents = false;
```
gjn: Similar code as above in the attribution directive. Could maybe be put into a service/common component.

src/components/layermanager/LayermanagerDirective.js
```diff
@@ -314,22 +316,40 @@ goog.require('ga_urlutils_service');
...
+          if (!link.data('bs.tooltip')) {
+            link.tooltip(tooltipOptions);
+          }
+          link.tooltip('show');
+        }).on('mouseout', tooltipOptions.selector, function(evt) {
+          if (!gaEvent.isMouse(evt)) {
+            return;
+          }
+          $(evt.target).tooltip('hide');
+          cancelMouseEvts = false;
```
gjn: Same comment as above. Similar code that might be put into common component.

src/components/attribution/AttributionDirective.js
```diff
@@ -24,23 +24,46 @@ goog.require('ga_map_service');
...
+          var link = $(evt.target);
+          if (!link.data('bs.tooltip')) {
+            link.tooltip(tooltipOptions);
+          }
+          link.tooltip('show');
+        }).on('mouseout', tooltipOptions.selector, function(evt) {
+          if (!gaEvent.isMouse(evt)) {
+            return;
+          }
+          $(evt.target).tooltip('hide');
```
gjn: nice bugfix!

oterral:
Yes totally . This touch/mouse event trick could be reused anywhere I will have a look where to put it. Maybe in the gaEvent service.

oterral:
Trick moved in the gaEvent service

oterral:
Can we merge this?  I have more coming.

oterral:
I update the branch with my last changes. All use of mobile variable have been removed except 2:
- in mainController because it's needed 
- in tooltip because i haven't found a solution yet

oterral:
Ready to review/test

oterral:
I've tested on safari and chrome mobile. All works fine.

src/components/feedback/FeedbackDirective.js
```diff
@@ -19,178 +22,176 @@ goog.require('ga_permalink');
...
+            resetF();
+            scope.success = true;
+          }, function() {
+            resetF();
+            scope.error = true;
+          });
+        };
+      }
+    };
+  });
```
gjn: Not related to this PR, but detected a small bug from the feedback dialog (it's like this in prod as well).

When I draw some elements from the feedback dialog, then I select one and delete it (with delete key), the object is not deleted directly. It's deleted only when clicking somewhere in the map.

Deletion is immediate in the drawing dialog (after confirmation), and it should behave the same from the feedback dialog.

Could you add this to this PR?

src/components/popup/PopupDirective.js
```diff
@@ -71,18 +76,15 @@ goog.require('ga_window_service');
           scope.options.isReduced = false;
 
           // Per default hide the print function
-          if (!angular.isDefined(scope.options.showPrint) ||
-              gaBrowserSniffer.mobile) {
+          if (!angular.isDefined(scope.options.showPrint)) {
```
gjn: This will now show the print button on mobile devices - which is not the case now and doesn't really make sense for mobile devices.

We could say that this is acceptable. Thoughts?

src/components/search/partials/searchtypes.html
```diff
@@ -1,22 +1,8 @@
...
-         ng-mouseenter="preview(res)"
-         ng-mouseleave="out()">
-      <div class="ga-search-item"
-           ng-bind-html="prepareLabel(res.attrs)"
-           title="{{cleanLabel(res.attrs)}}">
-      </div>
-      <button class="ga-btn ga-icon fa fa-info-sign"
-         ng-if="type === 'layers'"
-         ng-click="getLegend($event, res.attrs.layer)">
-      </button>
```
gjn: Nice refactoring

src/components/feedback/FeedbackDirective.js
oterral: A missing digest cycle probably

src/components/popup/PopupDirective.js
oterral: Not very useful but it works also on mobile. I don't think a user might be offended by this button. 

I can also hide it on small screen, if we receive complain about it.

src/components/EventService.js
```diff
@@ -8,12 +8,44 @@ goog.provide('ga_event_service');
...
       var EventManager = function() {
 
         this.isMouse = function(event) {
           var evt = event.originalEvent || event;
           return MOUSE_REGEX.test(evt.pointerType || evt.type);
         };
+
+        // Ensure actions on mouseover/out are only triggered by a mouse
+        this.onMouseOverOut = function(elt, onMouseOver, onMouseOut, selector) {
+          addMouseOnlyEvents(this, elt, 'touchstart mouseover', 'mouseout',
```
oterral: yes I don't like this `this` use either.

oterral:
PR modified according to your comment @gjn . Can you confirm the feedback work as expected?

gjn:
Yes, tested feedback, works as expected. Thanks alot for the adaptions.

oterral:
Thx for the review

PR #3904, opened 2017-07-14
Change topics sprite image

oterral:
[Open topics list](https://mf-geoadmin3.int.bgdi.ch/teo_img_topic/index.html?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege&layers_visibility=false,false,false,false&layers_timestamp=18641231,,,)


oterral:
Blw is the first image in the sprite so `background-position: 0px 0px`, it's the default value so no need to write it. 

And for dev we don't care which image is displayed so it displays by default the first image in the sprite.

PR #3909, opened 2017-07-25
Remove IE tests

loicgasser:
While the page clearly loads in IE these tests keep failing and I can't really tell why...

I think saucelabs drivers for IE are sub optimal. This should fix our nightly build errors once and for all.


loicgasser:
Jenkins test

gjn:
ok. but manual tests on IE11 and IE Edge are required now.

PR #3916, opened 2017-08-03
Use eslint as linter

### LINKED ISSUE #3868: Use eslint to fix indentation issue

oterral:
Fix #3868 
I use the [ standard config rules](https://standardjs.com/rules.html) with some additions to avoid too much changes in the code.

[Test](http://mf-geoadmin3.int.bgdi.ch/teo_lint/index.html)

Remove a lot of dead code


.eslintrc
```diff
@@ -0,0 +1,44 @@
...
+  rules: {
+    eqeqeq: ["off"],
+    indent: [ "error", 2, { 
+      CallExpression: {"arguments": 2},
+      FunctionDeclaration: {"body": 1, "parameters": 2},
+      FunctionExpression: {"body": 1, "parameters": 2},
+      MemberExpression: 2
+    }],
+    max-len: ["error", {
+      code: 80,
```
gjn: Didn't we have 82 before?

.eslintrc
```diff
@@ -0,0 +1,44 @@
...
+      ignoreUrls: true,
+      ignoreRegExpLiterals: true
+    }],
+    one-var: ["off"],
+    operator-linebreak: ["error", "after"],
+    padded-blocks: ["off"],
+    semi: ["off"],
+    space-before-function-paren: ["error", {
+      anonymous: "never",
+      named: "always",
```
gjn: This is quite a big change. As always with style, matter of taste...

.eslintrc
```diff
@@ -0,0 +1,44 @@
...
+      MemberExpression: 2
+    }],
+    max-len: ["error", {
+      code: 80,
+      ignoreUrls: true,
+      ignoreRegExpLiterals: true
+    }],
+    one-var: ["off"],
+    operator-linebreak: ["error", "after"],
+    padded-blocks: ["off"],
```
gjn: uff

src/components/ExportKmlService.js
```diff
@@ -88,17 +88,17 @@ goog.require('ga_browsersniffer_service');
...
-              //force the add of a <Document> node
+              // force the add of a <Document> node
               exportFeatures.push(new ol.Feature());
             }
             kmlString = gaKml.getFormat().writeFeatures(exportFeatures);
             // Remove no image hack
-            kmlString = kmlString.
-                replace(/<Icon>\s*<href>noimage<\/href>\s*<\/Icon>/g, '');
+            kmlString = kmlString
+                .replace(/<Icon>\s*<href>noimage<\/href>\s*<\/Icon>/g, '');
```
gjn: IMO, the older version is much more readably. It's clear that something will come in next line. With new version, it's not clear...

.eslintrc
```diff
@@ -0,0 +1,44 @@
+{
+  extends: "eslint-config-standard",
+  rules: {
+    eqeqeq: ["off"],
```
gjn: This one I would like to have, because it's about correctness.

src/components/UrlUtilsService.js
```diff
@@ -104,9 +104,9 @@ goog.provide('ga_urlutils_service');
         // Remove proxy from the URL
         this.unProxifyUrl = function(url) {
           if (this.isValid(url)) {
-            var reg = new RegExp(['^(http|https):\/\/(service-proxy.',
-                '(dev|int|prod).bgdi.ch|proxy.geo.admin.ch)',
-                '\/(http|https)\/(.*)'].join(''));
+            var reg = new RegExp(['^(http|https)://(service-proxy.',
```
gjn: Are you sure about the non-escapting here and below? It gives me a pattern error on regex101: https://regex101.com/r/dseDKx/1

gjn: When testing in console, it seems like constructor is taking care of it. Anyhow, it might be confusing, since most of our regex is tested using regex101.

src/components/map/KmlService.js
```diff
@@ -62,31 +61,31 @@ goog.require('ngeo.fileService');
         // Test regex here: http://regex101.com/r/tF3vM0/9
         // List of google icons: http://www.lass.it/Web/viewer.aspx?id=4
         kml = kml.replace(
-          /<href>http(?!(s:\/\/maps\.(google|gstatic)\.com[a-zA-Z\d\.\-\/_]*\.png|s?:\/\/[a-z\d\.\-]*(bgdi|geo.admin)\.ch))/g,
-          '<href>' + gaGlobalOptions.proxyUrl + 'http'
+            /<href>http(?!(s:\/\/maps\.(google|gstatic)\.com[a-zA-Z\d.\-/_]*\.png|s?:\/\/[a-z\d.-]*(bgdi|geo.admin)\.ch))/g,
```
gjn: And here escaping is needed again. ok ok.

gjn: Regarding \., don't we want the literal point here? or any charachter?

src/components/map/KmlService.js
```diff
@@ -62,31 +61,31 @@ goog.require('ngeo.fileService');
...
-          '<href>' + gaGlobalOptions.proxyUrl + 'https/');
+                new RegExp('<href>' + gaGlobalOptions.proxyUrl + 'https://', 'g'),
+                '<href>' + gaGlobalOptions.proxyUrl + 'https/');
 
         // Replace all http hrefs from *.geo.admin.ch or *.bgdi.ch by https
         // Test regex here: http://regex101.com/r/fY7wB3/5
         kml = kml.replace(
-          /<href>http(?=s{0}:\/\/[a-z\d\.\-]*(bgdi|admin)\.ch)/g,
-          '<href>https'
+            /<href>http(?=s{0}:\/\/[a-z\d.-]*(bgdi|admin)\.ch)/g,
```
gjn: Same question about literal point.

src/components/map/KmlService.js
```diff
@@ -62,31 +61,31 @@ goog.require('ngeo.fileService');
...
+            '<href>https'
         );
 
         // Replace all old maki urls image by the color service url
         // Test regex here: https://regex101.com/r/rF2tA1/4
         kml = kml.replace(
-          /<href>https?:\/\/[a-z\d\.\-]*(bgdi|geo.admin)\.ch[a-zA-Z\d\-_\/]*img\/maki\/([a-z\-0-9]*-24@2x\.png)/g,
-          '<href>' + gaGlobalOptions.apiUrl + '/color/255,0,0/$2'
+            /<href>https?:\/\/[a-z\d.-]*(bgdi|geo.admin)\.ch[a-zA-Z\d\-_/]*img\/maki\/([a-z\-0-9]*-24@2x\.png)/g,
```
gjn: same about literal point

src/components/map/KmlService.js
gjn: The above `[a-z\d.-]*` is matches exactly the same as  `.*`, which I'm not really a fain. I prefer as strict as possible - so really force literal point.

src/components/map/MapService.js
```diff
@@ -1292,11 +1290,12 @@ goog.require('ga_urlutils_service');
...
-          var dist = Math.sqrt(Math.pow(source[0] - dest[0], 2),
-              Math.pow(source[1] - dest[1], 2));
-          var duration = Math.min(Math.sqrt(300 + dist /
-              view.getResolution() * 1000), 3000);
+          // TODO: use or not the duration?
+          // var source = view.getCenter();
+          // var dist = Math.sqrt(Math.pow(source[0] - dest[0], 2),
+          //    Math.pow(source[1] - dest[1], 2));
+          // var duration = Math.min(Math.sqrt(300 + dist /
+          //    view.getResolution() * 1000), 3000);
```
gjn: What about this todo?

src/components/print/PrintLayerService.js
```diff
@@ -39,12 +39,12 @@ goog.require('ga_urlutils_service');
   });
 
   var pdfLegendString = '_big.pdf';
-  var POINTS_PER_INCH = 72; //PostScript points 1/72"
+  var POINTS_PER_INCH = 72; // PostScript points 1/72"
   var UNITS_RATIO = 39.37; // inches per meter
```
gjn: Not related to this PR: We have the same variables in the directives. Maybe define them in this service and expose them from here to be used in the directive.

src/components/search/SearchService.js
```diff
@@ -171,7 +171,7 @@ goog.require('ga_reframe_service');
     };
 
     var highlightWord = function(strIn, word) {
-      if (!(!!word.length)) {
+      if (!word.length) {
```
gjn: That's a nice one.

.eslintrc
oterral: no It was 80.

.eslintrc
oterral: Yes but there was too much changes. We can add it to another PR.

.eslintrc
oterral: In fact, I think we could put never for 'named' function too. The standard config added spaces everywhere.

src/components/UrlUtilsService.js
oterral: When you're a re using the Regexp constructor you don't need to escape things. I was concerned too but tests passes.

src/components/map/KmlService.js
oterral: no, [.]* is not equivalent to .*, inside a [] that matches the string '.' so it's equivalent to '\.*'

src/components/map/MapService.js
oterral: I was suprised to have a duration of 0 in the animate function, so the code commented was useless. I can remove it.

oterral:
@gjn 

> Do we have a vim configuration that would a) auto idents according to this style and b) allow to style blocks with this style?

I think you can. But I didn't look how yet,

> Shouldn't we apply it to test code as well?
Yes another PR.


Thanks for th review.

src/components/ExportKmlService.js
oterral: Yes. I see what I can do

oterral:
I've modify the PR according your comments @gjn:  
    - override dot-notation rule to put the dot at the end of the line
    - never put a space between the function name and the parameters declaration.

Future PRs: 
    - activate the eqeqeq rule
    - apply aon tests files too

oterral:
Tu use eslint and vim using syntastic.
http://remarkablemark.org/blog/2016/09/28/vim-syntastic-eslint/

gjn:
Perfect, thanks alot @oterral !

PR #3919, opened 2017-08-16
Remove mapproxy URLS, use col/row correctly

gjn:
This removes the mapproxy specific URLS. As we pass everything via CloudFront now, we don't have any difference between mapproxy generated tiles, tod tiles and pre-generated tiles.

Also, this PR assures that for all projections systems, we use col/row in our matrixset, with the only exception being lv03 tiles, which uses row/col.

This also means the this won't work anymore with the 4326 pre-generated tiles, because factually, they are wrong (as they are row/col).

@ltclm Before we go live with lv95, we need to take this into account. We'll need to decide on what to do with the pre-generated tiles.

### DIFF
--- a/src/components/map/MapService.js
+++ b/src/components/map/MapService.js
@@ -402,10 +402,9 @@ goog.require('ga_urlutils_service');
         return domainsArray;
       };
 
-      var Layers = function(dfltWmsSubdomains, dfltWmtsNativeSubdomains,
-          dfltWmtsMapProxySubdomains, dfltVectorTilesSubdomains,
-          wmsUrlTemplate, wmtsGetTileUrlTemplate,
-          wmtsMapProxyGetTileUrlTemplate, terrainTileUrlTemplate,
+      var Layers = function(dfltWmsSubdomains, dfltWmtsSubdomains,
+          dfltVectorTilesSubdomains, wmsUrlTemplate,
+          wmtsGetTileUrlTemplate, terrainTileUrlTemplate,
           vectorTilesUrlTemplate, layersConfigUrlTemplate,
           legendUrlTemplate, imageryMetadataUrl) {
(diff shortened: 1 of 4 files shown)


gjn:
This PR is for the LV95 branch.

ltclm:
you mean with the 4326 pregenerated tiles right, lv95 pregenerated tiles are already in col/row.

src/components/map/MapService.js
```diff
@@ -438,13 +437,16 @@ goog.require('ga_urlutils_service');
...
-        var getWmtsGetTileTpl = function(layer, time, tileMatrixSet,
-            format, useNativeTpl) {
-          var tpl;
-          if (useNativeTpl) {
-              tpl = wmtsGetTileUrlTemplate;
-          } else {
-              tpl = wmtsMapProxyGetTileUrlTemplate;
+        var getWmtsGetTileTpl = function(layer, time, tileMatrixSet, format) {
+          var tpl = wmtsGetTileUrlTemplate;
+          if (tileMatrixSet == '21781') {
```
oterral: This use case doesn't exist in our code. it's 2056 or 4326.

gjn: Correct. LV03 will be thing of the past once this is merged. I adapt.

gjn:
Ok, adapted according to @oterral input.

oterral:
I haven't tested. but code looks good to me.

loicgasser:
thanks!

gjn:
It didn't work. :=)

gjn:
Will post another PR.

loicgasser:
Lol just checked indeed.

gjn:
Added directly to lv95 branch, deploying also to s3 (int).

Issue #3937, opened 2017-08-30
Printing time series on IE9 does not work


rebert:
When printing a Time serie (Multiple PDF) in IE9 the progress stucks: The client gets the printprogress from the cache. Probably a HTTP HEADER no-cache would help.

oterral:
Fixed by #3939

PR #3939, opened 2017-08-30
Disable printprogress caching

loicgasser:
Fix for https://github.com/geoadmin/mf-geoadmin3/issues/3937

[test to use multiprint on ie9](//mf-geoadmin3.int.bgdi.ch/nocache_printprogress/index.html)

### DIFF
--- a/src/components/print/PrintDirective.js
+++ b/src/components/print/PrintDirective.js
@@ -397,8 +397,13 @@ goog.require('ga_urlutils_service');
                 if (!$scope.options.printing) {
                   return;
                 }
+                var noCacheUrl = url;
+                if (gaBrowserSniffer.msie === 9) {
+                  // #3937: Avoid caching of the request by IE9
+                  noCacheUrl += '&' + (new Date()).getTime();
+                }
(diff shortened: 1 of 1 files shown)


src/components/print/PrintDirective.js
```diff
@@ -399,7 +399,8 @@ goog.require('ga_urlutils_service');
                 }
                 canceller = $q.defer();
                 $http.get(url, {
-                  timeout: canceller.promise
+                  timeout: canceller.promise,
+                  cache: false
```
oterral: `false` is already  the default value.

gjn: In any case, the `cache` property is angular internally only, it caches responses in JS scope and doesn't use browsers's cache. Also, it doesn't impact http headers at all, see [doc](https://docs.angularjs.org/api/ng/service/$http).

We should use the 'headers` property to change http headers.

oterral:
I've corrected the PR it work snow.

loicgasser:
sorry I forgot this one

PR #3942, opened 2017-09-06
A littler more strict sanitation of kml content

gjn:
Title says it all.

[Testlink](https://mf-geoadmin3.int.bgdi.ch/gjn_san/index.html)


procrastinatio:
So goal it to avoid these?
```
<foo            >
</foo        >
<bar
/>
```
Looks OK

gjn:
Good find!

As discussed, we'll replace the matching regexp with spaces instead of empty strings.

gjn:
Failure because of Saucelab tests.

PR #3905, opened 2017-07-14
LV95

gjn:
** Do Not Merge Before Tiles On Demand is in Place **

[Testlink](https://mf-geoadmin3.int.bgdi.ch/gjn_lv95/index.html?lang=en&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege&layers_visibility=false,false,false,false&layers_timestamp=18641231,,,&E=2666150.00&N=1188975.00&zoom=2)

Needs https://github.com/geoadmin/mf-chsdi3/pull/2553

- [x] Context Popup
- [x] Map configured for LV95
- [x] Identify requests
- [x] Permalink. Now uses E and N instead of Y and X. Old Permalinks will continue to work.
- [x] WMTS/WMS layers work
- [x] Drawing works
- [x] Query Tool
- [x] Coordinates in search
- [x] Print (see https://github.com/geoadmin/mf-geoadmin3/pull/3982 and https://github.com/geoadmin/service-print/pull/44)
- [x] Height and profile services (see https://github.com/geoadmin/service-alti/pull/26)
- [x] GeoJSON layers
- [x] Search (location and feature search)
- [x] Link with crosshair, permalink in context popup
- [x] Switch to 3D
- [x] Check example WMS/WMTS services for LV95 support
etc...

use:
[Test print](https://mf-geoadmin3.int.bgdi.ch/gjn_lv95/index.html?print_url=//service-print.int.bgdi.ch)`using print_url=//service-print.int.bgdi.ch`
[Test height/profile](https://mf-geoadmin3.int.bgdi.ch/gjn_lv95/index.html?alti_url=//service-alti.int.bgdi.ch) using `alti_url=//service-alti.int.bgdi.ch`

Needs extensive testing

PR #3943, opened 2017-09-06
Tiles on Demand Beta Go Live

gjn:
Currently not working because back-ends are not ready yet.

The idea is to have quite a big flexibility to add/remove layers during the beta. As it's currently set-up, only 4326 will work with Tiles on demand and only the layers specified in code in the array.

[TestLink](https://mf-geoadmin3.int.bgdi.ch/gjn_tod/index.html)

### DIFF
--- a/src/components/map/MapService.js
+++ b/src/components/map/MapService.js
@@ -388,7 +388,7 @@ goog.require('ga_urlutils_service');
 
         var todLayers = [
           'ch.swisstopo.swisstlm3d-karte-farbe.3d',
-          'ch.swisstopo.swisstlm3d-wanderwege'
+          'ch.swisstopo.swisstlm3d-karte-grau.3d'
         ];
 
         var useToD = function(layer, tileMatrixSet) {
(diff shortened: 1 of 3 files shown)


davidoesch:
Volunteer to test... Once ready. https://mf-chsdi3.int.bgdi.ch/shorten/7582cc05de

gjn:
Production is ready now. So you can test.

But note that only 2 layers are passing via ToD now, and that's in 3D background color and 3D Wanderwege.

All else is using standard sources. This will be the behaviour at the go live next week.

gjn:
If you want to test ToD with map.geo, you use the [LV95](https://github.com/geoadmin/mf-geoadmin3/pull/3905) branch, which can be [tested here](https://mf-geoadmin3.int.bgdi.ch/gjn_lv95/index.html?mapproxy_url=%2F%2Ftod{s}.prod.bgdi.ch).

@davidoesch You testlink contains mainly WMS layers. They don't pass via ToD.

PR #3949, opened 2017-09-14
Adding more layers for 3d ToD

gjn:
We are adding more layers to the ToD beta in 3D.

Also, we decrease s-maxage to have shorter revert times.

@oterral @loicgasser I'll merge and deploy to integration for first tests there. Then we'll go live when good.


loicgasser:
+1 let me know

PR #3962, opened 2017-09-25
Import GPX

oterral:
[Test](https://mf-geoadmin3.int.bgdi.ch/teo_gpx/index.html?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen&layers_visibility=false&layers_timestamp=18641231&X=112724.29&Y=574977.13&zoom=6)


oterral:
jenkins test

procrastinatio:
[Very good](https://mf-chsdi3.int.bgdi.ch/shorten/760d9a375c) with simple GPX. Will test with more complexe one (raw from GPS devices)

Missing "WMS WMTS KML GPX Url"

davidoesch:
very nice! and I do love that it works as well in 3D ;-)

let me knwo when this goes to rpod since we  need to update help and translation

furthermore: let us know WHICH GPX feature  we WON'T support  and how the client has to workaround

loicgasser:
FYI `application/gpx+xml` is already in proxy, just checked.

oterral:
translations fixed

oterral:
jenkins test

procrastinatio:
Tested with real raw GPX from `Garmin`, `Holux`and `WinTec` device.
- tracks are always OK
-`waypoint` sometimes (?)
- `route`  are not disaplayed.
If we want to make use of `waypoint` we must add label for them.

src/components/map/GpxService.js
```diff
@@ -0,0 +1,74 @@
...
+
+        // Sanitize a feature.
+        this.sanitizeFeature = function(feature) {
+          feature.setStyle([this.getStyle()]);
+          return feature;
+        };
+
+        // Get name of the layer.
+        this.getName = function(gpx) {
+          var name = (gpx.match(/<metadata>\s*<name>(.*?)<\/name>/) || [])[1];
```
loicgasser: It can contain a name, it is optional in the spec. Maybe should we have a default name `GPX track` and move that in the if condition.

src/components/map/PermalinkLayersService.js
```diff
@@ -359,16 +361,16 @@ goog.require('ga_wmts_service');
...
           if (adminId) {
             gaFileStorage.getFileUrlFromAdminId(adminId).then(function(url) {
               try {
-                gaKml.addKmlToMapForUrl(map, url, {
+                gaVector.addToMapForUrl(map, url, {
                   adminId: adminId
                 });
               } catch (e) {
-                // Adding KML layer failed, native alert, log message?
+                // Adding vecotr layer failed, native alert, log message?
```
loicgasser: vector

src/components/map/GpxService.js
oterral: I already put a generic default value in the gaVectorService https://github.com/geoadmin/mf-geoadmin3/pull/3962/files#diff-9a44cf53146359d60d3bf77fa1a4d822R140

PR #3977, opened 2017-10-04
Get the correct tile key from tod urls

oterral:


oterral:
@gjn @loicgasser

gjn:
Merci!

gjn:
I'm deploying now to int/prod.

PR #3983, opened 2017-10-05
WMTS per env and parameter in url

loicgasser:
As for the other sevices we have now one service per env, so it make sense to point to the respective env.

Without it we may write unwanted tiles to the prod bucket.

Also you can now use `wmts_url` parameter.

### DIFF
--- a/rc_dev
+++ b/rc_dev
@@ -10,3 +10,4 @@ export SHOP_URL=//shop-bgdi.dev.bgdi.ch
 export PUBLIC_URL=//public.dev.bgdi.ch
 export PRINT_URL=//service-print.dev.bgdi.ch
 export PROXY_URL=//service-proxy.dev.bgdi.ch
+export WMTS_URL=//tod.dev.bgdi.ch
--- a/src/js/GaModule.js
+++ b/src/js/GaModule.js
@@ -158,10 +158,10 @@ goog.require('ga_waitcursor_service');
     gaLayersProvider.wmtsMapProxyGetTileUrlTemplate =
         gaGlobalOptions.mapproxyUrl +
         '/1.0.0/{Layer}/default/{Time}/{TileMatrixSet}/{z}/{x}/{y}.{Format}';
-    gaLayersProvider.wmtsToDUrlTemplate = '//tod{s}.prod.bgdi.ch/' +
-        '1.0.0/{Layer}/default/{Time}/{TileMatrixSet}/{z}/{x}/{y}.{Format}';
-    gaLayersProvider.wmtsToD03UrlTemplate = '//tod{s}.prod.bgdi.ch/' +
-        '1.0.0/{Layer}/default/{Time}/{TileMatrixSet}/{z}/{y}/{x}.{Format}';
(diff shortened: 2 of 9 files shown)


oterral:
looks good can we have a test link pls

loicgasser:
Yes it s deployed. Forgot to put it out there.

https://mf-geoadmin3.int.bgdi.ch/wmts_param_envs/index.html

test/specs/Loader.spec.js
```diff
@@ -74,9 +76,11 @@ beforeEach(function() {
     gaLayersProvider.dfltToDSubdomains = ['5', '6', '7', '8', '9'];
     gaLayersProvider.dfltWmtsMapProxySubdomains = ['5', '6', '7', '8', '9'];
     gaLayersProvider.dfltVectorTilesSubdomains = ['100', '101', '102', '103', '104'];
+    gaLayersProvider.dfltToDSubdomains = ['100', '101', '102', '103', '104'];
```
oterral: `dfltToDSubdomains` is already set upper. Delete one or the other.

rc_dev
```diff
@@ -10,3 +10,4 @@ export SHOP_URL=//shop-bgdi.dev.bgdi.ch
 export PUBLIC_URL=//public.dev.bgdi.ch
 export PRINT_URL=//service-print.dev.bgdi.ch
 export PROXY_URL=//service-proxy.dev.bgdi.ch
+export WMTS_URL=//tod{s}.dev.bgdi.ch
```
oterral: you can remove the {s}

src/js/GaModule.js
```diff
@@ -158,11 +158,12 @@ goog.require('ga_waitcursor_service');
...
-    gaLayersProvider.wmtsToDUrlTemplate = '//tod{s}.prod.bgdi.ch/' +
-        '1.0.0/{Layer}/default/{Time}/{TileMatrixSet}/{z}/{x}/{y}.{Format}';
-    gaLayersProvider.wmtsToD03UrlTemplate = '//tod{s}.prod.bgdi.ch/' +
-        '1.0.0/{Layer}/default/{Time}/{TileMatrixSet}/{z}/{y}/{x}.{Format}';
-    gaLayersProvider.dfltToDSubdomains = ['100', '101', '102', '103', '104'];
+    gaLayersProvider.wmtsToDUrlTemplate = gaGlobalOptions.wmtsUrl +
+        '/1.0.0/{Layer}/default/{Time}/{TileMatrixSet}/{z}/{x}/{y}.{Format}';
+    gaLayersProvider.wmtsToD03UrlTemplate = gaGlobalOptions.wmtsUrl +
+        '/1.0.0/{Layer}/default/{Time}/{TileMatrixSet}/{z}/{y}/{x}.{Format}';
+    gaLayersProvider.dfltToDSubdomains = gaGlobalOptions.staging === 'dev' ?
```
oterral: If you remove the {s} from rc_dev you don't need to check the staging, subdomains are just ignored.

loicgasser:
done, thanks for the review

oterral:
perfect thx

PR #3984, opened 2017-10-05
Use compressed cesium files in prod mode

### LINKED ISSUE #3972: Highlight for multipoint geometries not working
Issue:
When you click on a point no highlight is done if the geometry coming from the identify request is of type:MultiPoint.
This is clearly a regression, but I can't say when exactly it went to prod

Steps to reproduce:
- activate ch.astra.unfaelle-personenschaeden_motorraeder on map.geo.admin.ch
- click on a point: tooltip comes up normally but no highlight

oterral:
... and also update the compressed version of fastclick and slipjs
... and fix #3972 with the update of ol

[test](//mf-geoadmin3.int.bgdi.ch/teo_uplibs/index.html)


oterral:
jenkins test

PR #3990, opened 2017-10-11
Manage GPX elevation

oterral:
This PR uses the elevation provided by the GPX for the 3d

[Test](//mf-geoadmin3.int.bgdi.ch/teo_fix/index.html)
[447978.txt](https://github.com/geoadmin/mf-geoadmin3/files/1375785/447978.txt)

### DIFF
--- a/src/components/map/GpxService.js
+++ b/src/components/map/GpxService.js
@@ -37,7 +37,7 @@ goog.require('ga_styles_service');
         // Returns the default style for GPX feature
         this.getStyle = function() {
           if (!style) {
-            style = gaStyleFactory.getStyle('kml');
+            style = gaStyleFactory.getStyle('gpx');
           }
           return style;
         };
@@ -49,9 +49,7 @@ goog.require('ga_styles_service');
 
         // Sanitize a feature.
         this.sanitizeFeature = function(feature) {
-          var style = this.getStyle();
-          style.setText();
-          feature.setStyle([style]);
+          feature.setStyle([this.getStyle()]);
 
           var geom = feature.getGeometry();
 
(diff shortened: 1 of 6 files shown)


src/components/map/GpxService.js
```diff
@@ -61,6 +59,11 @@ goog.require('ga_styles_service');
             geom = new ol.geom.Point(geom.getCoordinates().slice(0, 2), 'XY');
             feature.setGeometry(geom);
           }
+
+          if (geom && /^XYZ/.test(geom.getLayout())) {
+            geom.set('altitudeMode', 'relativeToGround');
```
loicgasser: Not sure, best would be to ask the user what he wants.

It's really for flights. It could look like a bug if we don't clamp to ground stuff that we are supposed to be clamped like hikes.

davidoesch:
Imho : if there is no explicit z value: clamp to ground

loicgasser:
That s the case here. So maybe thisi is the way to go.

src/components/map/GpxService.js
oterral: Yes I thought the same, but if the elevation is correct it seems the display looks ok. We'll see if users complain about it.

loicgasser:
we could use the height service for the first point. if +/- 20m -> they are on the ground

oterral:
There is a use case like paragliding where the user walks from one hour so the track must be clampToGround then it takes his paraglide and the track must be relativeToGround. 
Like the gpx example I put in the description of the PR.

procrastinatio:
I guess GPX uses ellipsoidal height. Do we correct it?

davidoesch:
If this is the case.... Yes. Prio 2

procrastinatio:
IMHO, track should be on the ground, no over the building

http://www.via-alpina.org/stage/exportGpx/id/114


gpx [https://mf-chsdi3.int.bgdi.ch/shorten/7670af0173]

davidoesch:
Agree

oterral:
Cesium issue. It's the same for KML so not related to this PR. See https://github.com/geoadmin/mf-geoadmin3/issues/4004

PR #4005, opened 2017-10-24
[WIP] Use nice url for height/profile

procrastinatio:
For at least one year, we have an nice address set for `service-alti` service:
`service-alti.(dev|int|prod).bgdi.ch` and `alti.geo.admin.ch`

We may use it internally

[Test link](//mf-geoadmin3.int.bgdi.ch/mom_use_service_alti/index.html) (should use //service-alti.int.bgdi.ch)


Makefile
```diff
@@ -30,7 +30,7 @@ TECH_SUFFIX = .bgdi.ch
 API_URL ?= //api3.geo.admin.ch
 API_TECH_URL ?= //mf-chsdi3.
 LAST_API_URL := $(call lastvalue,api-url)
-ALTI_URL ?= //api3.geo.admin.ch
+ALTI_URL ?= //alti.geo.admin.ch
 LAST_ALTI_URL := $(call lastvalue,alti-url) 
 ALTI_TECH_URL ?= //mf-chsdi3.
```
oterral: Just see it should be service-alti no?

procrastinatio: `alti.geo.admin.ch` or `service-alti.prod.bgdi.ch`

oterral: So the ALTI_TECH_URL should be //service-alti.

gjn:
Not a fan.

I think we should use the same url as we officially publish. Otherwise, it's just confusing.

oterral:
Why not publishing the new urls ? In particular if we plan a new version of chsdi, that will make sense the user uses the new url right now.

danduk82:
will stick to api3 addresses

PR #4007, opened 2017-10-25
Makes the page fully responsive

### LINKED ISSUE #3820: Mobile enforcement on Desktop

oterral:
This PR fix #3820: 

 - query tool is only display on desktop
 - print, draw, import and compare tool are displayed on all devices 

[Test](//mf-geoadmin3.int.bgdi.ch/teo_3820/index.html)


oterral:
Just saw the responsiveness is applied after a first window resize

davidoesch:
Now link to desktop version has now no impact on iPhone. Current prod does - and works as designed.

davidoesch:
And draw does not load on iPhone neither.... But it does as designed on prod

oterral:
@david Draw added on phone and lin to desktop/mobile removed

davidoesch:
Hi @oterral @gjn 
 we do need a switch between desktop/mobile
There are the use cases where
A) mobile behaviour on desktop
- your mobile device is not recognized as such- and then you can force it to mobile
- you want to use your laptop offline with maps ( SMO use case) you need offline button)


B) Desktop behaviour on mobile
- Advanced user: print and draw also on mobile - it works
- you have a desktop/laptop with a small screen: and are set to mobile--you want to reverse this


I agree, this switch toggle we have now is not optimal but at least the customer has a chance to switch the state. One possible go for me is-- as you have on chrome mobile use the browser setting "switch to mobile" " switch to desktop Version"

your thoughts?

oterral:
LInk test on PR description

davidoesch:
mobile: ok
desktop: default download icon should not be available... only if forced to mobile. Here, the only possibility I see is to do this with a permalink parameter mobile=true added by the user if he wants it

davidoesch:
In this case: no Go. An additional button on desktop by default- a waste of space. 
So I propose: remove download from desktop and show it only on mobile/tablet . See if customer complaints will hit the inbox...

oterral:
on desktop space is not a problem.

This is the perfect  solution for iPad pro and Surface users.

davidoesch:
there is only 1% of the desktop user who will need offline -- so we have to remove uneccesary buttons --http://mapbrief.com/2013/02/19/an-iconography-of-confusion-why-map-portals-dont-work-part-iv/  buttons are like a hydra - you kill one and there wil be 2 others coming. Desktop Space IS a problem-..since not all desktops are big. 

So remove it from the desktop size
in addition it would be nice to have an overview how the new switch works on which screensize group eg based on http://gs.statcounter.com/screen-resolution-stats/all/switzerland/#yearly-2017-2017-bar

image

oterral:
At least let me display it on iPad pro by default. iPad is easily detectable.

davidoesch:
Ok 🤣

oterral:
Offline bt hidden on desktop other then iOS

oterral:
Can we decide what we do with this ?  The current approach seems reasonable to me.
@davidoesch

davidoesch:
+1 for a go. 
We can always roll back.

davidoesch:
With no download button on desktop

oterral:
Yes no download on desktop

PR #4008, opened 2017-10-26
Use official wmts also for 3D background

gjn:
Now that varnish is switched to use the 3D background (current) from ToD, we can remove the exception in the code.

The only remaing exception is swissimage-product in 3D. Varnish still points to traditional wmts bucket, pre-caclulated with AUTOMATA.

Note: when switching swissimage-prodcut, we also need to switch row/col....this will also happen in 3 steps: 1) adapt applicatoin to use tod addresses 2) switch on varnish level 3) switch back 1)

### DIFF
--- a/src/components/map/MapService.js
+++ b/src/components/map/MapService.js
@@ -395,14 +395,6 @@ goog.require('ga_urlutils_service');
           return urls;
         };
 
-        var getWmtsUrl = function(wmtsUrl, layer) {
-          if (layer === 'ch.swisstopo.swisstlm3d-karte-farbe.3d' ||
-              layer === 'ch.swisstopo.swisstlm3d-karte-grau.3d') {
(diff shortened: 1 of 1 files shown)


gjn:
I'll deploy as the change is minor - and lv95 should be merged ASAP.

ltclm:
> The only remaing exception is swissimage-product in 3D. Varnish still points to traditional wmts bucket, pre-caclulated with AUTOMATA.

since automata is not working anymore, we can remove this exception too (see your mail concerning Dop10, swissimage hist)

gjn:
We have one tricky stuff to address: swissimage-product/current/4326 is wrong in col/row alignement on the pre-caculated bucket. And it's used by chmobile.

When switching to ToD, this alignment will be correct, but chmobil would break.

PR #4013, opened 2017-10-27
Remove exception of swissimage-product

oterral:
Show the good image in 3d for swissimage-product layer:

[Before](https://map.geo.admin.ch/index.html?lang=fr&topic=ech&bgLayer=voidLayer&layers=ch.swisstopo.swissimage-product,ch.swisstopo.swissnames3d&lon=7.45102&lat=46.94809&elevation=1169&heading=2.874&pitch=-89.991&layers_visibility=true,false)
[After](https://mf-geoadmin3.int.bgdi.ch/teo_fix_3/index.html?lang=fr&topic=ech&bgLayer=voidLayer&layers=ch.swisstopo.swissimage-product,ch.swisstopo.swissnames3d&lon=7.45027&lat=46.94777&elevation=1061&heading=357.966&pitch=-84.968)

### DIFF
--- a/src/components/map/MapService.js
+++ b/src/components/map/MapService.js
@@ -400,26 +400,17 @@ goog.require('ga_urlutils_service');
           return urls;
         };
 
-        var useLV03Template = function(layer, tileMatrixSet) {
-          if (tileMatrixSet === '21781') {
-            return true;
-          }
-          // 3D swissimage-product and background layers were pre-generated with
-          // LV03 scheme (row/col), which is wrong. With ToD, we will use the
-          // correct scheme (col/row)
-          if (layer === 'ch.swisstopo.swissimage-product' &&
-              tileMatrixSet === '4326') {
-            return true;
-          }
-          return false;
-        };
-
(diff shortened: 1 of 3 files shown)


gjn:
Yes, that's the idea. Let's find the right moment for this. Automata needs to be set up to work with ToD, which is not yet the case.

Note that when we merge this, it will not work in production, as official wmts addresses (wmts*.geo.admin.ch) is using the old [wmts backends](https://git.bgdi.ch/camptocamp/puppetmaster-chtopo/src/staging4/site-modules/chtopo/files/etc/varnish/4.0/routing.production.vcl#L60). 

We will need a 3 step process as described here https://github.com/geoadmin/mf-geoadmin3/pull/4008

In addition to the comments in the other PR, we will also have to take care of chmobil, which are using the current timestamp as well. So they will need to do the 3 step process as well.

Complicated, I know. But it'll get better soon.

oterral:
I've done modifications . Is it what you meant?

oterral:
`swissimage-product` we can, `swissimage` we can't

oterral:
@gjn is it ok ?

PR #4035, opened 2017-11-09
Add sr parameter to htmlPopup as well

nadine-piveteau:
This is needed for cadastral web map html popup.


oterral:
looks good thanks

Issue #4031, opened 2017-11-07
PDF: annotation local file


davidoesch:
add a local kml/gpx file

print PDF

result 
![image](https://user-images.githubusercontent.com/4577727/32512861-ada6d710-c3f8-11e7-8d72-55dbc92d9761.png)

expected result; "map.geo.admin.ch" is not shown (since  we are NOT responsible for local draw) if we do show sth: the sho the drive letter

oterral:
It's also the case in the UI.
@davidoesch  What do you want as attribution for local files ?  nothing?

davidoesch:
local files : NTH is path to the directory (eg "C:\temp\mynice\data\onmynice\windwoscomputer\inmyfile.kml" or at tleast the first 35 char-- 
if not doable with pareto: keep it empty

oterral:
I'm in favor to keep it empty.

gjn:
Or a generic term like 'Local File' in read (to mark it external). Maybe 'Custom File', 'Custom Data', 'Import', or something. But I think it's important to show something, I think.

oterral:
User Data ?

oterral:
Local FIle is fin for me

davidoesch:
+1 ... 
use it for all lang

PR #4058, opened 2017-11-24
Attribution for local files (viewer & print)

procrastinatio:
[Demo](https://mf-geoadmin3.int.bgdi.ch/mom_localfile_attribution/index.html) (add a local GPX or KML)

Fixes https://github.com/geoadmin/mf-geoadmin3/issues/4031

### DIFF
--- a/src/components/map/VectorService.js
+++ b/src/components/map/VectorService.js
@@ -138,7 +138,6 @@ goog.require('ngeo.fileService');
                   id: options.id,
                   adminId: options.adminId,
                   url: options.url,
-                  type: options.type || type,
                   label: options.label || srv.getName(data) || type,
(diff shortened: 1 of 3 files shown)


src/components/map/VectorService.js
```diff
@@ -138,7 +138,7 @@ goog.require('ngeo.fileService');
                   id: options.id,
                   adminId: options.adminId,
                   url: options.url,
-                  type: options.type,
+                  type: options.type || type,
```
oterral: I think you can totally remove this line.  'type' is no longer settable in ol.

gjn:
Would be nice to have this for tomorrw. Can you please add tests, @procrastinatio ?

PR #4060, opened 2017-11-27
Remove offline bt in iframe mode

### LINKED ISSUE #4059: Offline in iframe: remove offline
with #4007 we introduced now the the offline button in iframe (since the average iframe size is now interpreted user is on mobile)

see
https://www.wandern.ch/de/wandern/unsere-wandervorschlaege/aktueller-wandervorschlag
https://www.geo.admin.ch/de/geo-dienstleistungen/geodienste/darstellungsdienste-webmapping-webgis-anwendungen/iframe.html

issue:
- iframe map is already small, now we cover it  with an additional button- and cover precious map space
- there is barely the use case that user wants to save map offline starting in an iframe

Proposal:

remove offline button in when in iframe per default

(BTW, the same fate for the 3d button, we will probabyl integrate it as an additional background type, getting rid of the grey mode and add 3d void as well tbd)

oterral:
[Test](https://codepen.io/oterral/pen/BmPEaG?editors=1000)

Fix #4059

### DIFF
--- a/src/index.mako.html
+++ b/src/index.mako.html
@@ -202,7 +202,9 @@ itemscope itemtype="http://schema.org/WebApplication"
         <div ga-time-selector-bt ga-time-selector-bt-map="map"></div>
   % endif
         <div ga-rotate ga-rotate-map="map"></div>
+  % if device != 'embed':
         <div ga-offline-bt></div>
+  % endif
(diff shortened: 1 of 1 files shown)


davidoesch:
Tested

PR #4068, opened 2017-11-30
Update olcesium and ol

oterral:
Cesium and ol are now loaded via npm modules in olcesium. So we use `npm link` to link ol cesium with our custom version of ol and cesium

[test](//mf-geoadmin3.int.bgdi.ch/teo_up/index.html)


oterral:
Ready to review

oterral:
This time I think it's good to review.

I sent you a big kml to test @loicgasser

loicgasser:
Works well, on Explorer I get unsufficient memory and the browser refuses to load the KML, but this is the same in prod.

oterral:
Labels bug is fixed through a patch. I will remove it in the future version if needed.

loicgasser:
labels ok, chrome, firefox and edge have similar perfs. Good to go for me.

PR #4072, opened 2017-12-05
Fix #4044: Simplify the line if the number of coordinates exceeds 1000

### LINKED ISSUE #4044: Inform user if too many points
While drawing, we are sending the drawn points on a regular basis to our backends. If the line/polygon is huge, this causes a huge request causing problems in the backend.

Once we detect that the number of points is > x points, we inform the user that we don't do the profile anymore. Maybe directly in the profile windows.

x=1000 to start, but needs to be verified.

@nadine-piveteau That could be something for you (no GIFS/Harlekins here, please ;=))

oterral:
[Test](//mf-geoadmin3.int.bgdi.ch/teo_prof/index.html)

Big kml file attached .
[ThomasUlrich (1).txt](https://github.com/geoadmin/mf-geoadmin3/files/1531191/ThomasUlrich.1.txt)

### DIFF
--- a/src/components/GeomUtilsService.js
+++ b/src/components/GeomUtilsService.js
@@ -194,6 +194,18 @@ goog.provide('ga_geomutils_service');
           }
           return geom;
         };
+
+        // Simplify a geometry only if the number of coordinate s exceeds
+        // nbPointsMax 
+        this.simplify = function(geom, nbPointsMax) {
+          if (!(geom instanceof ol.geom.LineString) || !nbPointsMax ||
+              geom.getCoordinates().length <= nbPointsMax) {
+            return geom;
+          }
+          // Magic number coming from service-alti
+          // https://github.com/geoadmin/service-alti/blob/master/alti/views/profile.py#L53
+          return geom.simplify(12.5);
+        };
       };
       return new GeomUtils();
     };
--- a/src/components/GeomUtilsService.js
+++ b/src/components/GeomUtilsService.js
@@ -195,7 +195,7 @@ goog.provide('ga_geomutils_service');
           return geom;
         };
 
-        // Simplify a geometry only if the number of coordinate s exceeds
+        // Simplify a geometry only if the number of coordinates exceeds
         // nbPointsMax 
         this.simplify = function(geom, nbPointsMax) {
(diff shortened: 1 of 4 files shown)


src/components/GeomUtilsService.js
```diff
@@ -194,6 +194,18 @@ goog.provide('ga_geomutils_service');
...
+        // Simplify a geometry only if the number of coordinates exceeds
+        // nbPointsMax 
+        this.simplify = function(geom, nbPointsMax) {
+          if (!(geom instanceof ol.geom.LineString) || !nbPointsMax ||
+              geom.getCoordinates().length <= nbPointsMax) {
+            return geom;
+          }
+          // Magic number coming from service-alti
+          // https://github.com/geoadmin/service-alti/blob/master/alti/views/profile.py#L53
+          return geom.simplify(12.5);
```
gjn: This is the simplification, but it doesn't assure the max number of points which happens here? https://github.com/geoadmin/service-alti/blob/master/alti/views/profile.py#L150

This will probably help already, so good to me to merge.

oterral: After disacuss wthis with @loicgasser 
The profile uses the same method but with 500 instead of 1000. There is no limitations it returns all the  points which are in the geometry simplified. For example ithe profile service returns 5000 points for the geometry ulrich.txt file.

oterral: And with a limitation  to 1000 coords the resulting profile was really not precise.

gjn: So the request on the backend stays large?

The goal is to reduce the size of the request to the backend.

oterral: 5000 points is not large @loicgasserr says [here](https://github.com/geoadmin/mf-geoadmin3/issues/4044) that 150000 points is large.

gjn: Probably true.

The exact number os not so relevant as the fact that we need to avoid sending too much. And as I understand the algorithm, we don't reduce the number of points sent.

So clients could still send things that will break our backends. This we need to avoid.

gjn:
ok, we go with that and see if it helps.

davidoesch:
Can you tell me how many user will be affected with those limit? Eg last six month how many drawings are above this limit?

PR #4077, opened 2017-12-08
 Add templating support for json style

loicgasser:
The templating support is needed for meteo data.

[Test Link](https://mf-chsdi3.dev.bgdi.ch/meteo_temp/shorten/7761f8d696)

I've heard there may be a problem with EDGE, I cannot test right now but keep that in mind before merging.

In the label section of the JSON style, you can now use:

```json
"label": {
  "template": "${firstprop} is better than ${secondprop}"
}
...
```

### DIFF
--- a/src/components/StylesFromLiteralsService.js
+++ b/src/components/StylesFromLiteralsService.js
@@ -79,7 +79,8 @@ goog.provide('ga_stylesfromliterals_service');
         } else {
           angular.forEach(style, function(value, key) {
             if (key === 'label') {
-              olStyles['text'] = getOlBasicStyles(style[key])['text'];
+              olText = getOlBasicStyles(style.label).text;
+              olStyles.text = olText;
             } else if (['stroke', 'fill', 'image'].indexOf(key) !== -1) {
               olStyles[key] = getOlBasicStyles(style)[key];
             }
@@ -105,6 +106,22 @@ goog.provide('ga_stylesfromliterals_service');
         if (value) {
           return value.property;
         }
+      }
+
+      function getLabelTemplate(value) {
+        if (value) {
+          return value.template || '';
+        }
+      }
+
+      function getStyleSpec(value) {
+        return {
+          olStyle: getOlStyleFromLiterals(value),
+          minResolution: getMinResolution(value),
+          maxResolution: getMaxResolution(value),
+          labelProperty: getLabelProperty(value.vectorOptions.label),
+          labelTemplate: getLabelTemplate(value.vectorOptions.label)
+        };
       }
 
       function getMinResolution(value) {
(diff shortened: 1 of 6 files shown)


davidoesch:
Ahh its an issue with IE --- EGDE is fine

loicgasser:
@oterral saucelabs keeps failing. Shall I open a ticket?

src/components/StylesFromLiteralsService.js
```diff
@@ -219,47 +228,69 @@ goog.provide('ga_stylesfromliterals_service');
         return this.defaultStyle;
       };
 
+      OlStyleForPropertyValue.prototype.setOlText_ = function(olStyle,
+        labelProperty, labelTemplate, properties) {
```
oterral: indent

src/components/StylesFromLiteralsService.js
```diff
@@ -219,47 +228,69 @@ goog.provide('ga_stylesfromliterals_service');
...
+        var text;
+        properties = properties || [];
+        if (labelProperty) {
+          text = properties[labelProperty];
+          if (text !== undefined && text !== null) {
```
oterral: The text value is already tested further, no need to do it twice.

src/components/StylesFromLiteralsService.js
```diff
@@ -219,47 +228,69 @@ goog.provide('ga_stylesfromliterals_service');
...
+            text = text.toString();
+          }
+        } else if (labelTemplate) {
+          text = labelTemplate;
+          angular.forEach(properties, function(prop, k) {
+            if (prop !== undefined && prop !== null) {
```
oterral: Can `prop` be null or undefined ?  

in any case,  `if (prop) ` should be enough

src/components/StylesFromLiteralsService.js
```diff
@@ -219,47 +228,69 @@ goog.provide('ga_stylesfromliterals_service');
...
+              text = text.replace('${' + k + '}', prop.toString());
```
oterral: prop is already a string I guess, no nee to convert it.

src/components/StylesFromLiteralsService.js
```diff
@@ -219,47 +228,69 @@ goog.provide('ga_stylesfromliterals_service');
...
+              styleSpec.labelTemplate, properties);
+        }
+        return this.defaultStyle;
+      };
+
       OlStyleForPropertyValue.prototype.getFeatureStyle = function(feature,
           resolution) {
-        var properties, value, geomType, olStyles, res, labelProperty, text;
+        var properties;
+        if (feature) {
```
oterral: In which case a fetaure can be null?

src/components/StylesFromLiteralsService.js
loicgasser: yes anything you could possibly find the properties of a GeoJSON. I can't use the toString method on null or undefined values but an empty string or a zero should be valid,

src/components/StylesFromLiteralsService.js
loicgasser: prop is the value not the key

src/components/StylesFromLiteralsService.js
loicgasser: yes

src/components/StylesFromLiteralsService.js
oterral: ok

src/components/StylesFromLiteralsService.js
oterral: Whatever prop is,  replace will work fine without the `toString`

loicgasser: yes but I'd rather have the template value instead.

src/components/StylesFromLiteralsService.js
loicgasser: In that case https://github.com/geoadmin/mf-geoadmin3/blob/99b4743709268a7619d44289ebec4c3fdd562ac1/test/specs/StylesFromLiterals.spec.js#L28

But you're right if in a style function feature should always be defined. I prefer it to be robust though.

oterral: Ok it's not so important.

loicgasser:
Bug on IE is fixed, apparently IE11 doesn't support em unit in canvas.

src/components/StylesFromLiteralsService.js
oterral: What I meant is the `replace` function does the `toString()` for you, so no need to use it.

loicgasser: yes ok

src/components/StylesFromLiteralsService.js
oterral: May be I see when it could be useful to transofrm the value before the further test.
Are you able to provide a test when a propety value is 0 ? This is typically this kind of thing that we forget  in a refactoring.

loicgasser:
test added, thanks for the review, let me know if you want me to adapt anything else.

loicgasser:
Test added for property and 0

oterral:
thx for your patience

PR #4093, opened 2017-12-21
Log the styles error

oterral:
The alert message is displayed on every rendring frame, it is very annoying. Log the errors is enough


loicgasser:
yes but not as obvious.
errors could easily go undetected. I agree it's annoying though. The real fix would be to validate the JSON schema validator before allowing pushes.
https://github.com/tdegrunt/jsonschema
Code LGTM

oterral:
The info is only useful for devs not for the user.

loicgasser:
yes

oterral:
Thx for the review

Issue #4094, opened 2017-12-22
Testing custom stylesheets JSON


davidoesch:
To test different JSON styles

https://map.geo.admin.ch?style_url=https://pathtomystyle.json&layers=ch.meteoschweiz.messdaten-lufttemperatur

This could be part of the accpetance infra (Int) but not production and would enable the customer to fiddel around with his json easiert than in https://codepen.io/geoadmin/pen/ZWzXgv?editors=0010 

cc @loicgasser @mariokeusen

gjn:
+1 from me.

We will need something similar for vector tiles as well, once we are there. We need to enable people to easily style their stuff themselfes.

We also e.g. use case for data presentations per  commune or canton, etc.

oterral:
Fixed

PR #4096, opened 2018-01-04
Add layers_styleurl support

loicgasser:
Add support for https://github.com/geoadmin/mf-geoadmin3/issues/4094

[Test Link](https://mf-chsdi3.dev.bgdi.ch/meteo_temp/shorten/77ed5bcd7f)

TODO:

- [x] Add tests
- [x] Proxify style url


src/components/map/LayersService.js
```diff
@@ -643,6 +648,8 @@ goog.require('ga_urlutils_service');
             olLayer.timestamps = config.timestamps;
             olLayer.geojsonUrl = config.geojsonUrl;
             olLayer.updateDelay = config.updateDelay;
+            olLayer.externalStyleUrl = opts && opts.externalStyleUrl ?
+                opts.externalStyleUrl : null;
```
gjn: can't we take `styleUrl` here?

gjn:
Because this potentially uses external style, we should mark the layer accordingly somehow in the layermanager and (possibly) in the attribution on the map.

We have this for imported layers (external WMS or KML) where we put the text in red and add a red icon in the layermanager.

It should be clear to the one looking at it that style is not coming from official site.

davidoesch:
I Love it-- will be a game changer indeed

to do:

- [ ] Dokument in https://help.geo.admin.ch/?id=54
- [ ] External Style warning : Same as for external Data, If an external styling for GeoJSON is in applied (overrriding the official from dataowner signed style) the treat it in the same way as we do all imported KML WMS WMTS layers with adding two icons
image

- [ ] Change mouseover third party from : "Dataset provided by third party " to "Dataset and/or style provided by third party"

- [ ] Change Warning Text on click from "Warning: Third party data shown (www.mapplus.ch).  Availability is ensured by third party data provider. The terms and conditions of the third party data owner do apply and have to be respected."  to "Warning: Third party data and /or styling shown (URL).  Availability is ensured by third party data provider. The terms and conditions of the third party data owner do apply and have to be respected."

loicgasser:
> Dokument in https://help.geo.admin.ch/?id=54

I think we should not document the parameter for now since styling specs may shift to mapbox flavours.

> External Style warning

yes good point

> Change mouseover third party 

ok

> Change Warning Text 

ok

src/components/map/LayersService.js
loicgasser: As we already have a `stlyeUrl` property on the geojson layers at the layers configuration level, I found it confusing to use the same word for native and external styles.  In theory it would be possible to switch between the native style and the external one, with this approach we can still do that.

test/specs/map/Layers.spec.js
```diff
@@ -845,10 +866,12 @@ describe('ga_layers_service', function() {
           var source = layer.getSource();
           expect(source instanceof ol.source.Vector).to.be.ok();
           expectCommonProperties(layer, 'geojson');
-          gaLayers.getLayerPromise('geojson').then(function(feats) {
-            expect(feats[0].getGeometry().getCoordinates()).to.eql([2600000, 1199999.999848965]);
+          $q.all([gaLayers.getLayerPromise('geojson'),
+            gaLayers.getLayerStylePromise('geojson')]).then(function(vals) {
+            expect(vals[0][0].getGeometry().getCoordinates()).to.eql([2600000, 1199999.999848965]);
```
oterral: maybe you could test the layer's style used too. Be sure it's the one we just load.

src/components/layermanager/partials/layermanager.html
```diff
@@ -2,7 +2,7 @@
   <li ng-repeat="layer in filteredLayers" class="ga-layer-folded"
       ng-class="{
         'ga-layer-time-enabled': layer.timeEnabled,
-        'ga-layer-warning': layer.useThirdPartyData
+        'ga-layer-warning': layer.useThirdPartyData || layer.externalStyleUrl
```
oterral: Set the `useThirdPartyData` in the gaLayers Service just after you set `externalStyleUrl`

test/specs/map/Layers.spec.js
loicgasser: test added

gjn:
What about the externalStyleUrl. I agree with @oterral that it's confusing. Why not simply overwrite existing styleUrl property?

loicgasser:
> As we already have a stlyeUrl property on the geojson layers at the layers configuration level, I found it confusing to use the same word for native and external styles. In theory it would be possible to switch between the native style and the external one, with this approach we can still do that.

> styleUrl is part of the layersConfig -> denote that it is not a layerSpec

If you overwrite the stlyeUrl from the layersConfig when you remove and add a layer loaded with an external style. you will always have the external one and I don't think that this is what we want.

In the issue we say that it is just a testing feature for external people, so no permanent change of the app should be applied if that's what you suggest.

loicgasser:
@oterral can I merge?

Issue #4095, opened 2018-01-04
Search results do not close when deleting (fast) input text with keyboard


loicgasser:
1. enter aba in input search
2. delete with back key fast until input text is empty

Result:
search results are still displayed

Expected Result:
search results are not displayed anymore

oterral:
Fixed

PR #4097, opened 2018-01-05
Remove search result when input empty

nadine-piveteau:
Related to issue #4095

### DIFF
--- a/src/components/search/partials/search.html
+++ b/src/components/search/partials/search.html
@@ -12,9 +12,8 @@
     <button class="fa fa-remove-sign ga-icon ga-btn" ng-click="clearInput()"
        ng-mousedown="preClear($event)" ng-hide="query==''"></button>
     <span class="ga-search-dropdown"
-          ng-show="restat.sets()"
-          ng-class="'ga-search-' + restat.sets()"
-          ng-if="!(query==''||query=='undefined')">
+          ng-show="query && restat.sets()"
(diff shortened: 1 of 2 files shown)


gjn:
Please always provide a testlink so people can test it.

Code looks good to me.

nadine-piveteau:
[test link](https://mf-chsdi3.int.bgdi.ch/shorten/78007ba47a)

try for instance to enter "Bahnhofstrasse" into the search bar and then remove fastly all letters with the backspace key. No search results should be displayed anymore.

src/components/search/partials/search.html
```diff
@@ -11,10 +11,10 @@
     <i class="fa fa-search"></i>
     <button class="fa fa-remove-sign ga-icon ga-btn" ng-click="clearInput()"
        ng-mousedown="preClear($event)" ng-hide="query==''"></button>
-
     <span class="ga-search-dropdown"
           ng-show="restat.sets()"
-          ng-class="'ga-search-' + restat.sets()">
+          ng-class="'ga-search-' + restat.sets()"
+          ng-if="!(query==''||query=='undefined')">
```
oterral: There is already a `ng-show`, you should use it: 
```
ng-show="query && restat.sets()"
```

nadine-piveteau: Ok thx a lot! I accordingly adapted the unittests. Test link is updated.

PR #4140, opened 2018-02-05
updated wms list

davidoesch:
Based on 
https://docs.google.com/spreadsheets/d/1BHAzi9__ZCgPYLpDmezyX3oy3AuwiPZUMnKrOTbjKv8/edit#gid=0&fvid=1913726399
fixed and removed a lot of wms

[Test link](https://mf-geoadmin3.int.bgdi.ch/[BRANCH_NAME]/index.html)
This link will work only if the build succeed


src/js/ImportController.js
```diff
@@ -53,15 +50,13 @@ goog.require('ga_wmts_service');
         'http://wms.geo.gr.ch/wms/amtlichevermessung_stand',
         'http://wms.geo.gr.ch/wms/wildruhezonen',
         'http://wms.geo.gr.ch/wms/wildschutzgebiete',
-        'http://wms.geo.gr.ch/wms/zonenplan',
         'http://www.sogis1.so.ch/wms/avwms',
         'http://www.sogis1.so.ch/wms/grundbuchplan',
         'http://www.sogis1.so.ch/wms/wms_lidar',
         'http://wms.vd.ch/public/services/wmsVD/Mapserver/Wmsserver',
-        'https://wms.geo.bs.ch/wmsBS',
+        'http://wms.geo.bs.ch/wmsBS?',
```
oterral: ? is not necessary

src/js/ImportController.js
```diff
@@ -164,7 +155,8 @@ goog.require('ga_wmts_service');
         // non-SwissProjected test urls
         'http://wms.ga.admin.ch/1GE',
         'http://wms.ga.admin.ch/LG_DE_Geologie_und_Tektonik/wms',
-        'http://discomap.eea.europa.eu/arcgis/services/Land/CLC2000_Dyna_WM/MapServer/WMSServer',
+        'http://bio.discomap.eea.europa.eu/arcgis/services/Ecosystem/Ecosystems/MapServer/WMSServer?request=GetCapabilities',
```
oterral: no need to write  "?request=GetCapabilities"

src/js/ImportController.js
```diff
@@ -164,7 +155,8 @@ goog.require('ga_wmts_service');
...
+        'http://copernicus.discomap.eea.europa.eu/arcgis/services/Corine/CLC2012_WM/MapServer/WMSServer?request=GetCapabilities'
```
oterral: Here too

src/js/ImportController.js
davidoesch: it _IS_ necessary ..othrwise it wont load in the pulldown....see the curretn implementation

src/js/ImportController.js
davidoesch: agree

src/js/ImportController.js
davidoesch: agree

src/js/ImportController.js
oterral: I've just tested on prod it works without. I think the WMS is just very  unstable.

davidoesch: according to OWSLib this one works always with ? ....

PR #4205, opened 2018-03-16
Browser print

oterral:
This PR simplifies the print with browser command (ctrl+P). It prints what's it displays (map, copyright, tooltip, profile)

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/teo_print/index.html)</jenkins>


procrastinatio:
Very good. The only issue I see, is when the browser height/width ratio doesn't match the print ratio (landscape/portrait), the map is printed on two pages. PR https://github.com/geoadmin/mf-geoadmin3/pull/4228 is addressing this issue.

PR #4240, opened 2018-03-29
Update map libs

oterral:
The tileset was just not up to date

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/teo_up/index.html)</jenkins>


oterral:
Name are there now.

PR #4248, opened 2018-04-04
Fix jenkins autodeploy on int

oterral:
Fix #100

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/teo_make/index.html)</jenkins>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,6 +155,7 @@ MAKO_LAST_VARIABLES_PROD = ${MAKO_LAST_VARIABLES} \
 	    .build-artefacts/last-version
 
 
+
 .PHONY: help
 help:
 	@echo "Usage: make <target>"
@@ -224,15 +225,20 @@ help:
 	@echo "- DEPLOY_TARGET               (current value: ${DEPLOY_TARGET})"
 	@echo
 
+showVariables:
+	@echo "DEPLOY_TARGET = $(DEPLOY_TARGET)"
+	@echo "S3_BASE_PATH = $(S3_BASE_PATH)"
+	@echo "S3_SRC_BASE_PATH = $(S3_SRC_BASE_PATH)"
+
 .PHONY: all
-all: lint debug release apache testdebug testrelease fixrights
+all: showVariables lint debug release apache testdebug testrelease fixrights
 
 .PHONY: user
 user:
 	source $(USER_SOURCE) && make all
 
 .PHONY: build
-build: .build-artefacts/devlibs .build-artefacts/requirements.timestamp $(SRC_JS_FILES) debug release
+build: showVariables .build-artefacts/devlibs .build-artefacts/requirements.timestamp $(SRC_JS_FILES) debug release
 
 .PHONY: dev
 dev:
@@ -267,6 +273,7 @@ release: .build-artefacts/devlibs \
 	prd/info.json \
 	prd/robots.txt \
 	prd/robots_prod.txt
+	$(call showVariables)
 
 .PHONY: debug
 debug: .build-artefacts/devlibs src/deps.js src/style/app.css src/index.html src/mobile.html src/embed.html src/404.html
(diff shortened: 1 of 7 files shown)


scripts/clonebuild.sh
```diff
@@ -29,39 +24,25 @@ DEPLOY_GIT_BRANCH=${@:$OPTIND+2:1}
 DEEP_CLEAN=${@:$OPTIND+3:1}
 NAMED_BRANCH=${@:$OPTIND+4:1}
 
+# Remove the clone folder
+if [ $DEEP_CLEAN = true ]; then
+  make cleanall
+fi
```
loicgasser: we want to `cleanall` in the snapshot not locally.

scripts/clonebuild.sh
```diff
@@ -29,39 +24,25 @@ DEPLOY_GIT_BRANCH=${@:$OPTIND+2:1}
...
-  export KEEP_VERSION="false"
-  make .build-artefacts/last-version
-  S3_BASE_PATH=$(get_s3_basepath $DEPLOY_GIT_BRANCH)
-fi
-S3_SRC_BASE_PATH=$S3_BASE_PATH"src/"
-
-echo "S3_BASE_PATH:"
-echo $S3_BASE_PATH
-echo "S3_SRC_BASE_PATH:"
-echo $S3_SRC_BASE_PATH
```
loicgasser: I think you could still display these. It's useful to know what's going on.

Makefile
```diff
@@ -397,7 +409,7 @@ flushvarnish: guard-DEPLOY_TARGET
 
 # This internal target has been created to have the good global variable values
 # from rc_XXX file.
-flushvarnishinternal: guard-VARNISH_HOSTS guard-API_URL guard-E2E_TARGETURL
+flushvarnishinternal: guard-API_URL guard-E2E_TARGETURL
```
loicgasser: why remove `guard-VARNISH_HOSTS`?

scripts/clonebuild.sh
oterral: `cleanall` remove the entire snapshot

scripts/clonebuild.sh
oterral: They are set in the makefile. I'm not sure they are accessible outside

Makefile
oterral: rc_ dev hasn't VARNISH_HOSTS

loicgasser: You can use an empty list for dev

scripts/clonebuild.sh
loicgasser: Ok

scripts/clonebuild.sh
loicgasser: Ok I hadn t see that. But still you are in the current directory and your also removing your built artefacts. Just remove the clonedir instead.

scripts/clonebuild.sh
oterral: We can display it when we execute a  Make target.

scripts/clonebuild.sh
oterral: yes good idea

Makefile
oterral: ok

scripts/clonebuild.sh
oterral: done

scripts/clonebuild.sh
oterral: I've added a showVariable target in make . It is used on each beginning of build or deploy targets

Makefile
oterral: I have still keep the guard-VARNISH_HOSTS but I have mprove the log message when VARNISH_HOST is empty

oterral: because "export VARNISH_HOSTS=" in rc_dev was detected as a failure

loicgasser: I was thinking `export VARNISH_HOSTS=()` but you changed that and it would be `export VARNISH_HOSTS=""`

LGTM

oterral: it didn't work with VARNISH_HOSTS="", I've alaso tried with a () but didn't work too. You can try if you want.

loicgasser:
wrong order of merge I'll fix

PR #4294, opened 2018-04-25
Adding Overview Layer with WMTS LV95 Pyramid

gjn:
Did the Update of my Windows machine, which gave me time to code. This is mainly for debugging purposes. It adds a layer to the application which shows the wmts pyramid.

Still needs cleaning up and
[ ] Make it only visible with debug=true parameter
[ ] Add other pyramids (web mercator, wgs84, etc)

### DIFF
--- a/src/components/map/GridLayersService.js
+++ b/src/components/map/GridLayersService.js
@@ -1,0 +1,137 @@
+goog.provide('ga_gridlayers_service');
+
+goog.require('ga_definepropertiesforlayer_service');
+goog.require('ga_layerfilters_service');
+goog.require('ga_maputils_service');
+goog.require('ga_permalink_service');
+goog.require('ga_tilegrid_service');
+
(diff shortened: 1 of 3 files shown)


oterral:
It seems similar to [this ol example](http://openlayers.org/en/master/examples/canvas-tiles.html)

You could use this ['ol.source.TileDebug'](http://openlayers.org/en/master/apidoc/ol.source.TileDebug.html)

oterral:
Maybe we could add automatically the grid for each layer using a grid. What do you think?

procrastinatio:
OK nice idea.
But tiles addreses are wrong. For instance, this one (Geneva) is labeled as `16/1/4`:
https://tod100.int.bgdi.ch/1.0.0/ch.swisstopo.swissimage/default/current/2056/16/1/3.jpeg

PR #4304, opened 2018-04-30
by default use pypi.org

danduk82:

@oterral could you please try to deploy with this fix?

Apparently now pypi.org is working again. We should use it as default. With the use lf PYPI_URL environment variable we are still able to overrwrite it if needed in our build environment. By default we should keep pypi.org in the Makefile...

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/ltbon_fix_pypi_mess/index.html)</jenkins>


oterral:
jenkins is currently deploying on int

danduk82:
seems good to me @oterral  what do you think?

PR #4343, opened 2018-05-22
Fix #4109: Don't update the profile when the popup is reduced

### LINKED ISSUE #4084: 3D in iframe: show arrows
when using 3D in iframe, eg https://bellevue.nzz.ch/reisen-entdecken/das-pro-natura-zentrum-am-neuenburgersee-im-feuchtgebiet-unterwegs-ld.1339442
the buttons are useless on desktop, since user can only reset but not rotate
image

proposal:
add as well arrows as on desktop for roatating the view ---  details to be discussed

### LINKED ISSUE #4109: Do not trigger profile requests when window is minimized
Step to reproduce:

1. open map.geo.admin.ch
2. draw a small line
3. minimize the profile
4. modify the line

Result:

A profile request is still sent to the server

Expected Result:

No profile request is sent when the profile is reduced, only when the window is reopened

oterral:
Fix #4109 

Draw a profile -> reduce the popup -> modify draw -> No request to the profile service is done

Fix #4084 also [test](https://codepen.io/oterral/pen/dmMrmM)

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/fix_4109/index.html)</jenkins>


procrastinatio:
Tilt arrow: don't see any differences https://codepen.io/procrastinatio/full/rKNLXm/

PR #4352, opened 2018-05-25
Hiking time

oterral:
This PR uses the same formala as schweizmobil.

It seems the last value in the constants array was wrong.

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/teo_ht/index.html)</jenkins>


procrastinatio:
Nice.

procrastinatio:
Check:
Took the very same section as Schweizmobil, the exact same LineString but did gave every point an altitude with the profile service and compute some hiking stats (see KML popup).
Fields like `HoeheAbE` or `Laenge` are thoose give by Schweimobile
Fields like `Laenge_`, ending with an `_` are thoose calculated on the basis of data from profile service.
Fields ending with `r` is the ratio of both fields above.

e.g. _Laupen-Bern_

Given `LaengeE`        21783.16
Caclulated `Laenge_ `       25397.1668608
Ratio: `Laenger`    0.857700393095
Looks like the given length is wrong...

Grey:  within 5% of hiking time from Schweizmobil reference section
Red: more than 5% difference. Some section are outside Switzerland, so no altitude. Some section length do not match, some up/down also not match

reference-sections

[Demo (slow, loading 100MB of KML)](https://s.geo.admin.ch/7acce62709)

davidoesch:
root cause why "Some section length do not match"? I do know that chmobil fixes on their side some bridges which are not included in TLM . but we also have effects where there ar no major bridges lik on this route https://s.geo.admin.ch/7ad231d34c

src/components/profile/ProfileService.js
```diff
@@ -198,44 +198,52 @@ goog.require('ga_urlutils_service');
...
-                  } else {
-                    minutesPerKilometer = (-9 * s);
-                  }
+              var data = this.data[i];
+              var dataBefore = this.data[i - 1];
+
+              // Distance betwen 2 points
+              var distance = data.dist - dataBefore.dist;
+
+              if (!distance) {
```
procrastinatio: Are you sure? You may have a vertical difference with absolutely no distance (via ferrata)

PR #4397, opened 2018-06-21
Using AWS instance role instead of profile

procrastinatio:
...instead of profile.

Try with somthing lile:
  ` make s3deploybranchint`

If you try to deploy an older snapshot, with e.g. `SNAPSHOT=201806211752 make s3deployint
` it won't work as it uses the _s3manage.py_ withih the snapshot.

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/mom_instance_role/index.html)</jenkins>

### DIFF
--- a/scripts/s3manage.py
+++ b/scripts/s3manage.py
@@ -491,7 +491,7 @@ def main():
     print(parse_arguments(sys.argv))
     cmd_type, deploy_target, base_dir, named_branch, git_branch, bucket_name, s3_path = \
         parse_arguments(sys.argv)
-    for var in ('AWS_PROFILE', 'AWS_ACCESS_KEY_ID', 'AWS_ACCESS_KEY_ID'):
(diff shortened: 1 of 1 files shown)


scripts/s3manage.py
```diff
@@ -485,21 +481,22 @@ def parse_arguments(argv):
...
 
 
 def main():
     global s3, s3client, bucket
     print(parse_arguments(sys.argv))
-    cmd_type, deploy_target, base_dir, named_branch, git_branch, bucket_name, s3_path, profile_name = \
+    cmd_type, deploy_target, base_dir, named_branch, git_branch, bucket_name, s3_path = \
         parse_arguments(sys.argv)
-    s3, s3client, bucket = init_connection(bucket_name, profile_name)
+    for var in ('AWS_PROFILE', 'AWS_ACCESS_KEY_ID', 'AWS_ACCESS_KEY_ID'):
```
oterral: Duplication of id AWS_ACCESS_KEY_ID

PR #4414, opened 2018-06-27
[WIP] Pegman is here

oterral:
THIS PER IS CONSOLIDATED IN https://github.com/geoadmin/mf-geoadmin3/pull/4466

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/teo_peg/index.html)</jenkins>


davidoesch:
a pity that help opens a new tab and 3d actually exits. but IMHO better than having all the control explainations always on screen.

black pegman -- i like 

let me now as soon as you have all msg id( "drag me on a location")

oterral:
@davidoesch  ids:

3d_pegman_help
3d_pegman_tooltip

Yes too bad for the help in another window. We could display everything in the help on bottom.

PR #4433, opened 2018-07-06
chore(package): update less to version 3.5.2

oterral:
Closes #4431

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/greenkeeper/less-3.5.2/index.html)</jenkins>


oterral:
This less version breaks this kind of [calculation](https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/backgroundselector/style/backgroundselector.less#L1079

Issue #4470, opened 2018-08-24
3D: in flight mode UP and DOWN should be applied to current orientation of the view


danduk82:
in flight mode when we press on arrow UP or DOWN, the orientation vector is tilted in the vertical direction. Instead it should be applied to the current local rotation axis. I.e. if the plane is rotated, UP should allow for a sharper turn, as in a real plane.

```
vx = V * cos(horizonzalAngle)
vy = V * sin(horizonzalAngle)
```
where V is the delta defined when pressing UP or DOWN

AFoletti:
Can't I give a +100 directly? It will be on par on how many times I raised the issue (ok, not officially here in a ticket, but anyways)

davidoesch:
Afaik : This Is a cesium feature ....

fredj:
See https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/controls3d/fps.js#L360-L402

danduk82:
indeed we shoud do something like:
```
if (this.buttons_.forward) {
    pitch -= 0.02 * sin(roll);
    heading -= 0.02 * cos(roll);
}
etc
```
instead of simply updating the pitch

danduk82:
fixed [here](https://github.com/geoadmin/mf-geoadmin3/pull/4476). Thanks @fredj for finding the src file :+1:

PR #4476, opened 2018-08-29
fix flight mode UP and DOWN arrows

danduk82:
Using UP and DOWN now is applied to the both heading and pitch, which seems a bit more realistic.

It is still a bit awkward because the center of the focus is slightly moving, but this is another issue which occurs also with the "normal" 3D view.

this fixes [this issue](https://github.com/geoadmin/mf-geoadmin3/issues/4470)

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/ltbon_fix_flight_mode/index.html)</jenkins>


davidoesch:
CHAMPAGNE!!!!

PR #4497, opened 2018-09-18
Update ImportController.js with swissdatacube WMTS

davidoesch:
added swissdatacube WMTS

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/ltoed_WMTS_swissdatacube/index.html)</jenkins>


davidoesch:
You will love that link https://s.geo.admin.ch/7d17d2b0ce , marc

procrastinatio:
@gjn Less snow is kind of sad

procrastinatio:
Would it not be better to have meaningfull WMS descriptions instead of url?

davidoesch:
Fully agree Marc- part of RE4 There is even a spec for it. The idea would be to have them description directly fed from geocat

Issue #4478, opened 2018-08-29
IE11: printprogress only use cache


procrastinatio:
Print fails because it uses a cached version of the print progress, i.e. _https://print.geo.admin.ch/printprogress?id=180829124648276_8)

davidoesch:
Shall i Test it on IE 11?

davidoesch:
works on IE11 here .. how to reproduce?

PR #4517, opened 2018-10-17
Do not cache print progress requests 

procrastinatio:
Open link in IE9, IE10 or IE11 and try to print a length job (failry easy on int). The print progress should carry on.

See https://github.com/geoadmin/mf-geoadmin3/issues/4478, https://github.com/geoadmin/mf-geoadmin3/issues/4535

[Test link](https://mf-geoadmin3.int.bgdi.ch/mom_fix4478/index.html)

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/mom_fix4478/index.html)</jenkins>

### DIFF
--- a/src/components/tooltip/TooltipDirective.js
+++ b/src/components/tooltip/TooltipDirective.js
@@ -473,10 +473,10 @@ goog.require('ga_window_service');
                 }
                 var url = layerToQuery.getSource().getGetFeatureInfoUrl(
                     coordinate, mapRes, mapProj,
-                      {'FEATURE_COUNT': 10,
-                       'INFO_FORMAT': 'text/plain',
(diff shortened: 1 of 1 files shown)


src/components/print/PrintDirective.js
```diff
@@ -174,6 +174,7 @@ goog.require('ga_urlutils_service');
         }
         canceller = $q.defer();
         $http.get(noCacheUrl, {
+          cache: false,
```
oterral: Is it the default value, no?  There is a hack 2 line before for ie 9, should work for ie11

procrastinatio: With jquery 3 you don't need the hack anymore (untested) and may add the `cache: false`.

PR #4524, opened 2018-10-22
[MVT] Mb style service

loicgasser:

Here is the idea we discussed @oterral 

We should probably break it down further. The sources creation via TileJSON (source url) should have its own service. Maybe `gaMapUtils.olSourceFromTileJSONUrl`. What do you think?
I already added the filtering capabilities.

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/mb_style_service/index.html)</jenkins>

### DIFF
--- a/src/components/GLStyleService.js
+++ b/src/components/GLStyleService.js
@@ -1,0 +1,102 @@
+goog.provide('ga_gl_style_service');
+(function() {
+  var module = angular.module('ga_gl_style_service', []);
+
+  module.provider('gaGLStyle', function() {
+    this.$get = function($http, $q, $window) {
+      var GLStyle = function() {
+        this.styleJSONCache = null;
+
+        this.spriteJSONCache = null;
+
+        this.filters = [];
+
+        this.edits = [];
+
+        this.loadStyle = function(styleUrl) {
+          var that = this;
+          var defer = $q.defer();
+          $http.
+              get(styleUrl, {
+                cache: true
+              }).
+              then(
+                  function(response) {
+                    that.styleJSONCache = response.data;
+                    // load sprite
+                    var spriteUrl = that.styleJSONCache.sprite + '.json';
+                    $http.
+                        get(spriteUrl, {
+                          cache: true
+                        }).
+                        then(
+                            function() {
+                              that.spriteJSONCache = response.data;
+                              defer.resolve({
+                                styleJSON: that.styleJSONCache,
+                                spriteJSON: that.spriteJSONCache
+                              });
+                            },
+                            function(reason) {
+                              that.spriteJSON = null;
+                              $window.console.error(
+                                  'Unable to load ' +
+                        spriteUrl +
+                        ' because ' +
+                        reason
+                              );
+                              // failing to load sprite is ok for now...
+                              defer.resolve({
+                                styleJSON: that.styleJSONCache,
+                                spriteJSON: that.spriteJSONCache
+                              });
+                            }
+                        );
+                  },
+                  function(reason) {
+                    that.styleJSONCache = null;
+                    that.spriteJSONCache = null;
+                    var msg =
+                  'Unable to load the style from ' +
+                  styleUrl +
+                  ' because ' +
+                  reason;
+                    $window.console.error(msg);
+                    defer.reject(msg);
+                  }
+              );
+          return defer;
+        };
+
+        this.cloneStyle = function() {
+          var newStyleJSON = {};
+          angular.copy(this.styleJSONCache, newStyleJSON);
+          return newStyleJSON;
+        };
+
+        // Hide / show layers base on layers props
+        this.filter = function(filters) {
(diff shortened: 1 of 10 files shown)


loicgasser:
jenkins test

src/components/GLStyleService.js
```diff
@@ -0,0 +1,122 @@
...
+        this.get = function(styleUrl) {
+          var that = this;
+          var defer = $q.defer();
+          $http.
+              get(styleUrl, {
+                cache: true
+              }).
+              then(
+                  function(response) {
+                    that.styleJSONCache = response.data;
```
oterral: So we keep only the one glStyle in cache ?  when we change the bg layer it will be replaced, am I right?

src/components/GLStyleService.js
```diff
@@ -0,0 +1,122 @@
+goog.provide('ga_gl_style_service');
+(function() {
+  var module = angular.module('ga_gl_style_service', []);
+
+  module.provider('gaGLStyle', function() {
+    this.$get = function($http, $q, $window) {
+      var GLStyle = function() {
+        this.styleJSONCache = null;
```
oterral: the name is a bit confusing can we just name it  'glStyle'   so we knew it's the complete glStyle  in a js object

src/components/GLStyleService.js
```diff
@@ -0,0 +1,122 @@
...
+          $http.
+              get(styleUrl, {
+                cache: true
+              }).
+              then(
+                  function(response) {
+                    that.styleJSONCache = response.data;
+                    // load sprite
+                    var spriteUrl = that.styleJSONCache.sprite + '.json';
+                    $http.
```
oterral: load the sprite in antoher function it will be more clear.

src/components/GLStyleService.js
```diff
@@ -0,0 +1,105 @@
...
+              });
+            }, function(reason) {
+              that.spriteJSON = null;
+              $window.console.error(
+                'Unable to load ' + spriteUrl + ' because ' + reason
+              );
+              // failing to load sprite is ok for now...
+              defer.resolve({
+                styleJSON: that.styleJSONCache,
+                spriteJSON: that.spriteJSONCache
```
oterral: You can use the next ` then`  to avoid duplicate resolvation, see https://github.com/geoadmin/mf-geoadmin3/blob/mvt/src/components/map/LayersService.js#L798

src/components/GLStyleService.js
```diff
@@ -0,0 +1,105 @@
...
+            });
+            if (addLayer) layers.push(layer);
+          });
+          newStyleJSON.layers = layers;
+          return newStyleJSON;
+        };
+
+        this.edit = function(edits) {
+          var newStyleJSON = this.cloneStyle();
+          this.edits = this.edits.concat(edits);
```
oterral: What is `edits` properties ?

src/components/map/LayersService.js
```diff
@@ -775,50 +778,21 @@ goog.require('ga_urlutils_service');
...
+              gaGLStyle.get(config.styleUrl).then((data) => {
+                var glStyle = data.styleJSON;
+                var spriteData = data.spriteJSON;
+                var spriteUrl = glStyle.sprite + '.png';
+                $window.olms.stylefunction(olLayer, glStyle,
+                    config.sourceId,
+                    undefined, spriteData, spriteUrl,
+                    ['Helvetica']);
+                // HACK: Make the Swiss style transparent to see the relief
+                glStyle.layers.forEach(function(style, idx) {
```
oterral: We need to check if we still need this Hack

src/components/map/PermalinkFeaturesService.js
```diff
@@ -25,6 +25,12 @@ goog.require('ga_previewfeatures_service');
           var featureIdsByBodId = {};
           var paramKey;
           var listenerKey;
+          var forceZoom;
+
+          if (queryParams.zoom !== undefined && isFinite(queryParams.zoom)) {
+            forceZoom = parseInt(queryParams.zoom);
```
oterral: Why do we need to force the zoom ?

src/components/GLStyleService.js
loicgasser: Edit paint or layout props of a layer, based on a prop name and prop value.
Since we don't have data driven use cases comparators will be the same as for filters. We extend this logic as needed.

src/components/map/LayersService.js
loicgasser: I tried to remove it and didn t see any difference. But provided that we don't really have a swiss use case... I left it just in case.

src/components/map/PermalinkFeaturesService.js
loicgasser: That is because I rebased on master and then mvt. Already in master

loicgasser:
@oterral I have addressed your comments thanks for the review

PR #4550, opened 2018-11-02
[MVT] Advanced styling

oterral:
add the posbility to change
If font
- [ ]  font size  
- [ ]  font width
if line
- [ ]  line width
 
see below the examples from mapbox studio *for inspiration only* (if you need to see the esri way https://www.youtube.com/watch?v=RqGTxYWgmC8"
close (even hide if possible) the feedback window since user could now edit style in the simple and advanced feature mode

font manipulation
image

line /Area manipulation
image

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/teo_advanced/index.html)</jenkins>


src/components/backgroundselector/BackgroundService.js
```diff
@@ -116,7 +124,7 @@ goog.require('ga_urlutils_service');
         } */
       };
 
-      function updateBgLayerStyleUrlParam(bgLayer) {
+      var updateBgLayerStyleUrlParam = function(bgLayer) {
         var styleUrlValue = bgLayer.externalStyleUrl;
         if (bgLayer.id) {
           getBgById(bgLayer.id).styleUrl = styleUrlValue;
```
loicgasser: we should avoid having to deal with `externalStyleUrl` + `styleUrl` at the moment it's a **HUGE** mess in the background directive.
where is the source of truth? We don't know anymore, the layer in the map? the properties in the dict `vtBgs` that we extend on the fly?
It's getting really ugly...

loicgasser: In the directive I am relying on the styleUrl which was used the first approach.
Try to change the background with a custom style and everything goes to hell.

src/components/FileStorageService.js
```diff
@@ -11,19 +11,29 @@ goog.provide('ga_filestorage_service');
...
+        var url = fileStorageUrl;
+        if (custom) {
+          url = url.replace(/files$/, custom);
+        }
+        return url + ((id) ? '/' + id : '');
       };
 
-      var getPublicUrl = function(fileId) {
-        return publicUrl + ((fileId) ? '/' + fileId : '');
+      var getPublicUrl = function(fileId, custom) {
```
loicgasser: I would pass custom once in the constructor of `FileStorage` and make these methods private so that you have access to the basePath.

src/components/GlStyleService.js
```diff
@@ -34,49 +28,31 @@ goog.provide('ga_glstyle_service');
...
-        this.getSprite = function(spriteUrl) {
-          var that = this;
+        this.getSpriteData = function(spriteUrl) {
           return $http.get(spriteUrl, { cache: true }).then(function(res) {
-            that.spriteCache_ = res.data;
-            return {
-              style: that.styleCache_,
-              sprite: that.spriteCache_
-            };
+            return res.data;
```
loicgasser: why not but, why remove the spriteCache?

davidoesch:
add the posbility to change
If font
- [ ]  font size  
- [ ]  font width
if line
- [ ]  line width

src/components/backgroundselector/BackgroundService.js
loicgasser: I gave it some thought, i think we could handle the style in gaGLStyle on a per layer basis, otherwise we won't be able to store mutiple map states. I propose that we also store the source url of the style if it differs from the original one (or even all the time). The problem with externalStyleUrl is that it is not persistent, each time we switch the background we loose this information.
What do you think?

src/components/GlStyleService.js
oterral: because:
 - we never use it. 
 - it 's already cache by the $http service.
 - add a complexity in gaGlStyle. 

gaGlStyle returns a raw glStyle object, it's more simple and clear.

src/components/FileStorageService.js
oterral: As you said on friday I can manage depending on content type and as I said using a 'gl-styles/dfgfgdfg'  as fileId.

oterral: I will do 2 services is more clean

src/components/GlStyleService.js
loicgasser: but i can leave with it if you prefer

oterral: the sprite is cached by $http, so you  make request only once. and you always get the style you then the sprite. It was the case before this too.

oterral:
@loicgasser ready to review/merge

oterral:
@davidoesch font size, font width, line width  . I will do this in another PR

loicgasser:
for testing, some small bugs:

the default background doesn't show in the feedback panel anymore on startup.
image

the blue border of the color picker is awkward.

image

thirdparty logo doesn't show on the background selector anymore
only at the layer level.
image

src/components/edit/Color.js
```diff
@@ -0,0 +1,55 @@
...
+
+        var colorInput = $('<input type="color" value="!" />')[0];
+        scope.useInputColorSelector = (colorInput.type === 'color' &&
+            colorInput.value !== '!');
+
+        scope.colors = [
+          { value: 'lightgray', label: 'light_gray' },
+          { value: '#acc864', label: 'light_green' },
+          { value: '#3a8841', label: 'green' },
+          { value: '#40b5bc', label: 'light blue' },
```
loicgasser: missing underscore in label

src/components/edit/EditGlStyleDirective.js
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * This directive add an interface where you can modify a glStyle.
+   */
+  module.directive('gaEditGlStyle', function(gaLayers, $rootScope) {
+    return {
+      restrict: 'A',
+      templateUrl: 'components/edit/partials/edit-glstyle.html',
+      scope: {
+        glStyle: '=gaEditGlStyle',
+        config: '=gaEditGlStyleConfig',
```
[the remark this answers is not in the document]
oterral: elle vient de layersConfig

src/components/backgroundselector/BackgroundService.js
```diff
@@ -116,46 +124,62 @@ goog.require('ga_urlutils_service');
...
+            deregFns.push(scope.$watch(function() {
+              return layer.externalStyleUrl;
+            }, function() {
+              updateBgLayerStyleUrlParam(layer);
+            }));
+          });
+        });
+      };
+
+      var createOlLayer = function(map, bg) {
```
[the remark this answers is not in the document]
oterral: yes it's multifunctionnal :)

src/components/map/LayersService.js
```diff
@@ -337,7 +338,35 @@ goog.require('ga_urlutils_service');
...
+                      ['layout', 'text-size', '{size}'],
+                      ['layout', 'text-transform', '{transform}'],
+                      ['layout', 'text-letter-spacing', '{spacing}'],
+                      ['layout', 'text-rotation-alignment', '{font}'] */
+                    ],
+                    'place_label_city': [
+                      ['paint', 'text-color', '{color}'],
+                      ['paint', 'text-halo-color', '{color}']
+                    ]
+                  }
```
loicgasser: layersConfig to the rescue

oterral:
all fixed @loicgasser

PR #4553, opened 2018-11-05
Use ng style

loicgasser:

For compatibility issues with IE.

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/ngstyle/index.html)</jenkins>


loicgasser:
trivial self merge

Issue #4546, opened 2018-11-01
[MVT] logo Edit and mobile third party


davidoesch:
Edit logo

![image](https://user-images.githubusercontent.com/4577727/47838821-d5750a00-ddb0-11e8-9b60-a4fa0bae60d6.png)

as same size of for edit in https://mf-geoadmin3.int.bgdi.ch/third_party/index.html?lang=fr&topic=are&bgLayer=omt.vt&E=2568726.58&N=1142792.88&zoom=10&catalogNodes=954,959,965&bgLayer_styleUrl=https:%2F%2Frawgit.com%2Fopenmaptiles%2Fosm-bright-gl-style%2Fmaster%2Fstyle.json 

![image](https://user-images.githubusercontent.com/4577727/47838958-34d31a00-ddb1-11e8-88b6-be58ea87c3c9.png)

for mobile
- we need only external style warning

- no edit (@pauloamado will provide a logo with a more centered person)
![image](https://user-images.githubusercontent.com/4577727/47839020-68ae3f80-ddb1-11e8-96c1-00832d55bb2f.png)

loicgasser:
done on my side

pauloamado:
I propose to use the user-edit icon from Font Awesome:
![icon_edit_vector](https://user-images.githubusercontent.com/6584138/48251415-d84faa80-e401-11e8-8429-cd9a83bd92bf.png)

Here is the result on viewer selector (real size view):
![icon_edit_vector_real_size_view](https://user-images.githubusercontent.com/6584138/48251672-c7536900-e402-11e8-9f27-73f7686da349.png)

For the mobile logo (only style warning) go on using the user icon from Font Awesome, but bigger and centered:
![bg_switcher_vector_mobile](https://user-images.githubusercontent.com/6584138/48253066-b0af1100-e406-11e8-8148-0bd975b9e3bc.png)

davidoesch:
awesome
@loicgasser you can add them to the testviewer

pauloamado:
Here is the new fonts set, with icon "fa-user-edit"
[font-awesome-4.5.0.zip](https://github.com/geoadmin/mf-geoadmin3/files/2565846/font-awesome-4.5.0.zip)

loicgasser:
we should end up with 3 layer right? 
basemap
hybrid
raster color

It's maybe best to wait for actual data, to have the same look and feel.

davidoesch:
See 
[...]

PR #4554, opened 2018-11-05
[MVT] Query mvt feature F-007

loicgasser:

@davidoesch you can already test with openmaptiles. That is what I understood from the specifications.

You can already make suggestions if you want.
image

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/query_mvt_feature/index.html)</jenkins>


davidoesch:
+1  well done. the key/values are the depending on the delivery by Karto
Question: multilang- needs to be tested

oterral:
@davidoesch  Why not having a popup display an the map ?  I think it s really hard to understand and to find the functionnality as it is. I spend 5 minutes to find something with something displayed in the value field.
I think a popup like maputnik inspect mode is better.

davidoesch:
@oterral  Agree, expected a little bit more info as well. The goal is to show the potential of vector attributes.
Thing is: background maps don't have tooltips, this is why I proposed  the footer info. at the same time clicking on maps is a PITA - the idea is to have a continous update of the info what is below the mouse pointer (which works only on desktop anyway)
Proposal: if you can add  maputnik style inspect mode using our tooltip CI/CD: great  (https://cloud.githubusercontent.com/assets/1288339/21468906/21cdb3d6-ca2e-11e6-955e-0e83b096fdba.png

loicgasser:
I agree inspect mode would make more sense. I can easily adapt it. Should it be a toggle button in the footer?

src/components/queryvector/QueryVectorDirective.js
```diff
@@ -0,0 +1,74 @@
...
+
+(function() {
+  var module = angular.module('ga_query_vector_directive', [
+    'ga_layerfilters_service'
+  ]);
+
+  var registerPointerMove = function(scope, map) {
+    return map.on('pointermove', function(evt) {
+      var featureFound;
+      map.forEachFeatureAtPixel(evt.pixel, function(feature, layer) {
```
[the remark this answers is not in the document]
loicgasser: Thanks

loicgasser:
@davidoesch I switched to an inspect mode version. Do you like it better?

oterral:
small improvments:
 - seperate feature's result with css
 - alphabetical order of the properties

loicgasser:
@oterral I ll create a table with the results and sort alphabetically. I ll try to make it a little seixier as well.
@davidoesh yes no problem, I ll add it in the feedback panel, it makes sense.

loicgasser:
This is now a fork of https://github.com/geoadmin/mf-geoadmin3/pull/4563, merge this one first.

Fix for https://github.com/geoadmin/mf-geoadmin3/issues/4562
Inspect can now be activated in the feedback panel.
Keys will be translated when we have real data.

loicgasser:
image

davidoesch:
On mobile or does not fit the screen

IMHO: completely deactivate inspect on mobile. Remove it from panel if a touch device is activated

loicgasser:
Good Point i forgot to remove it on mobile

loicgasser:
also a fix for https://github.com/geoadmin/mf-geoadmin3/issues/4546

src/components/queryvector/style/queryvector.less
```diff
@@ -0,0 +1,62 @@
+.ga-query-vector-popup {
+  position: absolute;
+  background-color: white;
+  -webkit-filter: drop-shadow(0 1px 4px rgba(0, 0, 0, 0.2));
```
oterral: no needed , autoprefixer will add it

loicgasser: Ok

loicgasser:
* the popup stays open when you mouse is on another element (like the bg selector or outside the window. it should be closed.
* the popup keeps the information of the last feature if no feature found.

done and done, it does feel better, if no feature is found I hide the popup entirely. I could also display an empty popup, but it looks a bit weird.


* we should remove the non bg vt layer from the bg list in the feedback directive.

I deactivate all the vector features when a non VT layer is added to the map. I think this makes sense the way it is. removing it from the dropdown is easy though if @davidoesch prefers it.

oterral:
> I deactivate all the vector features when a non VT layer is added to the map. I think this makes sense the way it is. removing it from the dropdown is easy though if @davidoesch prefers it.

I know  but for me it's not enough obvious what the user must do to have the style modification back

oterral:
looks very good thx

davidoesch:
"I deactivate all the vector features when a non VT layer is added to the map. I think this makes sense the way it is. removing it from the dropdown is easy "
Makes sense . Agree it is hard to know which layers have feedback but I propose that we start test.map.geo.admin.ch with a vt layer . So user can reset the viewer and can start over

PR #4563, opened 2018-11-09
[MVT] Use editConfig and remove binding to glStyle

loicgasser:
With https://github.com/geoadmin/mf-geoadmin3/pull/4550 and the persistence of olLayer and glStyle we introduced a couple of bugs.

I also thought it made sense to use the `editConfig` to avoid duplicating data everywhere you just added @oterral.
I also added the labelsFilters to editConfig.

List of bugs it solves:

**A**

1. Open test viewer
2. Change background to Pixelkartefarbe
3. Reload the page

Result:
Pixelkartefarbe is not selected in the feedback panel

Expected Result:
Pixelkartefarbe is selected

**B**

1. Open test viewer
2. Select Pixelkartefarbe in the background selector

Result:
Feedback panel is not synchronized with background selector

Expected Result:
Feedback panel is synchronized and vector features are disabled

**C**

1. Open test viewer
2. Select a color in the feedback panel
3. Reload the page

Result:
Color is still visible

Expected Result:
Color is not visible anymore (e.g. style is not saved)

My understanding is that we don't make any permanent change in the feedback panel. I may be wrong.
I am not sure about C. Maybe, it would make more sense to synchronize the styles entirely.
Because if you edit the style again in the feedback panel after editing it in the advanced edit mode and you reload, well now you see the last style you edited in the advance mode....
Maybe when we re-open the advance mode we should simply re-apply the last style.

ping @davidoesch 

**D**

Disable edit on raster background.

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/feedback_refactor/index.html)</jenkins>

### DIFF
--- a/src/components/GlStyleService.js
+++ b/src/components/GlStyleService.js
@@ -84,7 +84,7 @@ goog.provide('ga_glstyle_service');
                 var editLayerProperty = layer[editPropertyName];
                 if (editLayerProperty &&
                     editLayerProperty === editPropertyValue) {
-                  var editStyle = edit[2].split('|');
+                  var editStyle = edit[2];
                   if (!layer[editStyle[0]]) {
                     layer[editStyle[0]] = {};
                   }
--- a/src/components/backgroundselector/BackgroundService.js
+++ b/src/components/backgroundselector/BackgroundService.js
@@ -35,14 +35,6 @@ goog.require('ga_glstylestorage_service');
           disable3d: true
           // labels: false
         },
-        /*
-        'ch.swisstopo.wandern.vt': {
-          id: 'ch.swisstopo.wandern.vt',
-          label: 'wandern',
-          disable3d: true,
-          labels: false
-        },
-        */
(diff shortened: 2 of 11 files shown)


loicgasser:
@oterral I fixed the tests, you can review

davidoesch:
Agree on c) this is how it should work... But your statement makes sense
-> when we re-open the advance mode we should simply re-apply the last style.

oterral:
Bug found : 
 - Open the test link
 - Modify the color
 - change the bg
 - go back to the openmaptiles bg
 - try to change the color

->>>> the vector layers disappear

loicgasser:
This was introduced with your PR... Already on MVT

loicgasser:
fixed in this PR as well now

davidoesch:
for C) @loicgasser 
"Maybe when we re-open the advance mode we should simply re-apply the last style"

loicgasser:
code here merged in https://github.com/geoadmin/mf-geoadmin3/pull/4554

PR #4566, opened 2018-11-12
[MVT] Add modification of text size and line width

### LINKED ISSUE #4558: [MVT] F-005 : add Line and Font edit
add the possbility to change
besides color

If font
- [x]   font size
- [x]   font width
if line
- [x]   line width

### LINKED ISSUE #4560: [MVT] F-005 : msgids for "We are modifying the layer omt.vt "
create msgid "We are modifying the layer omt.vt "

### LINKED ISSUE #4564: [MVT] Clicking on Style the map should trigger a gaToggleEdit event
Because we want to be able to 

1. Close the feedback panel when enabling the edit mode
2. Highlight the correct tool in the bg selector

both have listeners on this event. Alternatively this can be passed as an argument to the directive. (isEditActive)

I'll let you choose whatever solution you prefer.

oterral:
@davidoesch for the color should leave the input color as it is (probably not working on all browser) ? or should I use the same design as in the feedback directive (list of  colored squares) 

Fix #4558 
Fix #4560 
Fix #4564

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/teo_more_styles/index.html)</jenkins>


davidoesch:
@oterral: so far I tested it on FF EDGE CHROME on win10, the only Issue I see is IE Bund see below
IMHO: leave the input color as it is
image

PR #4580, opened 2018-11-14
[MVT] Fixes

loicgasser:

@oterral with this PR we keep the glStyle reference but we re-create a new ol instance of the layer each time. This solves the bug where labels disappear.  Also, now we load the style once and pass the gl config to the sublayers. And we don't attach the style in applyGLStyleToLayer, but simply in getOlLayerById which makes also more sense.

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/gal_fixes/index.html)</jenkins>


src/components/backgroundselector/BackgroundService.js
```diff
@@ -214,20 +218,15 @@ goog.require('ga_glstylestorage_service');
...
-            var newBg = getBgById(newBgId);
-            if (newBg) {
-              bg = newBg;
+            bg = getBgById(newBgId);
+            if (bg) {
               var layers = map.getLayers();
-              if (bg.id === 'voidLayer') {
-
+              if (bg.id === 'voidLayer' && layers.getLength() > 0 &&
+                  layers.item(0).background) {
```
oterral: here you change the logic , if there is no layers on the map , you go in the 'else' and so create a layer. You don't want that

src/components/map/LayersService.js
```diff
@@ -799,44 +806,43 @@ goog.require('ga_urlutils_service');
...
+              // Load informations from tileset.json file of a source
+              var sourceConfig = glStyle.sources[olLayer.sourceId];
+              sourceConfig.minZoom = config.minZoom;
+              sourceConfig.maxZoom = config.maxZoom;
+              if (sourceConfig) {
+                gaMapUtils.applyGlSourceToOlLayer(olLayer, sourceConfig);
+              }
+            };
+            var sourceId = config.sourceId;
+            if (sourceId) { olLayer.sourceId = sourceId; }
```
oterral: noNeed to test the existance of the sourceId . It will be undefined in the worse case

src/components/map/LayersService.js
```diff
@@ -799,44 +806,43 @@ goog.require('ga_urlutils_service');
...
+            if (glStyle) {
+              gaMapUtils.applyGlStyleToOlLayer(olLayer, glStyle);
+              applyTilesetJson(olLayer, glStyle);
+            } else if (styleUrl) {
```
oterral: Here I propose a small refactoring , it will avoid to create a function before, help to understand better the code:
``` 
var p =  (glStyle) ? $q.when(glStyle) : gaGlStyle.get(styleUrl);
p.then(function(glStyle) {
￼      if (!glStyle) {		￼  
￼          return;
￼      }	
```

src/components/map/LayersService.js
```diff
@@ -730,33 +733,36 @@ goog.require('ga_urlutils_service');
...
               opacity: config.opacity || 1
             });
-            if (styleUrl) {
-              gaGlStyle.get(styleUrl).then((glStyle) => {
-                gaMapUtils.applyGlStyleToOlLayer(olLayer, glStyle);
-                olLayer.setLayers(new ol.Collection(subLayers));
+            gaDefinePropertiesForLayer(olLayer);
+            if (glStyle) {
+              createSubLayers(olLayer, glStyle);
+            } else if (styleUrl) {
```
oterral: Same here. I propose a small refactoring , it will avoid to create a function before, help to understand better the code:
``` 
var p =  (glStyle) ? $q.when(glStyle) : gaGlStyle.get(styleUrl);
p.then(function(glStyle) {
￼    var subLayers = new Array(len);
    olLayer.glStyle = glStyle;
    for (i = 0; i < len; i++) { ....
```

src/components/backgroundselector/BackgroundService.js
```diff
@@ -214,20 +218,15 @@ goog.require('ga_glstylestorage_service');
 
         this.setById = function(map, newBgId) {
           if (map && (!bg || newBgId !== bg.id)) {
-            var newBg = getBgById(newBgId);
-            if (newBg) {
-              bg = newBg;
+            bg = getBgById(newBgId);
+            if (bg) {
```
oterral: This can lead to errors potentially because you could assign ` undefined`   to the current bg (value returned by `gaBg.get` )  used through all the appliction. That's why this asignment was done after the if. I would leave it as it was.

oterral:
voidLayer is also not connected to the list of bgs in vectorfeedbackdirective

src/js/EditController.js
```diff
@@ -35,9 +35,20 @@ goog.require('ga_mvt_service');
...
+        // set timeout to smooth the transition
+        setTimeout(function() {
+          gaMapUtils.applyGlStyleToOlLayer(
+              $scope.layer, $scope.layer.glStyle)
+        }, 600);
       }
+    };
+
+    $scope.$watch('globals.isEditActive', function(active) {
+      activate(active);
```
oterral: you can 't reload the style, only here. For example, if someone click on the 'open advanced link' in the feedback directive and the advancedStyle is already open , the style will not be reloaded. You also need to activate it in gaToggleEdit as before.

src/components/backgroundselector/BackgroundService.js
```diff
@@ -145,18 +145,22 @@ goog.require('ga_glstylestorage_service');
...
+      var createOlLayer = function(bg) {
+        var externalStyleUrl;
+        if (bg.externalStyleUrl) {
+          externalStyleUrl = bg.externalStyleUrl;
+        } else {
+          externalStyleUrl = bg.olLayer ?
+            bg.olLayer.externalStyleUrl : undefined;
         }
-        return layer;
+        bg.olLayer = gaLayers.getOlLayerById(bg.id, {
```
oterral: Why we need to recreate the layer eachtime ?

oterral: because of the ol bug, I guess

src/js/EditController.js
loicgasser: >  For example, if someone click on the 'open advanced link' in the feedback directive and the advancedStyle is already open , the style will not be reloaded. 

And this is correct. I was correcting the error where clicking the menu "Style the map" wouldn't re-apply the style. Anyway since we move all the edit functions to the advance mode and all editions will now be permanent we won't have to handle 2 styles anymore.

src/components/backgroundselector/BackgroundService.js
loicgasser: ok but we should not pass in undefined here. We should check that before calling setById. I don't like methods that silently fail, I don't think it's good design.

src/components/backgroundselector/BackgroundService.js
```diff
@@ -145,18 +145,22 @@ goog.require('ga_glstylestorage_service');
...
+          externalStyleUrl: externalStyleUrl,
+          glStyle: bg.olLayer ? bg.olLayer.glStyle : undefined
```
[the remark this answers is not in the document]
loicgasser: indeed!

src/components/backgroundselector/BackgroundService.js
```diff
@@ -145,18 +145,22 @@ goog.require('ga_glstylestorage_service');
...
-          layer.background = true;
-          layer.displayInLayerManager = false;
-          bg.olLayer = layer;
+      var createOlLayer = function(bg) {
+        var externalStyleUrl;
+        if (bg.externalStyleUrl) {
+          externalStyleUrl = bg.externalStyleUrl;
+        } else {
+          externalStyleUrl = bg.olLayer ?
+            bg.olLayer.externalStyleUrl : undefined;
```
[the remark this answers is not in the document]
loicgasser: yes

src/components/map/LayersService.js
loicgasser: evrything is a promise!

src/components/map/LayersService.js
loicgasser: yep

src/components/backgroundselector/BackgroundService.js
loicgasser: yes correct

src/components/backgroundselector/BackgroundService.js
loicgasser: I will leave though

src/components/map/LayersService.js
loicgasser: yeah but styleUrl must be defined as well if we want to perform a request that is not always the case

loicgasser: I can do that

```javascript
            if (glStyle || styleUrl) {
              p =  (glStyle) ? $q.when(glStyle) : gaGlStyle.get(styleUrl);
              p.then(function(glStyle) {
                createSubLayers(olLayer, glStyle);
              });
            } else {
              createSubLayers(olLayer);
            }
```
if you prefer

src/components/backgroundselector/BackgroundService.js
loicgasser: We never re-used layers for a reason. I think we can prevent a lot of unwanted bugs by simply re-creating a new instance each time. This stands for future versions of ol as well.

loicgasser:
> And I find a bug:
open the app
change color of Landuse residential
change to landcover grass
----> the color of Landuse residential disappear. Probably the inial style is reloaded.

This how it was defined in the spec. Always restet the style only play around with temp changes. Now that we remove edition from the feedback we re good to go.

PR #4590, opened 2018-11-19
install node via nvm for build

danduk82:
This PR add a `make env` target, which installs `nvm` and uses it to install `node`. Both versions of nvm and node can be specified in environment variables `NVM_VERSION` and `NODE_VERSION`, but have default (tested) values in `Makefile`.
The `env` target is set as a dependency for `make user` target to ease the process so that we always only need to build our user environment and it works.

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/ltbon_add_node_nvm/index.html)</jenkins>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,6 +133,8 @@ LAST_VERSION = $(call lastvalue,version)
 VERSION := $(shell if [ $(KEEP_VERSION) = true ] && [ '$(LAST_VERSION)' != '-none-' ]; then echo '$(LAST_VERSION)'; else date '+%y%m%d%H%M'; fi)
 NAMED_BRANCH ?= true
 DEEP_CLEAN ?= false
+NVM_VERSION ?= v0.33.8
+NODE_VERSION ?= 6.13.1
 
 
 # S3 deploy variables
@@ -203,6 +205,7 @@ help:
 	@echo "Possible targets:"
 	@echo
 	@echo "- user                Build the app using user specific environment variables (see $(USER_SOURCE) file)"
+	@echo "- env                 Install NVM and set the correct node version to build the application"
 	@echo "- all                 Build the app using current environment variables"
 	@echo "- build               Build the app using current environment variables. No linting and testing."
 	@echo "- dev                 Build the app using dev environment variables (see rc_dev file). No linting and testing."
@@ -279,6 +282,13 @@ user:
 
 .PHONY: build
 build: showVariables .build-artefacts/devlibs .build-artefacts/requirements.timestamp $(SRC_JS_FILES) debug release
+
+
+.PHONY: env
+env:
+	curl -o- https://raw.githubusercontent.com/creationix/nvm/$(NVM_VERSION)/install.sh | bash ;\
+	source $(HOME)/.bashrc ;\
+	nvm install $(NODE_VERSION)
 
 .PHONY: dev
 dev:
--- a/Makefile
+++ b/Makefile
@@ -277,7 +277,7 @@ showVariables:
 all: showVariables lint debug release apache testdebug testrelease fixrights
 
 .PHONY: user
-user:
+user: env
 	source $(USER_SOURCE) && make all
 
 .PHONY: build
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -274,12 +277,19 @@ showVariables:
...
 
 .PHONY: build
 build: showVariables .build-artefacts/devlibs .build-artefacts/requirements.timestamp $(SRC_JS_FILES) debug release
 
+
+.PHONY: env
+env:
+	curl -o- https://raw.githubusercontent.com/creationix/nvm/$(NVM_VERSION)/install.sh | bash ;\
+	source $(HOME)/.bashrc ;\
+	nvm install $(NODE_VERSION)
```
loicgasser: I would try to launch the install command only when you want to update NVM. Otherwise you will always curl nvm program. Create a new target use the `lastvalue` function to store node and nvm versions.

danduk82: Hi @loicgasser , thanks for the feedback. Something like this?

procrastinatio: After `nvm install`, you have to activate a version `nvm use $(NODE_VERSION)`

danduk82: voilà

loicgasser: `.build-artefacts/node-version` will be executed only if `.build-artefacts/last-node-version` has changed so I don't think the if statement is needed.

loicgasser: see `cachelastvariable` https://github.com/geoadmin/mf-geoadmin3/blob/master/Makefile#L703

danduk82: didn't work without the `ifndef`, but I might have done something strange

loicgasser: Remove the phony from the materialized target

danduk82: If I try like this: 
```
.build-artefacts/nvm-version: .build-artefacts/last-nvm-version
        curl -o- https://raw.githubusercontent.com/creationix/nvm/$(NVM_VERSION)/install.sh | bash

.build-artefacts/node-version: .build-artefacts/last-node-version
        source $(HOME)/.bashrc ;\
        nvm install $(NODE_VERSION)

.PHONY: env
env: .build-artefacts/nvm-version .build-artefacts/node-version
        source $(HOME)/.bashrc && nvm use $(NODE_VERSION)
```
the targets `build-artefacts/nvm-version` and `.build-artefacts/node-version` are always executed

loicgasser: yes because
`.build-artefacts/nvm-version` and  `.build-artefacts/node-version` are not materialized actually.
Sticky situation. leave it like this if it works.

danduk82: "ça marche, on touche pas" ...

gjn:
Since this PR, the build is not working anymore for people who have/had nvm already installed. Maybe even the new ones.

The deploy yesterday also didn't work out of the box.

Makefile
```diff
@@ -274,12 +280,30 @@ showVariables:
...
+user: env
 	source $(USER_SOURCE) && make all
 
 .PHONY: build
 build: showVariables .build-artefacts/devlibs .build-artefacts/requirements.timestamp $(SRC_JS_FILES) debug release
 
+
+.PHONY: .build-artefacts/nvm-version
+.build-artefacts/nvm-version: .build-artefacts/last-nvm-version
+ifneq ($(LAST_NVM_VERSION),$(NVM_VERSION))
```
gjn: We should only install it when it's not already installed. So check first if nvm command exists.

Makefile
```diff
@@ -274,12 +280,30 @@ showVariables:
...
+	curl -o- https://raw.githubusercontent.com/creationix/nvm/$(NVM_VERSION)/install.sh | bash
+endif
+
+.PHONY: .build-artefacts/node-version
+.build-artefacts/node-version: .build-artefacts/last-node-version
+ifneq ($(LAST_NODE_VERSION),$(NODE_VERSION))
+	source $(HOME)/.bashrc ;\
+	nvm install $(NODE_VERSION)
```
gjn: Only install it if not already installed. So user nvm use first.

Issue #4587, opened 2018-11-19
layers_visibility=false doesn't work for external WMTS layers


smug007:
In the following link one **external** WMTS layer is added. layers_visibility is set to false. If the link is clicked, geo.admin.ch displays the WMTS layer even though it shouldn't. With **internal** WMTS (like slope inclination layer of Swisstopo) there is no such problem.

https://map.geo.admin.ch/?lang=de&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&zoom=7&Y=694000&X=168000&layers=WMTS%7C%7CAvalancheATHM%7C%7Chttps:%2F%2Fmap.skitourenguru.ch%2FATHM_GetCapabilities.xml&layers_opacity=0.3&layers_visibility=false

procrastinatio:
@smug007 Can you check it and give a feedback?

smug007:
Excellent, that bugfix **resolves** the problem.

PR #4591, opened 2018-11-19
Do not make WMTS visible if it should not be visible

procrastinatio:
[Test link mit WMTS](https://mf-geoadmin3.int.bgdi.ch/fix_4587/index.html?lang=de&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&zoom=7&Y=694000&X=168000&layers=WMTS%7C%7CAvalancheATHM%7C%7Chttps:%2F%2Fmap.skitourenguru.ch%2FATHM_GetCapabilities.xml&layers_opacity=0.3&layers_visibility=false) Should open with a layer Avalanche hazard in the catalogue but it should not already be active.

See  #4587

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/fix_4587/index.html)</jenkins>

### DIFF
--- a/src/components/map/WmtsService.js
+++ b/src/components/map/WmtsService.js
@@ -214,7 +214,8 @@ goog.require('ga_urlutils_service');
               layerIdentifier, options.capabilitiesUrl);
           if (layerOptions) {
             layerOptions.opacity = options.opacity || 1;
-            layerOptions.visible = options.visible || true;
+            layerOptions.visible = options.visible === false ?
+              options.visible : true;
             return createWmtsLayer(layerOptions);
           }
         };
--- a/src/components/map/WmtsService.js
+++ b/src/components/map/WmtsService.js
@@ -214,8 +214,7 @@ goog.require('ga_urlutils_service');
               layerIdentifier, options.capabilitiesUrl);
           if (layerOptions) {
             layerOptions.opacity = options.opacity || 1;
-            layerOptions.visible = options.visible === false ?
-              options.visible : true;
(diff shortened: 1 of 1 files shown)


src/components/map/WmtsService.js
```diff
@@ -214,7 +214,8 @@ goog.require('ga_urlutils_service');
               layerIdentifier, options.capabilitiesUrl);
           if (layerOptions) {
             layerOptions.opacity = options.opacity || 1;
-            layerOptions.visible = options.visible || true;
+            layerOptions.visible = options.visible === false ?
+              options.visible : true;
```
ltkum: After some testing ([Test Link](https://mf-geoadmin3.dev.bgdi.ch/ltkum/index.html?lang=de&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&zoom=7&Y=694000&X=168000&layers_opacity=0.3&layers_visibility=false&layers=WMTS||AvalancheATHM||https:%2F%2Fmap.skitourenguru.ch%2FATHM_GetCapabilities.xml)), it seems like transformation of string to boolean is already done somewhere else. which means whatever you input as a String, only the string "false" will return flase. in that context, yuo can use layerOptions.visible=options.visible && true.

ltkum: Okay so, I did the change and pushed it. If it works, you can merge. I would propose testing with false, but also with some other strings (null springs to mind, but any string is fine). I did on dev, but better safe than sorry.

ltkum:
True. let's leave it at that then ^^

Issue #4611, opened 2018-11-29
[MVT] 3D mode is not available


davidoesch:
_OS/Device_
all OS /Browsers

_Steps to reproduce_
start https://mf-geoadmin3.int.bgdi.ch/mvt/index.htm

_result_
no 3d Button

_expected result_
3d Button available, greyed /deactaived out when in vectormode (if we can't run it in 3d)

_Workaround:_
None

loicgasser:
Ok, 3d is simply deactivated. Because we also have swissimage.
Out of 4 layers, 2 are not 3d compatible.

davidoesch:
sad but true--- but we would like to test it. add it

loicgasser:
Fixed

Issue #4606, opened 2018-11-28
[MVT] Localization : red ballon not shown


davidoesch:
_OS/Device_
ALL OS
ALL Devices

_Steps to reproduce_

1.        Open application  https://mf-geoadmin3.int.bgdi.ch/mvt/index.html?
2.        Enter ""Genè"" in localization textbox and select ""Genève"" /same for ""Längmatt 19 3608 Thun""
3.        Enter 200000 600000 in localization textbox
4.        Enter 7.5 47 in localization textbox
5. Enter ""evech"" in localization textbox and select ""Rue de l'Évêché 1 1204 Genève""
6.        Write ""?swisssearch=payerne"" in the URL and then enter"

_result_
no red ballon on result

_expected result_
red ballon on position

_Workaround:_
none

loicgasser:
1. NA
2. This is correct and incorrect in prod. we only add a marker for the types address, parcel and gazeteer. See https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/search/SearchTypesDirectives.js#L29 for add overlay.
3. Fixed
4. FIxed. `7.5 47` is outside Switzerland, try `47 7.5` instead. I fixed it to have the same behavior than in prod, but that means we cannot input coords outside CH.
5. works
6. works search suggestions appear

davidoesch:
Thanks

loicgasser:
Fixed

Issue #4545, opened 2018-11-01
[MVT] change bg switcher


davidoesch:
bg switcher contains
- [x] existing ch.swisstopo.pixelkarte-farbe Color Map Icon
- [x] existing voidLayer "white background  Icon
- [x]  existing ch.swisstopo.swissimage 
- [x] new  Lightbase Map (Instead of Grey map) @pauloamado use Positron https://openmaptiles.org/
- [ ] remove temporary Openmaptiles (for the time beeing until swisstopo finally provides once in the blue moon lightbase and hybrid map)

linked to https://github.com/geoadmin/mf-geoadmin3/issues/4546

pauloamado:
Here is the new bg (png sprite):

1st part is the Hybrid Map
2nd part is the Lightbase Map
3rd part is the regular Color Map

Thx for feedback.
![bg_switcher_vector](https://user-images.githubusercontent.com/6584138/48249530-afc4b200-e3fb-11e8-944a-64c0a2d6e941.png)

davidoesch:
thx!

oterral:
@davidoesch do you really want wmts and void layers  in the bg list ? The point is to force user to test the vector tile layer, if he saves a link with the wmts map I m pretty sure it will not understand why  he can't modify the style anymore.

davidoesch:
Yes we want. Same bg  selector as in current prod. The only thing: grey is replaced with the light basemap. The goal is not only vector tile but also pixel Karte in web Mercator

oterral:
ok

loicgasser:
Fixed

PR #4613, opened 2018-11-30
Remove ovt, re-enable 3D, fix crossair

loicgasser:

Fixes for 

https://github.com/geoadmin/mf-geoadmin3/issues/4611
https://github.com/geoadmin/mf-geoadmin3/issues/4545
https://github.com/geoadmin/mf-geoadmin3/issues/4606

### DIFF
--- a/src/components/search/SearchService.js
+++ b/src/components/search/SearchService.js
@@ -173,18 +173,19 @@ goog.require('ga_reframe_service');
             left > right ? left : right,
             right < left ? right : left
           ];
-
+          var pos95 = ol.proj.transform(position, 'EPSG:2056',
+              gaGlobalOptions.defaultEpsg);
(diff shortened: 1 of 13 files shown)


src/components/search/SearchService.js
```diff
@@ -173,18 +173,19 @@ goog.require('ga_reframe_service');
             left > right ? left : right,
             right < left ? right : left
           ];
-
+          var pos95 = ol.proj.transform(position, 'EPSG:2056',
+              gaGlobalOptions.defaultEpsg);
```
oterral: No more use of no more use of gaReframe ?  we should ask @davidoesch if it s ok.

loicgasser:
No reframe for web mercator oli.

Issue #4600, opened 2018-11-28
IE11 3D controls are not centered


davidoesch:
open on win10 office bit laptop 3d

https://map.geo.admin.ch/?topic=ech&lang=de&bgLayer=voidLayer&layers=ch.swisstopo.zeitreihen,ch.swisstopo.swissnames3d&layers_timestamp=19501231,&lon=8.37866&lat=46.93821&elevation=3431&heading=318.131&pitch=-28.214&time=1950

switch to 3d

left aligned controls instead centered controls

![image](https://user-images.githubusercontent.com/4577727/49142507-3a147d80-f2f9-11e8-8412-84065ee7f4d1.png)

and 3d in edge is way faster than  in chrome

pakb:
fixed by https://github.com/geoadmin/mf-geoadmin3/pull/4630

PR #4630, opened 2018-12-04
Centering 3d tools on IE11

procrastinatio:

See https://github.com/geoadmin/mf-geoadmin3/issues/4600

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/fix_4600/index.html)</jenkins>

### DIFF
--- a/src/components/controls3d/style/controls3d.less
+++ b/src/components/controls3d/style/controls3d.less
@@ -7,7 +7,6 @@
   bottom: unit(@bt-bottom, px);
   margin-left: auto;
   margin-right: auto;
-  max-width: 250px;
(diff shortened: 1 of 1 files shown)


src/components/controls3d/style/controls3d.less
```diff
@@ -7,7 +7,6 @@
   bottom: unit(@bt-bottom, px);
   margin-left: auto;
   margin-right: auto;
-  max-width: 250px;
```
oterral: If you remove this, the div takes all the width of the window and you can't interact with  the map in this zone. can be annoying .

Issue #4631, opened 2018-12-04
[MVT] WM(T)S IMPORT: cover now global imports


davidoesch:
OS: all OS all Browser

To reproduce:
open basemap.at 
https://mf-geoadmin3.int.bgdi.ch/mvt_ovt/index.html?topic=ech&lang=en&bgLayer=ch.swisstopo.pixelkarte-farbe&lon=11.47744&lat=47.15315&zoom=7&layers=WMTS%7C%7Cbmaphidpi%7C%7Chttp:%2F%2Fwww.basemap.at%2Fwmts%2F1.0.0%2FWMTSCapabilities.xml

or

https://mf-geoadmin3.int.bgdi.ch/mvt_ovt/index.html?topic=ech&lang=en&bgLayer=ch.swisstopo.pixelkarte-farbe&lon=11.47744&lat=47.15315&zoom=7&layers=WMTS%7C%7CWorld_Street_Map%7C%7Chttps:%2F%2Fservices.arcgisonline.com%2Farcgis%2Frest%2Fservices%2FWorld_Street_Map%2FMapServer%2FWMTS%2F1.0.0%2FWMTSCapabilities.xml
Result:

clipped at old LV95 tileset border

Expected result:
I See now the whole world...this is the whole point about webmercator and its chances for interoperability

procrastinatio:
See https://github.com/geoadmin/mf-geoadmin3/pull/4637

PR #4625, opened 2018-12-04
Parsing webmercator in search box

procrastinatio:
Bern should be 828064.773289789, 5934093.187456297 Webmercator

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/mom_webmerc_search/index.html)</jenkins>

Issue #4618, opened 2018-12-03
[MVT] localization: wrong results for search with LV03 and LV95 coordinates


mariokeusen:
OS: WIn 10, desktop, Chrome (same issue also on EDGE)
Steps to reproduce: open app, enter "600000 200000" (or "2600000 1200000") into the search field
Result: app pans/zooms somewhere northeast of Stockholm, Sweden
Expected: app pans/zooms to Berne, Switzerland as on productive map.geo.admin.ch

mariokeusen:
Related to this: also the Swiss coordinates in the position window are wrong. Here what the app returns in Berne: 
![image](https://user-images.githubusercontent.com/4577749/49374864-f0fe6800-f702-11e8-847c-db4a57b3aa45.png)

procrastinatio:
Now we try to guess what the user inputs based on the extent of LV03, LV95 and WGS84. 
Is Webmerc extent restricted to Switzerland? If yo, me may assume northing larger than 5'000'000 to be Webmerc, if not we cannot guess.

procrastinatio:
Fixed with https://github.com/geoadmin/mf-geoadmin3/pull/4625

procrastinatio:
Small side remarks. 
Apparently, we are now global (see https://github.com/geoadmin/mf-geoadmin3/issues/4631)
There is **no** way to guess what system a user is using, as all LV95 and LV03 coordinates are also valid Webmercator coordinates.

p1d1d1:
What about "synchronizing" the search with the SRS selected in the pull-down menu bottom left?

davidoesch:
+1 for Pasquales proposal. Needs a hint when entering coordinates in the search field

loicgasser:
Coordinates in search are ok bow. We need to fix the context popup.

p1d1d1:
@loicgasser 
Not in here: https://mf-chsdi3.int.bgdi.ch/mvt/shorten/7ea9d07d3a
Try with 2563937 1129848 (EPSG:2056) and 775944 5831628 (corresponding EPSG:3857)

loicgasser:
Swiss coords do work. Web merc don't because we have feature parity with prod. Try F5

procrastinatio:
We are global now:  https://github.com/geoadmin/mf-geoadmin3/issues/4631

procrastinatio:
[...]

PR #4637, opened 2018-12-05
[MVT] Go global

procrastinatio:
[Test link with global WMS/WMTS and Swiss Extent](https://mf-chsdi3.int.bgdi.ch/mvt/shorten/7eaf3f8386)

1. Check if you can add any WMS around the world
2.  Check if you can still use LV03/LV95 to go home.

Should fixe https://github.com/geoadmin/mf-geoadmin3/issues/4631

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/mvt_global_extent/index.html)</jenkins>


oterral:
@procrastinatio  doesn't it break the swiss coordinate detection ?

oterral:
I will merge if @davidoesch is ok with that.

davidoesch:
> @oterral Jein, if coordinates are metrics (>180), it looks if inside the LV03 and LV95 bboxes, and if not, it is assumed it is epsg:3857. But any LV03 or LV95 coordinates are also valid epsg:3857 coordinates.

can we fix this with https://github.com/geoadmin/mf-geoadmin3/issues/4618#issuecomment-444437915 proposal?

davidoesch:
> I will merge if @davidoesch is ok with that.

I wonder which sideeffect it will have. anyway, the application has to start centered on switzerland as MVT branch currently does....

loicgasser:
This will break coords detection. We do need a an extent for Switzerland.

loicgasser:
It can be another variable though. But you need to adapt other part of the code. Notably the search service, search types also I believe and maybe some more.

davidoesch:
if this PR (which was about supporting about external WMS/WMTS) breaks most cases of https://help.geo.admin.ch/?id=24 

-> a no go. we need a solution here which works with LV95 without additional user interaction

procrastinatio:
@loicgasser So, search is the only issue? So this is not a problem. See my two last comments on https://github.com/geoadmin/mf-geoadmin3/issues/4618#issuecomment-444437915

loicgasser:
the problem is here:
https://github.com/geoadmin/mf-geoadmin3/blob/mvt/src/components/search/SearchService.js#L186

everything will always be interpreted as `EPSG:2056` since the point is always contained within the extent it will NEVER be matched against `EPSG:21781`.

Just try to input 700499.35, 182750.43
This should be interpreted as `EPSG:21781` it works on mvt but not on this branch.
So keep a ref to the swiss extent in another variable and replace in the SerachService.

Please have a look at what has been done in https://github.com/geoadmin/mf-geoadmin3/pull/4613/files

And I agree she don't care about inputting WGS coords.

so keep global extent to test lon/lat, but use swiss extent to test swiss coords.

procrastinatio:
@loicgasser Also my PR https://github.com/geoadmin/mf-geoadmin3/pull/4625 and my comment in https://github.com/geoadmin/mf-geoadmin3/issues/4618
I we assume the LV03 and LV95 bboxes as swiss, we cannot access the two area of the Gulf of Guniea and South Sudan with Webmercator. 

OK, I'll add a varaible swiss_extent to the mix.

procrastinatio:
And what do we do with Skyguide black magick? We were assuming something like 7 47 to be interpreted as 7°E and 47°N, but it could be anything else.

davidoesch:
> And what do we do with Skyguide black magick? We were assuming something like 7 47 to be interpreted as 7°E and 47°N, but it could be anything else.

We support officiall what is in
https://help.geo.admin.ch/?id=24&lang=de

46.97984 6.60757 (DD) ◈ 46°58.7904' 6°36.4542' (DM) ◈ 46°58'47.424'' 6°36'27.252'' (DMS, double single quote) ◈ 46°58'47.424" 6°36'27.252" (DMS, double quote)

so the y need to enter 7.0 47. 0

Makefile
```diff
@@ -111,26 +111,19 @@ TRANSLATE_EMPTY_JSON ?= src/locales/empty.json
...
 DEFAULT_ELEVATION_MODEL ?= COMB
 
-# Map (Mercator)
+# Map variables (Mercator)
 DEFAULT_EPSG = EPSG:3857
 DEFAULT_EPSG_EXTEND = '[-20037508.3428, -20037508.3428, 20037508.3428, 20037508.3428]'
-DEFAULT_EXTENT = '[558147.7958306982, 5677741.814085617, 1277662.36597472, 6152731.529704217]'
+# Only for WMS services
+# The Swiss bounds expressed as $DEFAULT_EPSG 
+SWISS_EXTENT = '[558147.7958306982, 5677741.814085617, 1277662.36597472, 6152731.529704217]'
```
loicgasser: but then use it no?

loicgasser:
@davidoesch this is now for the MVT branch. But it still requires some changes and data is incorrect at global scale.

davidoesch:
yes at MVT I still not global coverage: https://mf-chsdi3.int.bgdi.ch/mvt/shorten/7ec2e0dfcf 

image


or did I understand sth wrong?

procrastinatio:
Need more work.

- All tests in the branch `mvt` are still EPSG:2056 (see https://github.com/geoadmin/mf-geoadmin3/blob/mvt/test/specs/Loader.spec.js#L63)
- Too many black magick to guess the coordinates which need to be removed
- We may search outside `swissExtent`, do we?

davidoesch:
Wenn may ... But for test viewer stick to simplicity

oterral:
@procrastinatio outside swissextent, you use lon lat coord

procrastinatio:
Test
[Test with many WMS/WMTS](https://mf-chsdi3.int.bgdi.ch/mvt/shorten/7eceae85c1)

`37UDB 13433 79521`   Red Square, Moscow
`40.68921, -74.04454`  or  `40°41′21.139″N 74°02′40.332″W`  Liberty Statue, New York

`6.96948 46.9712`  Not in  lat/long, but assumed in Switzerland

`6.96948W 46.9712N`If you want to go west of Brest, use quadrant

procrastinatio:
jenkins test

procrastinatio:
@loicgasser `extent`defines where coordinates are valid. `swissExtent`defines only the area where `LV03`and `LV95` are valid, and where you may use `lat/long`and `lon/lat` without quadrant indication (normally `7.5° 46.7°`should be interpreted as `7.5°N 46.7°W` but we are nice to the user)

oterral:
@loicgasser are we good with this PR?  the seacrh of coordinates david mentionned works

oterral:
@davidoesch are you agree with this functionnality?

davidoesch:
@oterral 
-> OK 
I tested: https://mf-geoadmin3.int.bgdi.ch/mvt_global_extent/index.html?

and they work four our official supported coordinate imput like defined under https://help.geo.admin.ch/?id=24 

CH1903+ / LV95: 2600'000 1200'000 ◈ 2600000 1200000 ◈ 2600 000 1200 000 ◈ 2600000/1200000 ◈ 2600000 / 1200000 ◈ 2600000,1200000 ◈ 2600000 1200000 ◈ 1200000 2600000

CH1903 / LV03: 600'000 200'000 ◈ 600000 200000 ◈ 600 000 200 000 ◈ 600000/200000 ◈ 600000 / 200000 ◈ 600000,200000 ◈ 600000 200000 ◈ 200000 600000

WGS 84 (lat/lon): 46.97984 6.60757 (DD) ◈ 46°58.7904' 6°36.4542' (DM) ◈ 46°58'47.424'' 6°36'27.252'' (DMS, double single quote) ◈ 46°58'47.424" 6°36'27.252" (DMS, double quote)

MGRS: "32TLT 98757 23913"

what3words: "zufall.anders.blaumeise"


and we do now support global WGS84 which makes sense


-> LGTM

PR #4639, opened 2018-12-05
v004 of lbm is in production, we will give it to you

ltkum:

Version 4 is out and my previous pull request was merged a little bit too fast. v003 is no longer on prod, by the way.

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/ltkum_lbm_v004/index.html)</jenkins>

### DIFF
--- a/src/components/map/LayersService.js
+++ b/src/components/map/LayersService.js
@@ -308,9 +308,9 @@ goog.require('ga_urlutils_service');
               }, {
                 serverLayerName: 'ch.swisstopo.vektorkarte.vt',
                 opacity: 0.75 // Show swissalti
-              }/*, {
+              }, {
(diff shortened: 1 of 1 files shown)


ltkum:
@davidoesch @danduk82 

Just to keep you two updated, and perhaps let your try the test link

davidoesch:
yep... https://mf-geoadmin3.int.bgdi.ch/mvt/index.html? is broken letme knwo when we can continue with the test  cc @oterral @loicgasser

ltkum:
@davidoesch 
I redeployed a tileserver prod version with both v003 and v004, that way, you aren't hindered while waiting for this PR to be reviewed.

davidoesch:
thanks
@oterral @loicgasser  you can switch the app to 004

PR #4642, opened 2018-12-06
Fix nvm usage by adding the same loading script as in .bash_profile

pakb:
This assume that env variable NVM_DIR is defined

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/bap_fix_makefile/index.html)</jenkins>

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -26,6 +26,10 @@ node(label: 'jenkins-slave') {
       checkout scm
     }
     
+    stage('env') {
+      sh 'make env'
+    }
+
     stage('Lint') {
       sh 'make lint'
     }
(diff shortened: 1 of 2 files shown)


oterral:
Works for me

oterral:
I merge too fast, for ng-annotate use it's not useable.

oterral:
Just run ` make clean user`  see the difference with before.  Now it set the nvm version for each annotated file, which means  hundred of times.

pakb:
If ng-annotate doesn't need explicitly node version 6.13.0 but is fine with version 5.x.x (which is the default version delivered with debian) we could remove this NVM call before ng-annotate.
But that would leave us with a future issue when ng-annotate will need a higher version of node.

Alternatively, we could have as prerequisite for this project to have node 6.13.0 installed (and stop building it if not at the correct version)

PR #4658, opened 2018-12-14
[MVT] Update of ol

oterral:
I couldn 't wait monday :)
Needs https://github.com/openlayers/openlayers/pull/9063 to fix geoblocks build
@davidoesch @loicgasser 

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/teo_upol5/index.html)</jenkins>


davidoesch:
-iOS: a little bit better: bust still shows only hillshade when panning
-edge: significantely better

@gjn @procrastinatio  your thoughts?

loicgasser:
Yes better. I'd say we gain in fluidity for sur when panning. It is not as laggy.

oterral:
This update of ol breaks the gray rectangle we display for the offline and the print.
Caused by the change fo use for postcompose and precompose event.

oterral:
Print and offline rectangle are fixed. Ready to review/merge @loicgasser

oterral:
Contains also the ios fix

src/index.mako.html
```diff
@@ -345,7 +345,7 @@
         </div> <!-- .panel -->
 
         <!-- Print -->
-        <div class="panel" ng-show="false && !globals.is3dActive">
+        <div class="panel" ng-show="!globals.is3dActive">
```
loicgasser: Why do we leave the print? It won't work with mvt.

src/components/offline/OfflineDirective.js
```diff
@@ -115,7 +117,7 @@ goog.require('ga_networkstatus_service');
   });
 
   module.directive('gaOfflineSelector', function($timeout, $window, gaOffline,
-      gaStorage) {
+      gaStorage, gaGlobalOptions, gaVector) {
```
loicgasser: gaGlobalOptions is imported but unused

src/components/map/MapUtilsService.js
```diff
@@ -31,6 +31,18 @@ goog.require('ga_urlutils_service');
...
+      // For mobile, redefine disposeInternal function for mvt.
+      // TODO: verify if it's useful
+      var disposeInternal = ol.VectorImageTile.prototype.disposeInternal;
+      ol.VectorImageTile.prototype.disposeInternal = function() {
+        for (var key in this.context_) {
+          var context = this.context_[key];
+          context.canvas.width = context.canvas.height = 0;
+        }
+        disposeInternal.call(this);
+      };
```
loicgasser: I don't understand why you need to overload disposeInternal.
What does it fix?

loicgasser:
We need to fix the compare tool before this can be merged.
I'll do it over the vacations.

loicgasser:
Ok  compare fixed and print + offline deactivated

PR #4675, opened 2018-12-27
Fix #4484 (3d camera terrain collision)

### LINKED ISSUE #4484: 3d: positioning pegamn in a steep location: undergroudn view
All OS ALL Viewer

move pegman into a steep hil
move mouse
camera position goes suddenly undergroudn, screen starts flickering, trees buildings seen from below

expected result:
- collision detection si off
- camera position is never under ground

pakb:
Possible fix for https://github.com/geoadmin/mf-geoadmin3/issues/4484

In 3d pegman, check altitude 25m ahead (in camera direction) and raise the camera height if the camera could see through the terrain because of high slope.
If altitude 25m ahead is above the default height set to the camera, the camera is raised to this altitude.

Please review this thoroughly, as I'm not (yet) aware of possible side effect for this kind of changes.

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/bap_fix_3d_camera_terrain_collision/index.html)</jenkins>

### DIFF
--- a/src/components/controls3d/fps.js
+++ b/src/components/controls3d/fps.js
@@ -348,27 +348,31 @@ FPS.prototype.getTimeDifference_ = function() {
  * @return {!Cesium.Cartesian3}
  * @private
  */
-FPS.prototype.clampAboveTerrain_ = function(gpos, minHeight, optMaxHeight, cameraHeading) {
+FPS.prototype.clampAboveTerrain_ = function(gpos, minHeight, optMaxHeight, 
+  cameraHeading) {
   var lla = this.ellipsoid_.cartesianToCartographic(gpos),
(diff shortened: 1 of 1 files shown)


davidoesch:
tried several in https://mf-chsdi3.int.bgdi.ch/shorten/7f1b03b25b

works better now
  
observation once in a while : app starts to flicker in some cases  (under ground / above groumd)
I do wonder how the latest cesiumJS branch works on this

IMHO LGTM--- but codewise I suggest @procrastinatio  for a review

pakb:
Usually the piece of code adjusting camera height is called after the scene is loaded with default values. So the first time the camera is positioned, it might be "underground" until the routine that place it above ground is called. That's maybe the reason you've still seen some flickering

PR #4678, opened 2019-01-04
Add embedded scroll handling (need testing with iOs)

pakb:
While embedding the map, add handling for "wheel" event with mouse input : disabling zoom on map if CTRL key is not pressed while scrolling, and passing the event to the webpage embedding the map (so that the webpage embedding the map continues its scroll).
Same kind of behavior with touch event (for mobile) : disabling one finger gestures in favor of two fingers gesture, disabling two fingers map rotation to ease pan/zoom with two fingers.
Those behaviors are to avoid scrolling down the webpage embedding the map and accidentally zooming the map instead of continuing scrolling up or down (same behavior as embedded GoogleMaps).

Known issue (to test) : on iOs (last version), with codepen.io on either Chrome or Safari, the webpage jumps back to top when touch events are being handled by the map on my device.

@davidoesch @danduk82 I will need some help translating two new texts in Italian and Romansh, how should I do that?

**Example :**
live : https://codepen.io/pakb/full/bOroYP
screenshot :
image

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/bap_iframe_scrolling/1907040618/index.html)</jenkins>

### DIFF
--- a/src/components/embeddedscroll/EmbeddedScrollDirective.js
+++ b/src/components/embeddedscroll/EmbeddedScrollDirective.js
@@ -9,10 +9,16 @@ goog.provide('ga_embedded_scroll_directive');
   // embedded maps, à la GoogleMaps)
   // see : https://developers.google.com/maps/documentation/embed/guide
   module.directive('gaEmbeddedScrollEventHandler',
-      function($rootScope, $translate, $timeout, $window) {
+      function($rootScope, $translate, $timeout, $window, gaGlobalOptions) {
         return {
           restrict: 'A',
           link: function(scope, element, attrs) {
+            // if embedded but gaGlobalOptions.embeddedWithInteractions is true
+            // we don't attach listeners
+            if (gaGlobalOptions.embeddedWithInteractions) {
+              return;
+            }
+
             var datetimeLastAction = null;
             var hideAfterAWhileIfSameDatetime = function(datetime) {
               // check after 2.5 sec if no other action occured, hide overlay
--- a/src/components/embeddedscroll/EmbeddedScrollDirective.js
+++ b/src/components/embeddedscroll/EmbeddedScrollDirective.js
@@ -1,0 +1,95 @@
+goog.provide('ga_embedded_scroll_directive');
+
+(function() {
+  var module = angular.module('ga_embedded_scroll_directive', []);
+
+  // top level directive (on body or html) that will hook listener for mouse
+  // and touch events. It will trigger another directive (bellow) to show
+  // an overlay with hints for embedded maps (as we limit interaction with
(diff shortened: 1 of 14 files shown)


davidoesch:
- before merging: we need to test this in depth (50 percent of traffic is from iframe$
- languages: state the msgid I'll translate and commit

pakb:
Those two :
`"embedded_scroll_hint": "Use ctrl + scroll to zoom the map",`
`"embedded_touch_hint": "Use two fingers to move the map",`
Thanks

davidoesch:
[ ] make sure it only works on mobile on iframe per default it is on
[ ] on Desktop it is per default off : IMHO we need a parameter https://developers.google.com/maps/documentation/javascript/interaction 

[ ]  test sites on cms with hidden supages

procrastinatio:
On iPad Pro, pan is not disabled, but appears gray and misses the translation. Same behaviour on older iPad (only text is correcly translated). Rotate is disabled in both case.

davidoesch:
Hi I added the test site with the gmaps with the desired behaviour and  the current behaviour below based on https://mf-geoadmin3.int.bgdi.ch/bap_iframe_scrolling/index.html

-> https://cms.geo.admin.ch/test/pull_4678.htm

pakb:
Should I add a param to disable this on demand, or more like gmaps detect parent webpage's size and define with thresholds if scroll should be allowed or not?

davidoesch:
+1 for a param. default on--adn can be switched back by setting it off @gjn your thoughts
- center "use ctrl +scroll ..." to the center of the iframe
- Font size "use ctrl +scroll ..." can be bigger

davidoesch:
and 
DE
"embedded_scroll_hint": "Verwende Strg + Scrollen zum Zoomen der Karte",
"embedded_touch_hint": "Verschieben der Karte mit zwei Finger",

IT
"embedded_scroll_hint": "Utilliza CTRL + scorrimento per eseguire le zoom della mappa",
"embedded_touch_hint": "Utilliza due dita per spostare la mappa ",

RM
"embedded_scroll_hint": "smatgai CTRL + scrollar per  zoom charta",
"embedded_touch_hint": "Spustar charta cun  dus dets",

pakb:
@davidoesch param has been added. You can see results on the codepen linked above.

davidoesch:
On ios https://codepen.io/pakb/full/bOroYP does Not Work

davidoesch:
Hmm https://cms.geo.admin.ch/test/pull_4678.htm does not work on ios

pakb:
Yes we have some issues with `make s3deploybranchint` that I'll have to fix before we can test

PR #4687, opened 2019-01-10
Using a central repository to store map.geo.admin.ch config files

procrastinatio:
The main idea is to configure the application without redeploying it.

This PR does basically nothing, except managing the various config layers differently. Instead of statifying them within the application, there are stored in the same bucket but may be updated independently.

The basic layout for the config dir is:
```
  configs\<lang>service.json  (will be translated in the futur)
          \<lang>\layersConfig.json
          \<lang>\catalog.swisstopo.json
```
A `CONFIG_URL`is added to point at different configuration.

Let's say, we have a new `dummy` layer, basically the same as the previous, but with different settings (name, opacity, update time, etc.) with `CONFIG_URL` pointing to a non-default location.

[Alternative settings](https://mf-geoadmin3.int.bgdi.ch/config/index.html?config_url=//mf-geoadmin3.int.bgdi.ch/configs/5555&lang=de&layers=dummy)

On **prod** and **int**, all branches are using `configs` defined in //mf-geoadmin3.(int|prod).bgdi.ch/configs. On **dev**, only `master`.

On **dev**, as the project is served by `apache`, `release`code uses a local cache, `debug` code always points on the corresponding `api` service.  

Upload config with:
 `make s3uploaconfigint`  or
  `make s3uploadconfigprod`

**Example**

1/ Using the default config

[Default settings](https://mf-geoadmin3.int.bgdi.ch/config/index.html?lang=de&layers=ch.bafu.hydrologie-messstationen_gefahren)

2/ Using alternative config

You may test, the new layer `dummy` doesn't exist in the default config:
[Default settings](https://mf-geoadmin3.int.bgdi.ch/config/index.html?lang=de&layers=dummy)
Compared to default:

- Label is different
- Opacity is `0.5`
- No update
- Only four topics
- ...

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/config/index.html)</jenkins>

### DIFF
--- a/rc_dev
+++ b/rc_dev
@@ -14,3 +14,4 @@ export WMS_URL=//wms-bgdi.dev.bgdi.ch
 export WMTS_URL=//tod.dev.bgdi.ch
 export TERRAIN_URL=//3d.dev.bgdi.ch
 export VECTORTILES_URL=//vectortiles{s}.dev.bgdi.ch/3d-tiles
+export CONFIG_URL=//mf-geoadmin3.dev.bgdi.ch
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 /src/embed.html
 /src/TemplateCacheModule.js
 /prd
+/configs
 /.build-artefacts
 /apache/app.conf
 /test/karma-conf-*.js
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,9 @@ TECH_SUFFIX = .bgdi.ch
 API_URL ?= //api3.geo.admin.ch
 API_TECH_URL ?= //mf-chsdi3.
 LAST_API_URL := $(call lastvalue,api-url)
+CONFIG_URL ?= //map.geo.admin.ch
+CONFIG_TECH_URL ?= //mf-geoadmin3.
+LAST_CONFIG_URL := $(call lastvalue,config-url)
 ALTI_URL ?= //api3.geo.admin.ch
 LAST_ALTI_URL := $(call lastvalue,alti-url)
 ALTI_TECH_URL ?= //mf-chsdi3.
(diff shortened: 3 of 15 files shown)


gjn:
btw: the current PR doesn't change functionality _at all_ (note that the testlink causes an overwrite of api, which will not happen in case of real api outage).

gjn:
Discussed with @procrastinatio and @pakb . We agree that hardcoded stuff is not needed. Configurations (layersConfig, services and catalogserver) will be statified on same s3 bucket as map.geo. And we update the machnic to update this file (without having to re-deploy the application).

davidoesch:
Folks - just make sure that in any case a service (print, search, api etc) is not running/misconfigured: map.geo.admin.ch must allways
- load
- tiles loading must work
-adding removeing layers must work

if print and search are down--the above functions shall not be impacted

procrastinatio:
@davidoesch No worry, files are in the same bucket, only changing the way they are managed.

rc_dev
```diff
@@ -14,3 +14,4 @@ export WMS_URL=//wms-bgdi.dev.bgdi.ch
 export WMTS_URL=//tod.dev.bgdi.ch
 export TERRAIN_URL=//3d.dev.bgdi.ch
 export VECTORTILES_URL=//vectortiles{s}.dev.bgdi.ch/3d-tiles
+export S3_CONFIG_BASE=swisstopo-internal-filesharing.s3.amazonaws.com/config/mf-geoadmin3/dev
```
danduk82: @procrastinatio , `swisstopo-internal-filesharing.s3.amazonaws.com` is a bad candidate to store client configuration. Or do I get it wrong? Is the application using it directly or is it used by chsdi3?
There is no cloudfront in front of that bucket, and probably routing also is not working outside swisstopo (unverified). For the client, I would statify the configuration in a bucket with cloudfront. Might be a dedicated one or the one of map.geo.admin.ch.

gjn:
Use the same buckets as the applications. We also have stagings.

So map.geo config should be served from map.geo url. Easy to understand. 

Also, when we violate this, we probably break offline.

rc_dev
procrastinatio: It is still WIP. And with `mf-geoadmin3.dev.bgdi.ch`files are served directly by `mf-chsdi3.dev.bgdi.ch`as now.

pakb:
@boecklic how much of this has been integrated in https://github.com/geoadmin/mf-geoadmin3/pull/4764 ? seems pretty out of sync with your refactor (will be hard to rebase if needed)

boecklic:
without checking line-by-line I thinks these changes have been already merged with #4764 . I'll close therefore.

PR #4719, opened 2019-02-01
[WIP][MVT] Mvt offline

boecklic:
DONT MERGE! This PR is used to see the current difference between `mvt` including offline stuff (but not yet style 006) and `master` to be able to do the copy-pasting


danduk82:
see `mvt_clean`

PR #4720, opened 2019-02-02
add wmts canton solothurn

edigonzales:

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/wmts-kanton-solothurn/index.html)</jenkins>


edigonzales:
I don't get it. Why does the last element of an array need a comma? There was no comma before. Or am I too blind to see?

danduk82:
@edigonzales , the code is OK, but before merging this could you try to optimize a bit your WMTS config? When loading your layers there is a white bounding box around that overrides the basemap...
image

edigonzales:
https://github.com/sogis/docker-mapcache/issues/16

davidoesch:
@andrea - not our problem. Can't be solved. This is the price of wmts
U can merge

edigonzales:
btw: when will the updated url be visible?

davidoesch:
After the deploy - see release notes

PR #4758, opened 2019-02-15
improve submodule handling

boecklic:

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/boc_improve_build_external_libraries/index.html)</jenkins>

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@ Next generation map.geo.admin.ch
 
 # Getting started
 
+## Get the code
 Checkout the source code:
 
     $ git clone https://github.com/geoadmin/mf-geoadmin3.git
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -16,6 +17,19 @@ or when you're using ssh key (see https://help.github.com/articles/generating-ss
 
     $ git clone git@github.com:geoadmin/mf-geoadmin3.git
 
+If you checkout an `mvt`-branch (like `mvt_clean` or any descendants) for the first time, you have to 
+run
+
+    $ make init-submodules
```
pakb: Isn't it only mandatory if we want to update our two forked dependencies ? (ol-mapbox-style and ol-cesium)
Otherwise, the content of /src/lib is what is taken into account, right?

boecklic: For the build of the app yes, only what's in `src/lib/` is currently considered. If one is never going to touch anything concerning the two libs, the `make init-submodules` (or what's behind) can be ignored. But once you start changing sth there (be it changes by us or considering upstream changes from openlayers) this should be run.

PR #4771, opened 2019-02-20
Absolute path for appcache

procrastinatio:

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/mom_appcache/1902201850/index.html)</jenkins>


boecklic:
Fixed with #4774

PR #4773, opened 2019-02-21
fix paths for config mako

pakb:


pakb:
it wan't the problem, the target `deploydev` was not yet adapted to the new build mechanism

PR #4797, opened 2019-02-28
Project <MVT> has no deploy target <dev>

procrastinatio:
This PR is for not trying to deploy <MVT> project to a `dev` bucket.

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/mom_deploy_mvt/1903041330/index.html)</jenkins>

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -159,11 +159,12 @@ node(label: 'jenkins-slave') {
       if (project == 'mf-geoadmin3') {
         targets = ['dev', 'int']
       } else {
-       targets = ['int']
+        targets = ['int']
       }
       for (target in targets) {
         echo 'Activating on ' + target
-        sh 'echo "yes" | PROJECT=' + project + ' .build-artefacts/python-venv/bin/python ./scripts/s3manage.py activate --branch ' + deployGitBranch + ' --version ' + deployedVersion + ' ' + target
+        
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -57,8 +57,12 @@ node(label: 'jenkins-slave') {
       
       parallel (
        'dev': {
-         stdout = sh returnStdout: true, script: 'make s3copybranch PROJECT='+ project +   ' DEPLOY_TARGET=dev DEPLOY_GIT_BRANCH=' + deployGitBranch
-         echo stdout
+         if (project == 'mf-geoadmin3') {
+           stdout = sh returnStdout: true, script: 'make s3copybranch PROJECT='+ project +   ' DEPLOY_TARGET=dev DEPLOY_GIT_BRANCH=' + deployGitBranch
+           echo stdout
+         } else {
```
pakb: I would prefer to specify what `else` is with a `else if`, so that it auto-documents itself

Jenkinsfile
```diff
@@ -149,11 +153,18 @@ node(label: 'jenkins-slave') {
     }
 
     stage('Activate dev/int') {
-      def targets = ['dev', 'int']
+      echo 'Project <' + project + '>'
       echo 'Activating the new version <' + deployedVersion + ' of branch  <' + deployGitBranch + '>'
+      def targets = []
+      if (project == 'mf-geoadmin3') {
+        targets = ['dev', 'int']
+      } else {
```
pakb: Same as above, I would prefer a `else if`

Issue #4772, opened 2019-02-21
Remove GooglePlus icon from the share menu


fredj:
Google+ APIs will be shut down on 7 March 2019

davidoesch:
Yep remove it

PR #4799, opened 2019-03-01
Removing Google plus 

procrastinatio:
See #4772

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/google_plus/1903041259/index.html)</jenkins>


src/locales/en.json
```diff
@@ -248,9 +247,9 @@
 "medium_size": "Medium",
 "menu": "Menu",
 "metadata_window_title": "Infobox",
-"meteoschweiz": "MeteoSwiss",
+"meteoschweiz": "MeteoSuisse",
 "meteoschweiz_service_link_href": "http://www.meteoswiss.admin.ch",
-"meteoschweiz_service_link_label": "MeteoSwiss",
+"meteoschweiz_service_link_label": "MeteoSuisse",
```
pakb: I would prefer to change than in a separate PR, possibly linked to an issue. That's a bit out of the blue and I can't know if it's intended or not.

procrastinatio: OK, reverted MeteoSwiss changes

PR #4811, opened 2019-03-04
Build : node version checking

pakb:
Only call NVM if version mismatch, greatly increase build time if no need of NVM

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/build_node_version_checking/1903041642/index.html)</jenkins>


procrastinatio:
> 5 minutes to <1 minutes. Nice!

PR #4818, opened 2019-03-05
Autodetect if on MVT branch, and set PROJECT accordingly

pakb:

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/ltbtp_build_autodetect_mvt_project/1905071129/index.html)</jenkins>

### DIFF
--- a/mk/config.mk
+++ b/mk/config.mk
@@ -13,9 +13,6 @@ endef
 # rc file used
 USER_SOURCE ?= rc_user
 
-# mf-geoadmin3 or mvt_clean
-PROJECT ?= mf-geoadmin3
-
 # Map libs variables
 OL_VERSION ?= be12573# September 25 2018 (mind the absence of a space character after the version)
 OL_CESIUM_VERSION ?= 695253277d66a8917fa456f3346c049b88f88eb7 # October 12 2018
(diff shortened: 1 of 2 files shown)


mk/config.mk
```diff
@@ -32,8 +32,8 @@ GIT_BRANCH ?= $(shell git rev-parse --symbolic-full-name --abbrev-ref HEAD)
 
 
 # possible values : mf-geoadmin3 or mvt (in the end, anything else than mf-geoadmin3 could do the trick to switch to testviewer deploy)
-# auto-detect MVT branches if branch name contains "mvt_", otherwise falls back to standard mf-geoadmin3
-ifneq (,$(findstring mvt_,$(GIT_BRANCH)))
+# auto-detect mvt_clean branches, otherwise falls back to standard mf-geoadmin3 (same behavior as in the Jenkinsfile)
```
procrastinatio: Hmm...
Do we want to deploy branches for project `mvt`. If so, this is not enough
`Jenkins` deploys two branches `master`and `mvt_clean`, and `PR`. with PR, we know on which on which master it will be merged. `Makefile`has only a branch name to decide to which target to deploy.

gjn:
What speaks against having a dedicated mf-geoadmin4 project? Questions as in this PR won't even arise.

Exchanging code between 2 repositories is quite similar than exchanging between two branches. ga4 could have an (automatic) tracking branch of ga3 master where rebases/merge happen the sam as now.

Thoughts?

pakb:
That's a discussion we had a month ago or so with our team, and we decided that despite the seemingly appealing idea to have a separate repo, we should stick to a branch as long as this is not officially what we should do for next version (RE4).
IMHO whether we would be on a separate repo or not, this issue would have still come up, as this is configuration to ease test.map.geo.admin deployment.

gjn:
Well, this branch is RE4.

And a dedicated repo would make things easier as the only difference would be deploy targets (and maybe hostnames, if still used). No guessing (if else) needed if a branch is for RE4 or not.

PR #4863, opened 2019-03-27
Update license

davidoesch:
can a techie review content as well?

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/ltoed_license/1904250951/index.html)</jenkins>


pakb:
I've copy pasted the licence, without 80 char linting (don't know why it was linted this way).
Only thing not present in the original 3 ways BSD licence is `All rights reserved.`, do we keep it anyway?

PR #4866, opened 2019-03-29
remove package-lock and add *sublime* to .gitignore

boecklic:

<jenkins>[Test link](https://mf-geoadmin4.int.bgdi.ch/feature/remove_package_lock/1903290813/index.html)</jenkins>


pakb:
what do you mean by Sublime? things related to Sublime text?

boecklic:
> what do you mean by Sublime? things related to Sublime text?

jup, but just saw it's already in .gitignore... ;-)

pakb:
I would be ok having this change also on master, what do you think?

PR #4874, opened 2019-04-04
Run jenkins on everything but master

procrastinatio:
...hopefully, master will ever be called `master`

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/jenkins_all-slaves/2001090747/index.html)</jenkins>


pakb:
die Jenkins, die! (as we will be moving out of Jenkins, this is not relevant anymore)

PR #4907, opened 2019-05-02
[MVT] Link to testviewer gets the same url params as current app

pakb:
Test viewer doesn't react well when zoom is not an integer, something might have to change in `mvt_clean` to make it work properly

Will fix https://github.com/geoadmin/mf-geoadmin3/issues/4904

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/ltbtp_test_link_with_permalink_params/1905021228/index.html)</jenkins>


pakb:
https://github.com/geoadmin/mf-geoadmin3/pull/4908 needed to work properly

Issue #4604, opened 2018-11-28
[MVT] footer links


davidoesch:
if 

basemaplight or OSM bg is shown: footer has a link

"© OpenMapTiles, OpenStreetMap contributors, swisstopo” **with link ** to https://openmaptiles.org/, https://www.openstreetmap.org/copyright

davidoesch:
iOS Chrome and Safari. Footer not shown on test.map.geo.admin.ch.very same device for map.geo.admin.ch it works
![EBD2F385-80A6-4008-8BDD-43E50FFC5916](https://user-images.githubusercontent.com/4577727/57102438-c5b88d00-6d23-11e9-8339-db46dcc0311c.jpeg)

davidoesch:
Reported by: https://twitter.com/luxperisa/status/1124036218168184838?s=21

boecklic:
currently the footer (visible on desktop) says "geo.admin.ch   Copyright & data protection" (both parts with links, where the second one points to https://www.geo.admin.ch//about-swiss-geoportal/impressum.html#copyright, which doesn't exist). What's the correct swisstopo copyright link?

pakb:
seems like the page URL changes depending on the lang, see https://www.geo.admin.ch/de/ueber-geo-admin/impressum.html for instance.
So it seems also related to the issue https://github.com/geoadmin/mf-geoadmin3/issues/4913 as there's not `langId` in the URL you've linked

boecklic:
True, but the main problem is that the attribution is not shown for the vector tiles layer (all OS, all Browser)

pakb:
test link on https://github.com/geoadmin/mf-geoadmin3/pull/4917

davidoesch:
Not good on iPhone/mobile. It has to look like  Re3

PR #4917, opened 2019-05-03
[MVT] showing copyright links for mobile

pakb:
Will fix https://github.com/geoadmin/mf-geoadmin3/issues/4604

<jenkins>[Test link](https://mf-geoadmin4.int.bgdi.ch/mvt_clean_copyright_even_for_mobile/1905060551/index.html)</jenkins>

### DIFF
--- a/src/components/attribution/AttributionDirective.js
+++ b/src/components/attribution/AttributionDirective.js
@@ -95,6 +95,11 @@ goog.require('ga_event_service');
         $rootScope.$on('gaLayersTranslationChange', function() {
           updateDebounced(element, layersFiltered);
         });
+        $rootScope.$on('gaVectorTileInitDone', function() {
+          scope.layers = scope.map.getLayers().getArray();
+          layersFiltered = scope.layers.filter(scope.layerFilter);
+          updateDebounced(element, layersFiltered);
+        });
 
         // Watch layers with attribution from 3d globe
         scope.is3dActive = function() {
--- a/src/components/vectortile/VectorTileLayerService.js
+++ b/src/components/vectortile/VectorTileLayerService.js
@@ -177,10 +177,14 @@ goog.require('ga_browsersniffer_service');
               var layer = layerArrayAfterOlms[i];
               if (layer.get('mapbox-source')) {
                 layer.olmsLayer = true;
-                layer.parentLayerId = vectortileLayerConfig.serverLayerName;
+                layer.parentLayerId = getVectorLayerBodId();
+                layer.id = getVectorLayerBodId();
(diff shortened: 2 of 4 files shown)


danduk82:
in both [tileset.json](https://vectortiles.geo.admin.ch/mbtiles/ch.swisstopo.leichte-basiskarte.vt/v006.json) files, there is an `attribution` field, which can be used. In the [test mapbox viewer](https://vectortiles.geo.admin.ch/gl-styles/ch.swisstopo.leichte-basiskarte.vt/v006/#7/47/8/0/6), this field is used dynamically to aggregate all the various contributors displayed

davidoesch:
Not good....

davidoesch:
3D8AA04F-6C38-4C4F-9D72-55C5CF9AEE7E

DerKarlos:
Cyclic restart in iPad

pakb:
@DerKarlos with the test link on this PR or with test.map.geo.admin.ch?

DerKarlos:
> @DerKarlos with the test link on this PR or with test.map.geo.admin.ch?

Both of them. A view seconds the screen is getting full of dialogs, then iOS „thinks“, the page refresh cycle is to slow and reloads it.

pakb:
@DerKarlos Which model of iPad do you have (model and iOs version if possible)?
Do you run test.map.geo.admin.ch directly from Safari or do you use a third party browser?

We've done tests on iOs 12 on iPhone with Safari, and it worked well. So I'm wondering if you have maybe a different browser / system than the one we tested on.

davidoesch:
And make sure you are flushing local cache

DerKarlos:
I am using an iPad 5. Generation with iOS 12.2
I also could test with an iPhone 6 if you like (after clear cash, showing you page, Safari crashed)

Yes flushing (Verlauf und Websitedaten löschen) avoids using old versions of files.
And it offers the maximal load time :-)

I now only test this test.map.geo.admin.ch 
The iPad Safari did not crash but shows the “Wurde neu geladen, weil ein Problem …”

I will try to send a screenshot before reboot. It looks crowded and the function “auf ihre Startseite”
points to the frame.

Then I also used Firefox and got the same; probably  the same render engine anyway.


You certainly know this:
// To see errors on touch devices, activate this Errorhandler producing alert dialogs
window.onerror = function(msg,url,line){ alert(msg+'\n'+url+'\nLine: '+line); return true; };

boecklic:
Please don't comment on merged PRs (which solve sth unrelated to the issue anyhow.) The iOS issue is known and tracked here: #4924, hints and comments welcome!

Issue #4922, opened 2019-05-05
[MVT] Enabling GPS Undefined


m-kuhn:
When trying to enable the GPS on test.map.geo.admin.ch , a message box says "undefined". GPS on the device is turned on and working on the page map.geo.admin.ch .

![image](https://user-images.githubusercontent.com/588407/57194819-8a98a280-6f4b-11e9-83df-947cf51b2ff8.png)

davidoesch:
TBD is this is a wontfix during Testphase

fredj:
This one is probably caused by this change in OpenLayers: https://github.com/openlayers/openlayers/pull/8731

Quick (untested) fix:
```diff
diff --git a/src/components/geolocation/GeolocationDirective.js b/src/components/geolocation/GeolocationDirective.js
index 12c878b71..2b42fbacd 100644
--- a/src/components/geolocation/GeolocationDirective.js
+++ b/src/components/geolocation/GeolocationDirective.js
@@ -241,16 +241,16 @@ goog.require('ga_throttle_service');
           elts.removeClass(naClass);
           var msgId;
           switch (error.code) {
-            case error.PERMISSION_DENIED:
+            case 1:
               msgId = 'geoloc_permission_denied';
               break;
-            case error.POSITION_UNAVAILABLE:
+            case 2:
               msgId = 'geoloc_pos_unavailable';
               break;
-            case error.TIMEOUT:
+            case 3:
               msgId = 'geoloc_time_out';
               break;
-            case error.UNKNOWN_ERROR:
+            default:
               msgId = 'geoloc_unknown';
               break;
           }
```

Ref: https://developer.mozilla.org/en-US/docs/Web/API/PositionError

pakb:
working on a fix with @fredj's proposal here (thanks!) : https://github.com/geoadmin/mf-geoadmin3/pull/4923
Still need to fix our `flyTo` function that zooms where we are.

pakb:
Fix deployed on https://test.map.geo.admin.ch

PR #4923, opened 2019-05-06
[WIP] [MVT] geolocation fix

pakb:
Fixes the error message issue, but we still need to fix our `flyTo` function what we use in it is no longer available in OpenLayers 6.

will fix https://github.com/geoadmin/mf-geoadmin3/issues/4922

<jenkins>[Test link](https://mf-geoadmin4.int.bgdi.ch/mvt_clean_geolocation_error_fix/1905071247/index.html)</jenkins>

### DIFF
--- a/src/components/geolocation/GeolocationDirective.js
+++ b/src/components/geolocation/GeolocationDirective.js
@@ -241,16 +241,20 @@ goog.require('ga_throttle_service');
           elts.removeClass(naClass);
           var msgId;
           switch (error.code) {
-            case error.PERMISSION_DENIED:
+            // PERMISSION_DENIED
+            case 1:
               msgId = 'geoloc_permission_denied';
               break;
-            case error.POSITION_UNAVAILABLE:
+            // POSITION_UNAVAILABLE
+            case 2:
               msgId = 'geoloc_pos_unavailable';
               break;
-            case error.TIMEOUT:
+            // TIMEOUT
+            case 3:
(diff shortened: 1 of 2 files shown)


src/components/geolocation/GeolocationDirective.js
```diff
@@ -241,16 +241,20 @@ goog.require('ga_throttle_service');
...
               msgId = 'geoloc_pos_unavailable';
               break;
-            case error.TIMEOUT:
+            // TIMEOUT
+            case 3:
               msgId = 'geoloc_time_out';
               break;
-            case error.UNKNOWN_ERROR:
+            // UNKNOWN_ERROR
+            case 4:
```
fredj: According to the [spec](https://developer.mozilla.org/en-US/docs/Web/API/PositionError) there's no code number `4` (but it could be implemented in some browser).

And I'll suggest to add a `default` branch, something like:
```js
...
// UNKNOWN_ERROR
case 4:
default:
  msgId = 'geoloc_unknown';
  break;
};
```

davidoesch:
fyl to is a must

src/components/geolocation/GeolocationDirective.js
pakb: Good call, thanks

pakb:
Should be good, @davidoesch let me know if this is the way it should behave when activating geolocation (I had to change a little bit how flyTo defines the position it will go to, so be mindful to little details 😉 )

PR #4950, opened 2019-05-20
[MVT] Update OL to latest beta tag possible

pakb:
Still needs some fixing

<jenkins>[Test link](https://mf-geoadmin4.int.bgdi.ch/mvt_clean_update_ol/1905201514/index.html)</jenkins>


pakb:
going further than v6.0.0-beta.7 breaks our test, so we will stick to this tag for now.

Issue #4944, opened 2019-05-19
use offset=0 in profile and use d3 smoothing features instead


danduk82:
**as:** a user
**I want:** a better estimation of hicking time and kml length in the profile window
**because:** I am always complaining about it

MUST have:
- [ ] mf-geoadmin3 should request service-alti with paramter `offset=0`
- [ ] must be related with [this issue in service-alti](https://github.com/geoadmin/service-alti/issues/42) so that the points digitized by the user can be displayed in the profile
- [ ] use _d3_ for smoothing, because smoothing looks good

CAN have:
- [ ] request service-alti at maximal possible resolution [once it is implemented](https://github.com/geoadmin/service-alti/issues/43)

davidoesch:
Related to https://github.com/geoadmin/service-alti/issues/43

davidoesch:
Related to https://github.com/geoadmin/service-alti/issues/42

boecklic:
@boecklic commented: Related to https://github.com/geoadmin/service-alti/issues/43

boecklic:
@boecklic commented: Related to https://github.com/geoadmin/service-alti/issues/42

PR #4993, opened 2019-07-29
Profile cleanup integration

pakb:
fix https://github.com/geoadmin/mf-geoadmin3/issues/4944

This changes the way geoadmin request profile to the backend, it adds an extra param to integrate points given in the geom and to not exceed altitude model's resolution when filling points.
Therefore there's no more need of a smoothing as there won't be "steps" anymore

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/ltbtp_profile_with_resolution_param/2003091245/index.html)</jenkins>

### DIFF
--- a/src/components/profile/ProfileService.js
+++ b/src/components/profile/ProfileService.js
@@ -41,7 +41,7 @@ goog.require('ga_urlutils_service');
 
     // Utils functions
     var createArea = function(domain, height, elevationModel) {
-      return d3.area().curve(d3.curveBasis).x(function(d) {
+      return d3.area().x(function(d) {
         return domain.X(d.domainDist);
       }).y0(height).y1(function(d) {
         return domain.Y(d.alts[elevationModel]);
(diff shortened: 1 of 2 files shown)


pakb:
[Test link with service-alti (dev)](https://s.geo.admin.ch/8373ebe2b9)

pakb:
Before
image
https://s.geo.admin.ch/8373f02147

After
image
https://s.geo.admin.ch/8373f15c22

davidoesch:
We need then to investigate the final result. TBD :Smoothing below  resolution of the original data ... Does it make sense

pakb:
Let me clarify what has changed :

| Before | After |
| --- | --- |
| Filling linestring (what is referred sometime as the KML, but is in fact a GeoJSON) with 200 points whether it is useful or not (points were closer than 2m, creating this steps shape) | Placing a point every 2m (resolution of the altitude model, see screenshot below for a profile without smoothing), if this placement results in more than 500 points, resolution is increased to match 500 points |
| ignoring points given in the middle of the linestring | making sure points given are included in the profile.  |
| 200 points per default | 500 points |

here is a screenshot of the same profile without frontend smoothing
image

IMHO we could skip smoothing as you said, because it doesn't really enhance the visual

pakb:
here's an example with a peak, where a point is placed roughly on the peak (4274m)
before
image

after
image

Issue #5048, opened 2019-10-01
Touch sensitivity in iframe


gjn:
The touch sensitivit on mobile devices (iPad, iPhone, Android, etc) is not working very good. It's especially visible on the pages of meteo schweiz:

https://www.meteoschweiz.admin.ch/home/messwerte.html?param=messwerte-lufttemperatur-10min

When touching a point, it should be selected and the iframe will sent a message (as shown in this codepen https://codepen.io/geoadmin/pen/yOBzqM?editors=0010) to the parent application which shows more information about the selected station. But currently, this is close to unusable on touch devices.

Note that the meteoschweiz integration uses `notooltip=true` and `showtooltip` parameters.

pakb:
What do you mean by it's not working? Here's what I have on my phone https://photos.app.goo.gl/tgsZfSprbAxKRRnHA
I don't see any big issue with this.

pakb:
If the problem is « big fingers » or lack of accuracy on the user side, we could add an extra URL param (`touchSnapping=true` or something similar) that would search for a bounding box around the touch event.
This way we could return the feature closest to the touch position.

pakb:
Tolerance wasn't taken into account for identify on vector layers (so touch had to be pixel perfect, hence the difficulty).
It's now implemented and can be tested (with a meteoschweiz layer) here : https://mf-chsdi3.int.bgdi.ch/shorten/84d53ab7cd

davidoesch:
Now it works fine

PR #5052, opened 2019-10-07
Adding tolerance for VectorLayer identify requests

pakb:
Will fix https://github.com/geoadmin/mf-geoadmin3/issues/5048


pakb:
not optimal yet, as the first occurence return by OL is selected (which is not necessarily the closest to the point)

pakb:
done, please review as we want to deploy this very soon

Issue #5055, opened 2019-10-11
identify tolerance on vector tooltip wrong


eraymann:
all OS 
all platform

**how to reproduce**

open map.geo.admin.ch
add two kmls https://s.geo.admin.ch/84e500f436

klick on one point which is close to antoher point of the other kml

**result**
highlight ok but tooltip contains info from a not highlighted  point from other kml

![image](https://user-images.githubusercontent.com/29711349/66631673-012a8200-ec07-11e9-8369-b08ba09e0837.png)

**expected result**
only highhlighted geometries are shown in tooltip

**workaround**
none

@davidoesch  thinks: regress of 
https://github.com/geoadmin/mf-geoadmin3/pull/5052

davidoesch:
Raised the prio to a showstopper, since the kml choosen is arbitrary.
and this feature is core for the validation of the addresses with the cantons

pakb:
would it make more sense to highlight multiple features with a single click, like the identify for any other type of layers (WMTS, WMS, ...), and show all highlighted features in the tooltip?

davidoesch:
it should highlight and show only the closest one. if there is a need to select multiple  points: then the user should use select by rectangle.
 do mind: this should work next tuesday 15.10 since there will be a communication to all cantons

pakb:
so what happend when there's two sources of features?
for instance I activate both "Temperature 2 m, 10 min" and "Sunshine duration, 10 min", I click on a station, which tooltip should I show? (both layers have the same stations so same positioning)

pakb:
Underlying question : why are we doing something different only for vector layers? our API works as I have described above, a click returns every features under the click taking tolerance into account.

AFoletti:
I think it actually gets just one feature, but actually duplicates it and opens a tooltip with different headers.
One is 100% right, the second one has all the attributes of the first (and is actually the same object) but has the header of the second layer.

davidoesch:
> Underlying question : why are we doing something different only for vector layers? our API works as I have described above, a click returns every features under the click taking tolerance into account.

it dos not work like this for the issue stated on top. you get arbitrary results.....

davidoesch:
@pakb pls do quick call @AFoletti he will explain

AFoletti:
Additional problem: the tooltip for the light blue polygons does not work, while it does for the violet ones

pakb:
[...]

PR #5056, opened 2019-10-14
Fix and improve layer filter with vector layer identify

pakb:
Fix https://github.com/geoadmin/mf-geoadmin3/issues/5055

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/ltbtp_vectorlayer_identify_better_filtering/1910141015/index.html)</jenkins>


AFoletti:
Functionality is OK.
I'll let a technician review the code and merge that.

As already discussed, this should be deployed today

AFoletti:
By the way: great work @pakb fixing that so quickly

PR #5062, opened 2019-10-23
Update ImportController.js

dkottmann:
@davidoesch 

added https://geodienste.ch/db/npl_nutzungsplanung/ita


davidoesch:
@dkottmann layers are empty?

p1d1d1:
@davidoesch zoom in

davidoesch:
Not that useful. This is why for visualization thou shall make your data visible for all all scales.  I was browsing in Ticino (hey it is an ita lang parameter)
From a best practice view: @dkotmann for the  convenience of your users —- activate the corresponding scales. 
Otherwise lgtm

PR #5103, opened 2020-01-06
chore(package): update sinon to version 8.0.2

pakb:
Closes #5098


pakb:
won't upgrade to sinon v8 yet, it breaks stuff in the build pipeline

PR #5154, opened 2020-04-17
fix uri for wms wildruhezonen

dkottmann:
@davidoesch


davidoesch:
@procrastinatio  why do checks fail?

pakb:
Jenkins is not able to check forks, but AWS CodeBuild does.

pakb:
you can check the result on int https://mf-geoadmin3.int.bgdi.ch/

Issue #5175, opened 2020-05-25
improve layerconfig make targets


ltclm:
some of the catalog json on s3  and in cloudfront are not valid:
https://map.geo.admin.ch/configs/de/catalog.bafu.json -> error
https://map.geo.admin.ch/configs/fr/catalog.bafu.json -> error
https://map.geo.admin.ch/configs/it/catalog.bafu.json -> correct

The makefile targets for the generation of the config json should be improved
- [x] check if mf-chsdi response is a valid json 

@pakb what do you think?

_Originally posted by @ltclm in https://github.com/geoadmin/mf-geoadmin3/issues/5174#issuecomment-633555854_

PR #5176, opened 2020-05-25
makefile configs target assure output quality

ltclm:

triggered by: https://github.com/geoadmin/mf-geoadmin3/issues/5175

config file generation with api calls ``make configs/`` will throw an error if:
* http status is not 200
* response is not a valid json

<jenkins>[Test link](https://mf-geoadmin3.int.bgdi.ch/feature_curl_fail/2005260650/index.html)</jenkins>

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -53,6 +53,7 @@ node(label: 'jenkins-slave') {
     }
 
     stage('Build') {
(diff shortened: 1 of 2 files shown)


ltclm:
thanks for review, i have added a jq test and installation for jenkins.

PR #5186, opened 2020-07-07
Allow bigger KMLs

procrastinatio:
Allow KML up to 25MB

This won't work
https://map.geo.admin.ch/master/1912040901/index.html?layers=KML%7C%7Chttps:%2F%2Fcms.geo.admin.ch%2FTopo%2FKML%2Fswissbathy3d_verfuegbarkeit_de.kml

With this PR
https://mf-geoadmin3.dev.bgdi.ch/feature_bigger_kml/2007071838/index.html?layers=KML%7C%7Chttps:%2F%2Fcms.geo.admin.ch%2FTopo%2FKML%2Fswissbathy3d_verfuegbarkeit_de.kml

:bomb: But the main issue, is that the KML imported via params is check for size or via the `all-in-one`  bar, while the manually imported one is not.

### DIFF
--- a/src/components/import/FileService.js
+++ b/src/components/import/FileService.js
@@ -16,7 +16,7 @@ goog.provide('ga_file_service');
 
         // Test the validity of the file size
         this.isValidFileSize = function(fileSize) {
-          return !(fileSize > 20000000); // 20 Mo
+          return !(fileSize > 25000000); // 25 Mo
         };
 
         this.isWmsGetCap = function(fileContent) {


davidoesch:
Because:a 20MB Kml will not Render on a weak pc / phone

That is why. And it is good so

pakb:
So should we really raise this bar if a weak device is already not comfortable handling a 20Mo KML? Or could we also speculate that a weak device from the time when this limit was decided is now not on the market anymore and that today's weak devices are more efficient?

procrastinatio:
The main issue is that depending on how you import you KML, the size is check or not.

You main drop a KML of any size if you want. You may also import it. But you link won't work, that's the issue.

:bomb: But the main issue, is that the KML imported via params is check for size or via the all-in-one bar, while the manually imported one is not.

PR #5190, opened 2020-07-28
Remove cache prefix in API URLs

pakb:
Now that the whole API is behind CloudFront, this will make the cache invalidation easier.

Test with (using prod api3): https://mf-geoadmin3.dev.bgdi.ch/ltbtp_remove_cache_prefix/2203091629/index.html?api_url=//api3.geo.admin.ch

### DIFF
--- a/src/index.mako.html
+++ b/src/index.mako.html
@@ -806,7 +806,6 @@ itemscope itemtype="http://schema.org/WebApplication"
           qrcodePath: qrcodePath,
           adminUrlRegexp: adminUrlRegexp,
           hrefRegexp: new RegExp(cfg['href_regexp']),
-          cachedApiUrl: apiUrl + cacheAdd,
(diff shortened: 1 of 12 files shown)


pakb:
from what I can see, this was used for these 4 endpoints
```
/rest/services/all/MapServer
/rest/services/{Topic}/SearchServer
/rest/services/{Topic}/MapServer/{Layer}/{Feature}
/rest/services/{Topic}/MapServer/{Layer}/{Feature}/htmlPopup
```

procrastinatio:
So
`cached` have headers for one year
Non `cached` one, have header for half-hour

procrastinatio:
Apparently, the services that have issues are `identify` (which it not cached) and `htmlPopup` (which is cached)

boecklic (approved):

hansmannj (commented):
Sorry, I was too late. Looks good, as far as I can tell, thanks!

PR #5218, opened 2021-02-03
update cadastre RM

davidoesch:
Updated cadastre RM https://www.cadastre.ch/de/home.html


ltclm:
https://github.com/geoadmin/mf-geoadmin3/pull/5221

PR #5221, opened 2021-02-03
update translations

ltclm:


src/locales/it.json
```diff
@@ -48,7 +48,7 @@
 "blw_service_link_label": "info@blw.admin.ch",
 "browse": "Browse",
 "cadastre": "Informazioni sui fondi",
-"cadastre_service_link_href": "http://www.cadastre.ch/internet/kataster/it/home.html",
+"cadastre_service_link_href": "https://www.cadastre.ch/en/home.html",
```
ltclm, line 51: @davidoesch it translation seems to be wrong in the google spreadsheet

davidoesch: bugger , fixed. sorry

ltclm:
should be good now, is it ok if we deploy this on the next data deploy 24.02.2021 or do we have to deploy this before?

PR #5224, opened 2021-02-23
Using last pip version using python2

procrastinatio:
The subsequent versions of `pip` have ditched support for python2


pakb (approved):
Good catch! thanks

PR #5228, opened 2021-03-05
Use the new proxy

procrastinatio:
Dropbox uses redirect, we had to adapt the proxy.

[Demo](https://mf-geoadmin3.dev.bgdi.ch/fix_5227_GPX_CORS/2103051805/index.html)

Try using this GPX: https://www.dropbox.com/s/i4qtmflv7vdpfhp/2021-03-04_325666018_TransalpMelsnachScuol.gpx?dl=1

and you should see a thin red line in the Graubünden

See also: https://github.com/geoadmin/service-proxy/pull/53

Looks like `geoadmin`is stilll using the old `ranjer`proxy on `dev`


davidoesch:
thanks ---

procrastinatio:
Merging and deploying next week...

hansmannj (approved):
:+1:

pakb (commented):
Are we only modifying `rc_dev`? I don't really get what we are doing here

boecklic:
> Are we only modifying `rc_dev`? I don't really get what we are doing here

Yeah, `service-proxy.int|prod.bgdi.ch` are already pointing to the new runtime location of service-proxy, `service-proxy.dev.bgdi.ch` is pointing to the old one. So only dev needs to be changed

PR #5203, opened 2020-08-27
add av-wms ita

dkottmann:
@davidoesch

PR #5233, opened 2021-04-07
add geodienste.ch AV Situationsplan

dkottmann:
@davidoesch


davidoesch:
@dkottmann  kannst du mich kurz anrufen? do morgen

davidoesch:
LGTM

pakb:
CORS header for the WMS server are not set on your side. I can merge that but geoadmin won't be able to retrieve the `GetCapabilities` because of this (so in essence it won't work as long as CORS aren't set properly on your server)

We already had this discussion on https://github.com/geoadmin/mf-geoadmin3/pull/5203 maybe it would be better to implement something server wide on your side, as your have many external WMS that are available in the Import function.

PR #5235, opened 2021-06-01
Update profile service documentation

davidoesch:
Update profile service documentation , based on feedback from  swisstopo APP and SWW


pakb (approved):
👍

PR #5242, opened 2021-10-21
BGDIINF_SB-2002: Bugfix on mf geoadmin build

ltkum:
the --no-sites-package was in legacy mode since a long time ago (before virtualenv 16.0), and was the standard way of functionning afterwards. With virtualenv 20, they removed it completely while forgetting to mention it in their changelogs. 

So... I also removed the test reports group (not the test themselves). I don't know in which capacity the group is used, and if we can live without it until mf-geoadmin is finally replaced. 

@davidoesch , @boecklic  what is your opinion on this ?

### DIFF
--- a/buildspec.yml
+++ b/buildspec.yml
@@ -60,8 +60,3 @@ phases:
     commands:
       - echo Build completed on `date`
 
-reports:
-  reportGroupJunit:
-    files:
-      - 'junit-reports/**/*'
-    file-format: JunitXml # default is JunitXml


ltkum:
> > My fix managed to let the application build, I'm now looking into the report group configuration, as it's what's making this fail for now.
> 
> Report group is a way of feeding back build results into codebuild. If not easily fixable it can be removed.

Well then, I'll try to fix this and, if I fail, at least we have something that works

PR #5256, opened 2022-01-11
Adding allow='geolocation' to the iframe code given to the user

pakb:
so that the geolocation can function properly if the user then decide to replace embed.html with the index or the mobile page


gjn:
LGTM

PR #5259, opened 2022-02-17
Update context tooltip with link to elevation

davidoesch:
context tooltip with link to elevation explanation

@pakb if we can do this in mf-geoadmin3 without huge effort lets do it there otherwise we take it into web-mapviewer


pakb:
will be added in the new viewer through https://jira.swisstopo.ch/browse/BGDIINF_SB-2274
this shortcut version of the link only redirects to the German version, so we've added all possible URLs to the translation file (new viewer key only)

PR #5273, opened 2022-05-23
Corrected host names

ltshb:
public.int.bgdi.ch has been migrated under sys-public.int.bgdi.ch, same as for service alti which is now under sys-api3.int.bgdi.ch


ltshb:
@pakb could you then re-deploy int on INT with this correction ?

PR #5276, opened 2022-06-03
Add WMS canton Jura, remove old WMS GA links

mariokeusen:
Added WMS service from canton JU to the list. Removed broken WMS GA links, because WMS GA was decommissioned recently


davidoesch (approved):
LGTM

PR #5274, opened 2022-05-23
BGDIINF_SB-2328, BGDIINF_SB-2493: Uses all backend from Frankfurt

ltshb:

PR #5278, opened 2022-07-04
BGDIINF_SB-2493: Use backend from Frankfurt on INT

hansmannj:
This PR is not meant to be merged. It should be delpoyed to allow testing the Frankfurt backend on INT.

### DIFF
--- a/mk/serviceconfig.mk
+++ b/mk/serviceconfig.mk
@@ -23,7 +23,7 @@ configs/: .build-artefacts/last-version \
           .build-artefacts/last-config-url \
           guard-LAYERSCONFIG_VERSION
 	mkdir -p $@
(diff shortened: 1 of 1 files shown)


ltshb (approved):
I suggest to create a separate PR with the `-L` fixes in `curl` command and merge this into master. Otherwise when we will want to merge the DEV, INT and PROD PRs we will have conflicts.
NOTE: as info the `-L` options allow curl to follow redirect which is required for `sys-api3` which redirects `http` to `https` while `mf-chsdi3` not.

pakb:
I don't know why, but I'm unable to build this specific branch in order to deploy it to INT (`make s3deployint` fails for some obscure reasons...)

PR #5280, opened 2022-07-07
BGDIINF_SB-2493: Use backend from Frankfurt on INT

hansmannj:
This PR is not meant to be merged. It should be delpoyed to allow testing the Frankfurt backend on INT.


hansmannj:
Seems the branch name could still be causing problems due to its length. 
Will close this PR and open yet another one with an even shorter branch name

PR #5281, opened 2022-07-07
BGDIINF_SB-2493: Uses all backend from Frankfurt on INT

hansmannj:
:warning: This PR is not meant to be merged. It should be delpoyed to allow testing the Frankfurt backend on INT.

(Deploying on INT now works with this short branch name)


pakb:
As we had issue deploying this specific branch on INT (no idea why though...) we merged the commit regarding `rc_int` on https://github.com/geoadmin/mf-geoadmin3/pull/5274 and the deploy worked.

PR #5284, opened 2022-08-17
BGDIINF_SB-2546: localhost:port addresses are using http

ltclm:
this will force localhost:port values in these parameters to http:
* api_url
* wms_url
* config_url
* ...

[Testlink](https://mf-geoadmin3.dev.bgdi.ch/feature_BGDIINF_SB-2546_localhost_http/2208172006/index.html)


rebert:
I realised, that the legend images do not appear when using localhost adresses

pakb (approved):
Let's try with that already, deploying this on DEV and INT, and then we can fix this legend issue if we see it's something related to the JS and not the backend

PR #5288, opened 2022-09-09
Reactivate tests while building

pakb:
Skipping test on KML download for iOs as we have now the same logic with iOs than with other browser, no need for specific testing


hansmannj (approved):
:+1: Thanks

PR #5289, opened 2022-09-09
Deactivating tests for deploy

pakb:
there is still issues while doing make s3deploydev (and int/prod) with the clonebuild.sh script, tests do not run properly in this instance


hansmannj (approved):
:+1:

PR #5291, opened 2022-11-07
BGDIINF_SB-2573: Build app in CI

ltshb:
This PR solve the following issues:

- Fixed minor makefile errors
- Now you can build the application from any device; I could do it from my own laptop and from my user account on `vpc-mf1-dev-1`
- The CI also build the app and upload the result on `build-artifacts-swisstopo`
- Removed some relic from an old time.

I tried to automatically deployed on DEV, but I failed due to permission issue. We can simply reactivate the auto deploy later on.


pakb (commented):
It might be a good idea to use nvm in the buildspec to set node to v10 (if this version is not available anymore with AWS CodeBuild images), otherwise the Makefile does it each command, and the build takes ages

pakb (approved):
🚀 

it seems like PhantomJS is behaving as it was during last deploys on our vhost and is going timeout for some reasons...
```
14 11 2022 12:21:33.195:WARN [PhantomJS 2.1.1 (Linux x86_64)]: Disconnected (0 times), because no message in 30000 ms.
```

pakb:
Nice! I don't know if we can reduce how verbose is PhantomJS, as it is now flooding our logs with one liner about each test (it's concatenated on a regular bash as a one liner, but not on the CI...)
Great work so far @ltshb you've done the impossible 😉

ltshb:
@pakb reducing the verbosity should be also fixed :-) see last commit.

pakb (approved):

PR #5299, opened 2023-04-12
BGDIINF_SB-2849: update copyright link and label

pakb:
After CMS migration, the link to the copyright is no more and is now replaced with a 'Terms of use' page. As the link is different for every language (can't just change the lang ISO code in the URL), I've also added the URL into the translation files.

Some unrelated translations have also been updated in the process


procrastinatio (approved):
OK for me.

PR #5313, opened 2024-05-15
Remove URL from testlink A tag

pakb:
so that only the click handler is used to determine the end location. With a href defined, some browser (like Firefox) would not let the handler do its things but open the URL created at startup, leading to the map always be on the test viewer where the legacy viewer started


ltshb (approved):

ltshb:
While testing I noted that the cursor on the try-test-map is now not anymore the pointer but the text selection.

PR #5315, opened 2024-05-28
No golive disclaimer in embed

pakb:
We support everything feature-wise (especially postMessage of feature selection), and the disclaimer takes way to much space on the screen for small embed use-cases


hansmannj (approved):
Thanks! :+1:

PR #5316, opened 2024-06-12
Fix service-proxy PROD URL

pakb:


ltshb (approved):

ltshb:
@pakb No need to deploy that on the old we can leave with it as I readded the ingress in k8s and we have a clean up ticket after go live PB-624

hansmannj (approved):
:+1: Thanks for fixing!

